PR #5, opened 2020-10-27
Add spec for AWS CDK

rk-yen:


falky97:
Finally we have some basic AWS stuff. Thank you @rk-yen. I will take a look this weekend at sprucing it up a little bit.

PR #6, opened 2020-10-27
Add spec for tar

rk-yen:
Something to think about ... should the specs be organized per 'OS' (in anticipation of support of different platforms).


falky97:
Yes, agree. Seperating (or tagging) specs by OS and/or shell makes sense. I think we will add this in later. We will likely have different installation steps which mean this won't be a breaking change.

Thank you again for this contribution 😄

PR #7, opened 2020-11-01
added some common options/args and single line commit suggestions for…

Chair4ce:
… commands that can use hashes for input


Chair4ce:
This might not be the best time to approve this one. The long strings returned by the generators gets cut off. Maybe when there is a good solution for displaying them this might be good to go.

PR #12, opened 2020-11-20
Add completion spec for brew services

LI0131:
This PR adds completion spec for `brew services`. I have included a generator for listing the different services a user has installed on their machine. It is used with each of the `brew services` subcommands outside of `cleanup` and `list`.

I am accomplishing this listing using `brew services list | sed -e 's/ .*//' | tail -n +2`. There may be a more elegant way to grab that input data, so I would love some criticism on it. :grin:

### DIFF
--- a/specs/brew.js
+++ b/specs/brew.js
@@ -1,3 +1,19 @@
+let servicesGenerator = {
+    variadic: true,
+    suggestions: [],
+    generator: {
+        script: "brew services list | sed -e 's/ .*//' | tail -n +2",
+        postProcess: function(out) {
+            return out.split('\n').filter((line) => {
+                return !line.includes('unbound') && {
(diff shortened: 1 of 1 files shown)


mattschrage:
This looks great! Thanks for contributing 🎉

Your approach seems like a fine way of handling the output. The alternative would be parsing with a regex in JS. I'm not sure which is more performant. 

Do you have any thoughts? I could add some parsing utilities on the Javascript side if that would be helpful.

LI0131:
@mattschrage I think the difference in the two approaches for such small input is probably negligible (I can't imagine people are managing _too_ many services with brew). If you're cool with it then that works for me 😎

Loving this tool so far. Hope I can help out some more going forward!

PR #13, opened 2020-12-12
Yarn autocomplete spec

samwinslow:
This covers most of the Yarn CLI although the generators could be improved with some feedback/clarification. I will leave some comments below on specific lines. 🍐🍐🍐


specs/yarn.js
```diff
@@ -0,0 +1,746 @@
...
+        {
+            name: "run",
+            description: "",
+            options: [
+            ],
+            args: [
+                // TODO get this generator to work and combine the logic of both of these
+                //     {
+                //         generator: {
+                //            script: "ls -1 $(yarn bin)", // ISSUE: this runs in /bin/sh, yarn may not be defined in sh PATH
```
samwinslow, line 616: This generator should get the binary scripts that are runnable in this package. A list of folder names in the `yarn bin` directory is sufficient to make these suggestions, but if Yarn is not present in the /bin/sh PATH spec, as is the case if it's not installed globally or is installed through a shim like `yvm`, this generator fails with "Command not found".

Is there a way to force Fig to run this script in the currently active shell, e.g. `zsh`?

specs/yarn.js
```diff
@@ -0,0 +1,746 @@
...
+                                let workspaces = package["workspaces"]
+                                if (workspaces) {
+                                    return workspaces
+                                }
+                            } catch(e) { }
+                            return []
+                        }
+                    }
+                },
+                // TODO arg 1 is script suggestion from the workspace specified in arg 0.
```
samwinslow, line 727: Here the next argument depends on the first. The script for arg1 would look something like `cat $(arg0)/package.json` and would then read package scripts from that output.

Is there a way to access the preceding args?

copy.sh
```diff
@@ -0,0 +1,3 @@
+# Copy specfiles to the Fig resource dir
```
samwinslow, line 1: this little script saved me a lot of time!

specs/yarn.js
mattschrage: Currently this isn't possible.

We could add an API like this:
```
script: (context) => {
    // where context is an array of tokens eg. [ "yarn", "workspace", "arg0"]
    return `cat ${context.pop()}/package.json`
}

```

I'm not sure whether it makes more sense for the `context` array to only contain arguments or the full set of tokens.

specs/yarn.js
mattschrage: Thanks for pointing this out. The shell process has to be the same across all completion specs because everything is being handled by a single pseudo-terminal in the background. 

Perhaps Fig should use the user's default shell rather than /bin/sh? What do you think?

copy.sh
mattschrage: Nice!

mattschrage (approved):
Looks great! I wrote a few comments - curious to hear your thoughts about the new APIs?

falky97:
Sam this is awesome! You rock

PR #15, opened 2020-12-22
ENG-115: Build `chmod` completion spec

jsj:
ezgif com-gif-maker

https://linear.app/fig/issue/ENG-115/build-chmod-completion-spec
Signed-off-by: James Jackson <james.sorley.jackson@gmail.com>

### DIFF
--- a/specs/chmod.js
+++ b/specs/chmod.js
@@ -1,0 +1,84 @@
+var completionSpec = {
+    name: "chmod",
+    description: "Change file modes or Access Control Lists.",
+    args: [
+        {
+            name: "mode",
+            optional: false,
+            suggestions: [ 
+                // Some of the most common chmod's (non-exhaustive)
+                {   
+                    name: ["u+x",],
+                    type: "argument",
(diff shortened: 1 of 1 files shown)


specs/chmod.js
```diff
@@ -0,0 +1,79 @@
...
+                    name: ["777"],
+                    type: "argument",
+                    description: "⚠️ allows all actions for all users",
+                    icon: "🔐",
+                },
+            ],
+        },
+        {
+            // Modifying
+            optional: false,
```
jsj, line 44: This should be specified `optional` correct? It is mandatory, but in something like [`cat.js`](https://github.com/withfig/autocomplete/blob/master/specs/cat.js) it does not specify optional for it's file path, but I think it's just because when it's at the end it doesn't matter.

falky97: we haven't set up the parsing for optional yet. Makes things a little more complicated. doesn't exactly matter for the moment. But good to have the note here for future reference

PR #16, opened 2020-12-22
Add vercel spec

rolandtshen:
Added spec for Vercel CLI following https://vercel.com/docs/cli#commands/

### DIFF
--- a/specs/vercel.js
+++ b/specs/vercel.js
@@ -146,7 +146,7 @@ var completionSpec = {
             args: {
                 name: "team name",
                 description: "team to execute commands from",
-                generator: teamList
+                // generator: teamList
             }
         },
         {
@@ -223,7 +223,7 @@ var completionSpec = {
                         {
                             name: "name",
                             description: "name of the variable to remove",
-                            generator: envVarList
+//                             generator: envVarList
                         },
                         {
                             name: "environment",
@@ -324,7 +324,7 @@ var completionSpec = {
                 {
                     name: "team name",
                     description: "team to switch to",
-                    generator: teamList
+                    //generator: teamList
                 }
             ]
         },
@@ -344,8 +344,8 @@ var completionSpec = {
             args: {
                 name: "deployment url",
                 description: "URL of the deployment to remove",
-                variadic: true,
-                generator: deploymentList
+                // variadic: true,
+                //generator: deploymentList
             },
             options: [
                 {
@@ -365,7 +365,7 @@ var completionSpec = {
                 {
                     name: "deployment url",
                     description: "URL of the deployment to remove",
-                    generator: deploymentList
+                    // generator: deploymentList
                 }
             ]
         },
@@ -384,7 +384,7 @@ var completionSpec = {
                         {
                             name: "domain",
                             description: "domain to inspect",
-                            generator: domainList
+                            // generator: domainList
                         }
                     ]
                 },
@@ -405,7 +405,7 @@ var completionSpec = {
                         {
                             name: "domain",
                             description: "domain to remove",
-                            generator: domainList
+                            // generator: domainList
                         }
                     ]
                 },
@@ -426,12 +426,12 @@ var completionSpec = {
                         {
                             name: "domain",
                             description: "domain to move",
-                            generator: domainList
+                            // generator: domainList
                         },
                         {
                             name: "account name",
                             description: "account to move the domain to",
-                            generator: teamList // double check if teams = accounts
+                            // generator: teamList // double check if teams = accounts
                         }
                     ]
                 },
@@ -452,7 +452,7 @@ var completionSpec = {
                         {
                             name: "domain",
                             description: "domain to verify",
-                            generator: domainList
+                            // generator: domainList
                         }
                     ]
                 }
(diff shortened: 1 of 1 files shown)


specs/vercel.js
```diff
@@ -0,0 +1,574 @@
+var completionSpec = {
+    name: "vercel",
+    description: "CLI Interface for Vercel.com",
+    args: {},
```
falky97, line 4: does vercel actually take an argument? I haven't looked at the CLI docs. if yes, then this is correct. If no, then this is wrong and needs to be removed. Only include arguments when the command, subcommand, or option actually takes an argument. This is v important for our parsing

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+            description: "Add metadata for the deployment (e.g.: `-m KEY=value`). Can appear many times."
+        },
+        {
+            name: ["-C", "--no-clipboard"],
+            description: "Do not attempt to copy URL to clipboard"
+        },
+        {
+            name: ["-S", "--scope"],
+            description: "Set a custom scope",
+            args: {
```
falky97, line 79: possible to do a generator for this?

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                        {
+                            name: "environment",
+                            description: "environment to add the variable to"
+                        }
+                    ]
+                },
+                {
+                    name: "rm",
+                    description: "remove an environment variable",
+                    args: [
```
falky97, line 140: specifying the arguments here is good. now let's add generators to this. I'm almost certain you can run a command to generate a list of suggestions

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                        {
+                            name: "environment",
+                            description: "environment to remove from"
+                        }
+                    ]
+                },
+                {
+                    name: "pull",
+                    description: "download dev env variables from cloud and write to .env",
+                    args: [
```
falky97, line 154: if it's a file, specify template: "filepaths"

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                        {
+                            name: "file",
+                            description: "the file to write downloaded variables to"
+                        }
+                    ]
+                },
+                {
+                    name: "ls",
+                    description: "List environment variables for a specific environment",
+                    args: [
```
falky97, line 164: again, check for a generator

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                            description: "Environment to list variables for"
+                        }
+                    ]
+                }
+            ]
+        },
+        { 
+            name: "init",
+            description: "Initialize an example project",
+            args: [
```
falky97, line 176: are you sure this takes two arguments? seems a little weird. but if so all good

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                {
+                    name: "new project name",
+                    description: "initialize specific project locally and rename directory"
+                }
+            ]
+        },
+        { 
+            name: "list", 
+            description: "Lists deployments",
+            args: {
```
falky97, line 190: add a generator

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                    args: {
+                        variadic: true
+                    }
+                }
+            ]
+        },
+        { 
+            name: "ls", 
+            description: "Lists deployments",
+            args: {
```
falky97, line 207: add generator

Seems like this will overlap with the above generator. You can create a new object at the very top of the file called `generators`, then you could

e.g.
```
var generators = {
   list_projects: ()=> {
      script: "GET ALL PROJECTS",
      splitOn: "\n:
    }
}
```

then just call it from `generators.list_projects`

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                name: "url",
+                description: "The URL of the deployment to inspect"
+            }
+        },
+        { name: "login", description: "Logs into your account or creates a new one" },
+        { name: "logout", description: "Logs out of your account" },
+        { 
+            name: "switch", 
+            description: "Switches between teams and your personal account",
+            args: [
```
falky97, line 225: add a generator. there should be a command to list team names. or a command to list personal and team and you can do some filtering with a postProcess function to work it out

specs/vercel.js
```diff
@@ -0,0 +1,574 @@
...
+                {
+                    name: "command",
+                    description: "command to detailed information about"
+                }
+            ]
+        },
+        { 
+            name: "rm", 
+            description: "Removes a deployment",
+            args: {
```
falky97, line 245: I am going to stop comment "add a generator" from now on. But you get the drift. You've done a great first pass mapping out the CLI tool. Now let's add generators to as many arguments as we can and make the vercel CLI tool really good 👍

falky97 (changes requested):
You've done a great first pass mapping out the CLI tool. Now let's add generators to as many arguments as we can and make the vercel CLI tool really good 👍

rolandtshen:
Generators included, but still need to be tested once node bug is resolved

PR #18, opened 2020-12-23
Add SSH spec

rolandtshen:


rolandtshen:
Added past connections generator, pending Fig `history` support

PR #23, opened 2021-01-21
Add lint all command

rolandtshen:
Added support to lint all files in /specs at once, added npm test support, and fixed outdated specs with new format


rolandtshen:
Merging to add auto linter functionality

PR #28, opened 2021-02-08
add pre-subcommand Git options

samwinslow:
Expanding the autocomplete spec for Git to support options that come before the subcommand, e.g. I often use `git --no-pager diff <file>`.

Currently no support for options before subcommand

**Additional info:**


mattschrage:
Thanks Sam!

PR #33, opened 2021-03-06
add ruby specs

rolandtshen:

**Additional info:**

### DIFF
--- a/specs/rake.js
+++ b/specs/rake.js
@@ -1,0 +1,59 @@
+var completionSpec = {
+    name: 'rake',
+    description: 'a ruby build program with capabilities similar to make',
+    args: {
+      name: "targets",
+      variadic: true
+    },
+    options: [
+      {
+        name: [ '-n' , '--dry-run' ],
+        description: 'Do a dry run without executing actions.'
+      },
+      {
+        name: [ '-H', '--help' ],
+        description: 'Display this help message.'
+      },
+      {
+        name: [ '-I', '--libdir=LIBDIR' ],
+        description: 'Include LIBDIR in the search path for required modules.'
+      },
+      {
+        name: [ '-P', '--prereqs' ],
+        description: 'Display the tasks and dependencies, then exit.'
+      },
+      {
+        name: [ '-q', '--quiet' ],
+        description: 'Do not log messages to standard output'
+      },
+      {
+        name: [ '-f', '--rakefile=FILE' ],
+        description: 'Use FILE as the rakefile.'
+      },
+      {
+        name: [ '-r', '--require=MODULE' ],
+        description: 'Require MODULE before executing rakefile.'
+      },
+      {
+        name: [ '-s', '--silent' ],
+        description: "Like --quiet, but also suppresses the 'in directory' announcement."
+      },
+      {
+        name: [ '-T', '--tasks' ],
+        description: 'Display the tasks and dependencies, then exit.'
+      },
+      {
+        name: [ '-t', '--trace' ],
+        description: 'Turn on invoke/execute tracing, enable full backtrace.'
+      },
+      { name: [ '-h', '--usage' ], description: 'Display usage.' },
+      {
+        name: [ '-v', '--verbose' ],
+        description: 'Log message to standard output (default).'
+      },
+      {
+        name: [ '-V','--version' ],
+        description: 'Display the program version.'
+      }
+    ]
+  }
--- a/specs/rbenv.js
+++ b/specs/rbenv.js
@@ -1,0 +1,84 @@
+var completionSpec = {
+    name: 'rbenv',
+    description: '',
+    subcommands: [
+      {
+        name: 'commands',
+        description: 'List all available rbenv commands',
+        options: [
+          {
+            name: ["--sh"]
+          },
+          {
+            name: ["--no-sh"]
+          }
+        ],
+        subcommands: []
+      },
+      {
+        name: 'global',
+        description: 'Set or show the global Ruby version',
+        options: [],
+        subcommands: [],
+        args: {}
+      },
+      {
+        name: 'install',
+        description: 'Install a Ruby version using ruby-build',
+        options: [
+          {
+            name: [ '--version' ],
+            description: 'Show version of ruby-build ',
+            args: {}
+          }
+        ],
+        subcommands: []
+      },
+      {
+        name: 'local',
+        description: 'Set or show the local application-specific Ruby version',
+        options: [],
+        subcommands: []
+      },
+      {
+        name: 'rehash',
+        description: 'Rehash rbenv shims (run this after installing executables)',
+        options: [],
+        subcommands: []
+      },
+      { name: 'shell', description: 'Set or show the shell-specific Ruby version', subcommands: [] },
+      {
+        name: 'uninstall',
+        description: 'Uninstall a specific Ruby version',
+        options: [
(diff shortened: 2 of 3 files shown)


coreyja:
I opened a PR for `bundle` today, just an FYI since we are both in the Ruby realm lol

https://github.com/withfig/autocomplete/pull/49

dev/rails.ts
```diff
@@ -0,0 +1,28 @@
...
+    },
+    {
+      name: ["-F,", "--force-overwrite"],
+      description: "Overwrites all previously deployed files, if any.",
+    },
+    {
+      name: ["-D,", "--database"],
+      description:
+        "Specify the database to be used. Supported values are mysql, oracle, postgresql, sqlite3, frontbase and ibm_db",
+    },
```
QuiiBz, line 22: I believe this should take these suggestions: `mysql, oracle, postgresql, sqlite3, frontbase and ibm_db`

dev/rails.ts
```diff
@@ -0,0 +1,28 @@
...
+    {
+      name: ["-I,", "--internal"],
+      description: "Options to be passed to the upstream rails script.",
+    },
```
QuiiBz, line 26: I believe this should take a variadic argument.

dev/rake.ts
```diff
@@ -0,0 +1,60 @@
...
+      description: "Do a dry run without executing actions.",
+    },
+    {
+      name: ["-H", "--help"],
+      description: "Display this help message.",
+    },
+    {
+      name: ["-I", "--libdir=LIBDIR"],
+      description: "Include LIBDIR in the search path for required modules.",
+    },
```
QuiiBz, line 20: We should add `insertValue` of `--libdir=` and add a `LIBDIR` argument (which is a template of `filepaths`?)

dev/rake.ts
```diff
@@ -0,0 +1,60 @@
...
+    {
+      name: ["-f", "--rakefile=FILE"],
+      description: "Use FILE as the rakefile.",
+    },
+    {
+      name: ["-r", "--require=MODULE"],
+      description: "Require MODULE before executing rakefile.",
+    },
+    {
```
QuiiBz, line 37: See above

dev/rbenv.ts
```diff
@@ -0,0 +1,87 @@
+export const completionSpec: Fig.Spec = {
+  name: "rbenv",
+  description: "",
```
QuiiBz, line 3: Could you add a description?

dev/rbenv.ts
```diff
@@ -0,0 +1,87 @@
...
+  subcommands: [
+    {
+      name: "commands",
+      description: "List all available rbenv commands",
+      options: [
+        {
+          name: ["--sh"],
+        },
+        {
+          name: ["--no-sh"],
```
QuiiBz, line 13: `name` can be a string since there is only one value.  Same for underneath.

dev/rbenv.ts
```diff
@@ -0,0 +1,87 @@
...
+        },
+      ],
+      subcommands: [],
+    },
+    {
+      name: "global",
+      description: "Set or show the global Ruby version",
+      options: [],
+      subcommands: [],
```
QuiiBz, line 22: No need to specify `options` and `arguments` if they don't have values. Same for underneath.

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
+export const completionSpec: Fig.Spec = {
+  name: "ruby",
+  description: "Interpreted object-oriented scripting language",
+  options: [
+    {
+      name: ["--copyright"],
```
QuiiBz, line 6: `name` can be a string since there is only one value. Same for underneath.

QuiiBz (changes requested):

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
...
+    {
+      name: ["-0"],
+      description:
+        "Specifies the input record separator ($/) as an octal number.",
+      args: { name: "octal" },
+    },
+    {
+      name: ["-C"],
+      description: "Causes Ruby to switch to the directory.",
+      args: { name: "directory" },
```
cstrnt, line 22: We should add `template: "folders"` here

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
...
+    {
+      name: ["-F"],
+      description: "Specifies input field separator ($;).",
+      args: { name: "pattern" },
+    },
+    {
+      name: ["-I"],
+      description:
+        "Used to tell Ruby where to load the library scripts. Directory path will be added to the load-path variable ($:).",
+      args: { name: "directory" },
```
cstrnt, line 33: We should add `template: folders` here

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+    {
+      name: ["-S"],
+      description:
+        "Makes Ruby use the PATH environment variable to search for script, unless if its name begins with a slash. This is used to emulate #! on machines that don't support it, in the following manner: #! /usr/local/bin/ruby # This line makes the next one a comment in Ruby \\ exec /usr/local/bin/ruby -S $0 $*",
+    },
+    {
+      name: ["-T"],
+      description:
+        "] Turns on taint checks at the specified level (default 1).",
```
cstrnt, line 48:
```suggestion
        "Turns on taint checks at the specified level (default 1).",
```

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+    {
+      name: ["-h", "--help"],
+      description: "Prints a summary of the options.",
+    },
+    {
+      name: ["-i"],
+      description:
+        "Specifies in-place-edit mode. The extension, if specified, is added to old file name to make a backup copy.",
+      args: { name: "extension" },
```
cstrnt, line 78: this should be `isOptional`

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+    {
+      name: ["-w"],
+      description:
+        "Enables verbose mode without printing version message at the beginning. It sets the $VERBOSE variable to true.",
+    },
+    {
+      name: ["-x"],
+      description:
+        "] Tells Ruby that the script is embedded in a message. Leading garbage will be discarded until the first that starts with “#!” and contains the string, “ruby”. Any meaningful switches on that line will applied. The end of script must be specified with either EOF, ^D (control-D), ^Z (control-Z), or reserved word __END__. If the directory name is specified, Ruby will switch to that directory before executing script.",
```
cstrnt, line 116:
```suggestion
        "Tells Ruby that the script is embedded in a message. Leading garbage will be discarded until the first that starts with “#!” and contains the string, “ruby”. Any meaningful switches on that line will applied. The end of script must be specified with either EOF, ^D (control-D), ^Z (control-Z), or reserved word __END__. If the directory name is specified, Ruby will switch to that directory before executing script.",
```

dev/ruby.ts
```diff
@@ -0,0 +1,125 @@
...
+      args: { name: "directory" },
```
cstrnt, line 117: `template: "folders"`

cstrnt (changes requested):

dev/rake.ts
cstrnt: I think it's rather `template: "folders"` because it's a directory

PR #34, opened 2021-03-09
feat(specs): add support for git commit --amend

ygaitonde:
fig now suggests --amend for autocomplete

### DIFF
--- a/specs/git.js
+++ b/specs/git.js
@@ -241,6 +241,24 @@ var completionSpec = {
                     name: ["-v", "--verbose"],
                     description: "show unified diff of all file changes",
                 },
+                { 
+                    name: "--amend",
+                    insertValue: "--amend `{cursor}`",
+                    description: "amend staged changes to last commit",
+                    options: [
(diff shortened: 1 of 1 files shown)


specs/git.js
```diff
@@ -241,6 +241,24 @@ var completionSpec = {
...
+                            description: "keep the previous commit message",
+                        },
+                        {
+                            name: ["-m"],
+                            description: "update the commit message",
+                            args: {
+                                name: "branch"
+                            }
+                        },
+                    ]
```
rolandtshen, line 260: Fig doesn't currently support nested options within another options object, but you could move these option objects one level higher, so they're on the same level as '--amend', and they'd still pop up after someone enters --amend. Not ideal because those two flags would pop up following other option flags also, but it's a limitation with the spec right now

rolandtshen (changes requested):

ygaitonde:
duplicate of https://github.com/withfig/autocomplete/pull/68

PR #39, opened 2021-03-11
initial commit - created knex.js file

juancaruizc:
feature - add knex commands autocomplete

No knex completionSpec

knex completionSpec with up option

**Additional info:**

### DIFF
--- a/specs/knex.js
+++ b/specs/knex.js
@@ -1,20 +1,156 @@
 var completionSpec = {
   name: 'knex',
-  description: 'SQL in JavaScript',
+  description: 'SQL query builder for JavaScript',
   subcommands: [
     {
-      name: 'migrate:',
-      description: 'Traverse your database',
+      name: 'init',
+      description: 'Create a fresh knexfile.',
       args: {},
-      options: [
-        {
-          name: ['up'],
-          description: 'Traverse your database one up',
-          args: {
-            name: 'up',
-          },
-        },
-      ],
+      options: [],
+    },
+    {
+      name: 'migrate:make',
+      description: 'Create a named migration file.',
+      args: { name: 'migration name' },
+      options: [],
+    },
+    {
+      name: 'migrate:latest',
+      description: 'Run all migrations that have not yet been run.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'migrate:up',
+      description:
+        'Run the next or the specified migration that has not yet been run.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'migrate:rollback',
+      description: 'Rollback the last batch of migrations performed.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'migrate:down',
+      description:
+        'Undo the last or the specified migration that was already run.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'migrate:currentVersion ',
+      description: 'View the current version for the  migration.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'migrate:list|migrate:status',
+      description: 'List all migrations files with status.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'migrate:unlock ',
+      description: 'Forcibly unlocks the migrations lock table.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'seed:make ',
+      description: 'Create a named seed file.',
+      args: {
+        name: 'seed name',
+      },
+      options: [],
+    },
+    {
+      name: 'seed:run',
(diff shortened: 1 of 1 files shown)


mattschrage:
Hey Juan! Thanks so much for contributing. It looks like your spec might be slightly off. I think all of the commands that start with `migrate:` should be included as subcommands (rather than as arguments).

Arguments are parameters that are specified by the user. For instance, in the `knex` CLI things like `path`, `name` and `address` are arguments.
 
```
Usage: knex [options] [command]

Options:
  -V, --version                   output the version number
  --debug                         Run with debugging.
  --knexfile [path]               Specify the knexfile path.
  --knexpath [path]               Specify the path to knex instance.
  --cwd [path]                    Specify the working directory.
  --client [name]                 Set DB client without a knexfile.
  --connection [address]          Set DB connection without a knexfile.
  --migrations-directory [path]   Set migrations directory without a knexfile.
  --migrations-table-name [path]  Set migrations table name without a knexfile.
  --env [name]                    environment, default: process.env.NODE_ENV || development
  --esm                           Enable ESM interop.
  --specific [path]               Specify one seed file to execute.
  --timestamp-filename-prefix     Enable a timestamp prefix on name of generated seed files.
  -h, --help                      display help for command

Commands:
  init [options]                          Create a fresh knexfile.
  migrate:make [options] <name>           Create a named migration file.
  migrate:latest [options]                Run all migrations that have not yet been run.
  migrate:up [<name>]                     Run the next or the specified migration that has not yet been run.
  migrate:rollback [options]              Rollback the last batch of migrations performed.
  migrate:down [<name>]                   Undo the last or the specified migration that was already run.
  migrate:currentVersion                  View the current version for the  migration.
  migrate:list|migrate:status             List all migrations files with status.
  migrate:unlock                          Forcibly unlocks the migrations lock table.
  seed:make [options] <name>              Create a named seed file.
  seed:run [options]                      Run seed files.
  help [command]                  display help for command
  ```

File: specs/knex.js [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+    {
+      name: 'migrate:',
+      description: 'Traverse your database',
+      args: {},
+      options: [
+        {
+          name: ['up'],
+          description: 'Traverse your database one up',
+          args: {
+            name: 'up',
```
rolandtshen, line 14: Having up as an option would cause autocomplete to return something like `knex migrate: up` instead of `knex migrate:up`. You could set the subcommand's name to `migrate:up` if that's the behavior you're going for!

juancaruizc: Aweseom thank you for the feedback Roland! Will get this fixed

juancaruizc:
Built out most (or all) knex commands and options. I suspect I did it correctly this time haha but please let me know otherwise!

Looking forward to building out the rest of the options and arguments once this PR is ready to be merged

specs/knex.js
```diff
@@ -0,0 +1,156 @@
...
+      name: 'migrate:unlock ',
+      description: 'Forcibly unlocks the migrations lock table.',
+      args: {},
+      options: [],
+    },
+    {
+      name: 'seed:make ',
+      description: 'Create a named seed file.',
+      args: {
+        name: 'seed name',
```
rolandtshen, line 65: for a few of these, like migrate:currentVersion, migrate:unlock, and seed:make have an extra space at the end. This will cause Fig to enter something like `knex migrate:currentVersion\ ` into the terminal. Was this intentional? If you're looking to add a space afterwards, try using `insertValue: migrate:currentVersion ` instead!

specs/knex.js
```diff
@@ -0,0 +1,156 @@
...
+    },
+    {
+      name: ['--knexfile'],
+      description: 'Specify the knexfile path.',
+      args: {},
+    },
+    {
+      name: ['--knexpath '],
+      description: 'Specify the path to knex instance.',
+      args: {},
```
rolandtshen, line 101: same here: if you want to enter `--knexpath `, use `insertValue: '--knexpath '`

specs/knex.js
```diff
@@ -0,0 +1,156 @@
...
+    },
+    {
+      name: ['--cwd '],
+      description: 'Specify the working directory.',
+      args: {},
```
rolandtshen, line 106: spacing

specs/knex.js
juancaruizc: Definitely not intentional! Just pushed another commit fixing this :)

rolandtshen:
Thanks for making the changes! Looks good and am merging now :)

juancaruizc:
Great! Thank you so much for the feedback Roland!

Issue #43, opened 2021-03-11
Switch to TypeScript?


QuiiBz:
Hi,

I think that the autocomplete specs are really powerful and customizable. But writing them in plain JS is pain without auto-completion from IDEs. Do you plan to switch to `TypeScript` in the future?

I really think that it could help a lot of peoples to contribute to new autocomplete specs.

falky97:
Agree - we aren't typescript gods, would you be open to helping with this? 

We set up schema validation with https://joi.dev/ but obviously assistance when you are coding would be better

QuiiBz:
Sounds good - I will open a PR tomorrow.

falky97:
You're awesome, thank you so much. Everything here should be up to date: https://withfig.com/docs/autocomplete/api

QuiiBz:
Closing issue since we merged it in #44.

PR #44, opened 2021-03-12
Migration to TypeScript for specs

QuiiBz:
As in https://github.com/withfig/autocomplete/issues/43, I started a PR to add TypeScript support to Fig.

Hard to write error-prone specs without IDEs auto-completion.

Provide types for the specs, allowing for awesome auto-completion and faster development of new specs.

**Additional info:**
I'm still working on it, so expect type bugs for now.


QuiiBz:
I've added `TODO` comments to remove lines/files when the migration to TS is ready.

The specs now compile to the `dist` folder, so I changed the `copy.sh` to include files in it. I've allowed JS along with TS, so JS specs are still working great.

The new command to "lint" the specs is `typecheck`. The old one `test` is still working, but only on `.js` files, instead of the new one which works on both JS and TS specs.

You can take a look at how to write specs in TS with the types here: https://github.com/withfig/autocomplete/pull/44/commits/3b3d751c620a8f4a091967f11122179e23d84342#diff-6f6f4fc563a4439827032c3278ba487e9d4fa6c24a7ec4c6e41ca21005c726f4 Has you can see, it's very simple:
- Rename the extension to `.ts`
- Change the `var completionSpec` to `const <specName>: Fig.Spec`: This prevent errors from the TypeScript compiler to to find many variables with the same name (`completionSpec`).

And now you will get auto-completion and errors warning, thanks to the added type `Spec` from the `Fig` namespace.

I've also make sure that the existing JS specs still work great, along with added support to new/migrated specs in TS.

package.json
```diff
@@ -4,11 +4,14 @@
   "description": "Fig Autocomplete Spec Linter",
   "main": "linter.js",
   "scripts": {
-    "test": "node linter.js"
+    "test": "node linter.js",
+    "typecheck": "tsc --noEmit",
+    "dist": "tsc"
```
cstrnt, line 9: I think `build` would be a more fitting name imo

schemas/fig.d.ts
```diff
@@ -0,0 +1,100 @@
...
+    export interface NamedCompletion extends Omit<Completion, 'name'>{
+
+        name: SingleOrArray<string>;
+    }
+
+    // All the available templates
+    export type Template = 'filepaths' | 'folders';
+
+    // A single T object or an array of this T object
+    export type SingleOrArray<T> = T | T[];
```
cstrnt, line 28: Maybe move this one to the top of the file? I don't like using something before it's defined (Completion uses this)

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist",
+    "target": "es5",
+    "module": "ESNext",
+    "lib": ["dom", "ESNext"],
```
cstrnt, line 6: I don't think we need `dom` here, right?

package.json
QuiiBz: Instead of the `dist` command? Both are used, let's wait for the maintainer's choice.

tsconfig.json
QuiiBz: I added it because there was some `console.log` in the `fig` spec. I don't know if they should be removed or not.

schemas/fig.d.ts
QuiiBz: Right, let's change it.

QuiiBz:
The generated specs written TS are not working in fig: I found that changing the name of the spec (initially to `completionSpec`, changed to the name of the spec) make fig not using theses specs. I tried manually changing back to `completionSpec` and it work properly:
`cd` spec generated from TS, with name set to `cd`, not working:
```js
var cd = {
    name: "cd",
    description: "Change the shell working directory",
    args: {
        template: "folders"
    }
};
```
After setting the name back to `completionSpec`, it work:
```js
var completionSpec = {
    name: "cd",
    description: "Change the shell working directory",
    args: {
        template: "folders"
    }
};
```

We have two solutions:
- Change the fig app to allow any variable name for the spec
- Use the TypeScript compiler API to change the variable name back to `completionSpec` when compiling from TS to JS

schemas/fig.d.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // A string or a function which can have a T argument and a R result,
+    // both set to void by default
+    export type StringOrFunction<T = void, R = void> = string | Function<T, R>;
+
+    // A function which can have a T argument and a R result, both
+    // set to void by default
+    export type Function<T = void, R = void> = (param: T) => R;
```
cstrnt, line 15: Could you please move the declaration before it's being used? :)

schemas/fig.d.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @see https://withfig.com/docs/autocomplete/api#generator-object
+     */
+    export interface Generator {
+
+        template?: string;
+        script?: StringOrFunction<Array<String>>;
+        splitOn?: string,
+        postProcess?: StringOrFunction<string, Suggestion[]>;
```
cstrnt, line 94: Can this really be a string?

cstrnt (changes requested):

schemas/fig.d.ts
QuiiBz: I’ve made some mistakes on the specs types - will fix it tomorrow. Thanks!

schemas/fig.d.ts
QuiiBz: Sure!

compiler/log.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/**
+ * Log a message in the console. Can specify a Level, which
+ * will be INFO by default.
+ *
+ * @param message The message to log
+ * @param level The level to log
+ */
+export const log = (message: string, level: Level = Level.INFO) => {
```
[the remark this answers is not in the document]
QuiiBz: Could you provide an example please? I don’t get the point of using an abstract class for this.

compiler/compiler.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param file The file to process
+ */
+const processSpec = (file: string) => {
+
+    const source = fs.readFileSync(file).toString();
+    const result = ts.transpileModule(source, options);
+
+    let newName = path.basename(file, '.ts');
+
+    if(!newName.endsWith('.js'))
```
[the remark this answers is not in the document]
QuiiBz: I usually never add curly braces to one-instructions blocks, since it adds useless characters. Let’s wait for the maintainer's opinion.

schemas/fig.d.ts
QuiiBz: Fixed in cbeff29

schemas/fig.d.ts
QuiiBz: Fixed in cbeff29

tsconfig.json
QuiiBz: Fixed in cbeff29

package.json
QuiiBz: Changed at https://github.com/withfig/autocomplete/pull/44/commits/66b3d28faf850e10d267612fa9d0e4c9cb713ed2#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L9-L10 with the custom compiler.

compiler/log.ts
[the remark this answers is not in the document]
cstrnt: Sure

You can create the class like this:
```js
abstract class SpecLogger {
  static log(message: string, level: Level = Level.INFO){
      // your code here
  }
}
```

and then call it like that

`SpecLogger.log(...args)`

By doing that the code is more verbose and you don't have to search for the `log` implementation.

compiler/compiler.ts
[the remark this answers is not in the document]
cstrnt: IMHO characters are NEVER useless if they improve readability. And in the end the code will also be optimized by the typescript compiler either way. I mean it's just a small thing here, but as long as you don't have to write some extremely optimized code which won't be transpiled / compiled in any way, I would always focus on readability - especially if it some open source code.

compiler/compiler.ts
```diff
@@ -0,0 +1,68 @@
...
+        newName += '.js';
+
+    const out = path.resolve('dist', newName);
```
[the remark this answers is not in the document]
cstrnt: any reason why you resolved this?

compiler/log.ts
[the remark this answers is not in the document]
QuiiBz: I just figured out why you wanted to use an `abstract` class - we don't want to instantiate `SpecLogger` mistakenly. Let's change it to this:
```suggestion
/**
 * Log a message in the console. Can specify a Level, which
 * will be INFO by default.
 *
 * @param message The message to log
 * @param level The level to log
 */
export default abstract class SpecLogger {

    static log(message: string, level: Level = Level.INFO) {
        
        console.log(chalk[level](message));
    }
}
```

compiler/compiler.ts
```diff
@@ -0,0 +1,69 @@
+import * as fs from 'fs';
+import * as ts from 'typescript';
+import * as path from 'path';
+import { variableNameTransformer } from './transformer';
+import { Level, log } from './log';
```
QuiiBz, line 5:
```suggestion
import SpecLogger, { Level } from './log';
```

compiler/compiler.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    fs.writeFileSync(outFilePath, result.outputText);
+};
+
+// Process all the files in the specs directory
+fs.readdir('specs', (err, files) => {
+
+    if(err) {
+
+        log('Could not find specs folder', Level.ERROR);
```
QuiiBz, line 48:
```suggestion
        SpecLogger.loc('Could not find specs folder', Level.ERROR);
```

compiler/compiler.ts
```diff
@@ -0,0 +1,69 @@
...
+        return;
+    }
+
+    const specs = files.filter(file => file !== '.DS_STORE');
+    log(`Processing ${specs.length} specs...`);
```
QuiiBz, line 53:
```suggestion
    SpecLogger.log(`Processing ${specs.length} specs...`);
```

compiler/compiler.ts
```diff
@@ -0,0 +1,69 @@
...
+        incomplete: ' ',
+    });
+
+    specs.forEach(spec => {
+
+        processSpec( path.join('specs', spec));
+        bar.tick({ spec });
+    });
+
+    log('Specs compiled successfully to /dist folder!', Level.SUCCESS);
```
QuiiBz, line 68:
```suggestion
    SpecLogger.log('Specs compiled successfully to /dist folder!', Level.SUCCESS);
```

compiler/compiler.ts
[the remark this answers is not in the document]
QuiiBz: I resolved this because I made the changes in https://github.com/withfig/autocomplete/pull/44/commits/cbeff291380b20272e976e9c5a212007aaef14dd.

PR #46, opened 2021-03-13
feat: move js specs to ts

cstrnt:
Feature

There are still some specs written in javascript

It ports all old javascript specs to typescript

**Additional info:**


specs/brew.ts
```diff
@@ -1,19 +1,20 @@
...
     servicesgenerators: {
         script: "brew services list | sed -e 's/ .*//' | tail -n +2",
-        postProcess: function (out) {
-            return out.split('\n').filter((line) => {
-                return !line.includes('unbound') && {
-                    name: line,
-                    type: "option"
-                }
-            })
+        postProcess: function (output) {
```
QuiiBz, line 4: I would go with arrow functions since we are not anymore in vanilla JS:
```suggestion
        postProcess: output => {
```

specs/brew.ts
```diff
@@ -1,19 +1,20 @@
...
+                    name: currentLine,
+                    type: 'option'
+                });
+            }, [])
         }
     }
 }
 
-var completionSpec = {
+const brewCompletionSpec = {
```
QuiiBz, line 17: You forgot to add the type:
```suggestion
const brewCompletionSpec: Fig.Spec = {
```

compiler/transformer.ts
```diff
@@ -1,49 +1,51 @@
...
-                }
-            }
-
-            return visitEachChild(node, visitor, context);
-        };
-
-        return visitNode(sourceFile, visitor);
+export const variableNameTransformer: TransformerFactory<SourceFile> = (
+  context
+) => {
```
QuiiBz, line 24: I would prefer inline this expression since there is only one parameter. Also, we can remove the parenthesis for the same reason.
```suggestion
export const variableNameTransformer: TransformerFactory<SourceFile> = context => {
```

compiler/transformer.ts
```diff
@@ -1,49 +1,51 @@
...
+  return (sourceFile) => {
```
QuiiBz, line 25: We can also remove the parenthesis here.

development/brew.ts
```diff
@@ -0,0 +1,194 @@
+const brewGenerators: Record<string, Fig.Generator> = {
+    servicesgenerators: {
```
QuiiBz, line 2: It should have a type of `Fig.Generator`.

development/brew.ts
```diff
@@ -0,0 +1,194 @@
...
+                    options: [
+                        {
+                            name: "--all",
+                            insertValue: "--all",
+                            description: "Start all services"
+                        }
+                    ],
+                    args: {
+                        variadic: true,
+                        generators: generators.servicesGenerator
```
QuiiBz, line 182: As defined above, the name of the generator should be `servicesgenerators`, not `generators.servicesGenerator`.

development/brew.ts
cstrnt: It's an Object of Generators. The person which created the spec wanted to bundle them in one object thus a Record with a key of type `string`and value `Fig.Generator`

mattschrage (approved):
Looks good! Can't wait to have Typescript support

PR #49, opened 2021-03-14
Add spec for `bundle` and all its sub commands

coreyja:

New spec for `Bundler` https://bundler.io/

There is no spec for `bundle`

There is a spec for `bundle`

**Additional info:**

This spec was generated from their MAN pages in a partially automated
way. I would love another set of eyes on everything!


specs/bundle.js
```diff
@@ -0,0 +1,538 @@
...
+    {
+      name: "exec",
+      description: "Execute a command in the context of the bundle",
+      options: [
+        {
+          name: ["--keep-file-descriptors"],
+          description: "Pass all file descriptors to the new process",
+        },
+      ],
+      args: { isCommand: true },
```
coreyja, line 161: Is this `isCommand` supported?

It seems be be working and is documented: https://withfig.com/docs/autocomplete/api#arg-object

But I don't see other specs with it currently and its failing the linter!

rolandtshen: Fixed thanks to @DannyAziz's PR!

falky97 (approved):
This is fantastic. Very well structured and commented. And thank you for including the empty arg objects when needed!! A great first spec and I can't wait to flesh this out more with some generators. Ruby community owes you a lot :)

PR #52, opened 2021-03-15
feat: isCommand field for Arg

DannyAziz:
Feature

isCommand not in Arg.js schema so the linter fails

Add isCommand in Arg.js so the linter passes :)

Additional info:


rolandtshen:
Thanks for doing this! I added a few more things to the Arg linter as well, should fix it up for the new properties

PR #53, opened 2021-03-17
(Docker compose): Global Options & -f generator & --env-file generator & remove unnecessary whitespace

DannyAziz:
Feature
No global options
All global options and a generator for `-f`, including modifying the existing `getServices` generator to handle the custom config set by `-f`
**Additional info:**


specs/docker-compose.js
```diff
@@ -1,5 +1,11 @@
 var getServices = {
-    script: "docker-compose config --services",
+    script: (context) => {
+        if (context.includes('-f')) {
+            let index = context.indexOf('-f');
+            return `docker-compose -f ${context[index + 1]} config --services`
+        }
```
rolandtshen, line 6: So this bit of code will only run when a -f option is also available. For example, in the `build` subcommand, add a -f option — otherwise, Fig's parsing will assume something's broken since it couldn't find -f, and just not show any suggestions at all. That option should be added everywhere this generator is called.

DannyAziz: Hey @rolandtshen, the `-f` option is a global option, I have defined it at the bottom (https://github.com/withfig/autocomplete/pull/53/files#diff-a6ad66cb36777cfc5141947a18592b3874ac6357b81fe1e01c7f94233b71523eR879) 

The current way I have it set up seems to be working for me

falky97 (approved):

falky97:
Lol. 90% of this was just removing "\n" haha. good stuff @DannyAziz

PR #54, opened 2021-03-18
Update git fetch

Jubeki:
This adds all options and args for the `git fetch` command.

The options `--recurse-submodules` and `--recurse-submodules-default` are currently not active, because the delimiter `=` is currently not working


rolandtshen:
Thanks a ton for adding more support to the git spec. It'll be well-used by tons of devs :)

PR #59, opened 2021-03-18
nvm

reubn:
Autocompletion for [`nvm`](https://github.com/nvm-sh/nvm)

N/A

N/A

**Additional info:**
This could be better by having a generator for node versions, but `nvm` isn't accessible due to the way Fig runs commands, so getting this data is hard (haven't explored why this isn't working too deeply)

### DIFF
--- a/specs/nvm.js
+++ b/specs/nvm.js
@@ -308,16 +308,20 @@ var completionSpec = {
     },
     {
       name: "cache",
-      subcommands: [
-        {
-          name: "dir",
-          description: "Display path to the cache directory for nvm"
-        },
-        {
-          name: "clear",
-          description: "Empty cache directory for nvm"
-        }
-      ],
+      args: {
+        suggestions: [
+          {
+            name: "dir",
+            description: "Display path to the cache directory for nvm",
+            type: "subcommand"
+          },
+          {
+            name: "clear",
+            description: "Empty cache directory for nvm",
+            type: "subcommand"
+          }
+        ],
+      },
     },
     {
       name: "set-colors",
--- a/specs/nvm.js
+++ b/specs/nvm.js
@@ -47,27 +47,27 @@ var colorCodes = {
 
 // options
 var noColors = {
-  name: ["--no-colors"],
+  name: "--no-colors",
   description: "Suppress colored output"
 }
 
 var noAlias = {
-  name: ["--no-alias"],
+  name: "--no-alias",
   description: "Suppress `nvm alias` output"
 }
 
 var silent = {
-  name: ["--silent"],
+  name: "--silent",
   description: "Silences stdout/stderr output",
 }
 
 var lts = {
-  name: ["--lts"],
+  name: "--lts",
   description: "Uses automatic LTS (long-term support) alias `lts/*`, if available.",
 }
 
 var ltsWithName = {
-  name: ["--lts="],
+  name: "--lts=",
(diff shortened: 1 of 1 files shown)


specs/nvm.js
```diff
@@ -0,0 +1,345 @@
...
+      ]
+    },
+    {
+      name: "alias",
+      description: "Set an alias named <name> pointing to <version>",
+      args: [
+        name,
+        version
+      ]
+    },
```
rolandtshen, line 268: Having two subcommands with the same name causes Fig to always default to the first one, which has pattern as an arg. We need to merge both subcommands into one that just supports 2 args. Something like:

```js
args: [
{
name: "pattern or name",
description: "pattern or name
},
{
name: "version",
isOptional: true
}
]

File: specs/nvm.js [hunk 2]
```diff
@@ -0,0 +1,345 @@
...
+  isVariadic: true
+}
+
+var pattern = {
+  name: "pattern"
+}
+
+var name = {
+  name: "name"
+}
```
rolandtshen, line 37: It's good to reuse code, but when an argument isn't used super often, it could be more readable to just have name: "name" as an argument inline. Just for future reference!

specs/nvm.js
reubn: The reason I did it as I have is that the first arg is not the same between the 2 versions of the `alias` command. The first type of `alias` command expects the first arg to be a `<pattern>`, while the second form expects the first arg to be a `<name>`. If we coalesce `<name>` and `<pattern>` into `<pattern or name>`,  then they can't have different descriptions or icons etc.: the semantic separation is lost. 

Coalescing as above also means that `<version>` will be suggested for the first form of `alias` that expects only 1 `<pattern>`. Here it would not be a valid arg (i.e. you cannot pass `<pattern> <version>`) - and so it should not be suggested.

`nvm --help` probably illustrates the requirements best:
```
nvm alias [<pattern>]                       Show all aliases beginning with <pattern>
  --no-colors                               Suppress colored output
nvm alias <name> <version>                  Set an alias named <name> pointing to <version>
``` 

I think the issue is that fig defaults to the first one, as both mine and your solution don't result in perfectly correct behaviour.

If you feel I've missed the point, or especially strongly that your suggestion is better, then I'm always happy to update the PR.

rolandtshen: Yes, this is a limitation with the current spec. With the original version, whether or not the user selects the first or second alias option, there will be no change in what gets entered to the terminal, since the arg objects aren't inserting any content anyways.
Screen Shot 2021-03-19 at 4 42 02 PM

I think it'd make for a better user experience if there were one alias in the autocomplete menu rather than two, and naming it something like "pattern or name" is still clear enough for them to understand what to input in that box. If they don't start typing the second argument out,  the `version` just won't show up anywhere on the terminal.

reubn: 👌

mattschrage:
@cstrnt @reubn do either of you have any idea why the linting is failing here? Would love to get this merged in!

QuiiBz:
Looks like the issue is coming from this line: https://github.com/withfig/autocomplete/blob/91aa7843b9d41333183cf6f5b244eeeef5dfb27b/scripts/eslint-plugin-fig-linter/rules/no-invalid-name.js#L21, where `name` is undefined. I can tackle this in another PR.

PR #61, opened 2021-03-19
Add jest spec

omichowdhury:


rolandtshen:
Woo! Thanks for contributing and for taming the linter :)

PR #62, opened 2021-03-19
working youtube-dl spec

berbaroovez:


rolandtshen:
Really nifty use of generators here, there's definitely a good use for pbpaste in other applications as well. Thanks for contributing!

PR #63, opened 2021-03-19
Change brew install subcommand argument generator

vicentube:
Bug fix
It works on intel macs but not on M1 macs since brew is installed in different path (/opt/homebrew)
using brew --prefix it gets the right homebrew installation path
**Additional info:**


specs/brew.js
```diff
@@ -34,7 +34,7 @@ var completionSpec = {
                 name: "formula",
                 description: "Formula or cask to install",
                 generators: {
-                    script: "ls -1 /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask/Casks",
+                    script: "homebrew_path=$(brew --prefix); ls -1 $homebrew_path/Library/Taps/homebrew/homebrew-core/Formula $homebrew_path/Library/Taps/homebrew/homebrew-cask/Casks",
```
rolandtshen, line 37: I tested this out on a non-M1 mac, and believe it should be `$homebrew_path/Library/Taps/homebrew/homebrew-core/Formula $homebrew_path/Library/Taps/homebrew/homebrew-cask/Casks`. `brew --prefix` is just returning usr/local for me, not usr/local/Homebrew

vicentube: I can see where is the problem: `brew --prefix` point to different points of the homebrew directory tree. But I solved it with `brew --repository` command. That command gives the same path than `brew --prefix` for M1 macs but the right one (`/usr/local/Homebrew`) for Intel macs. I sent a new commit with the change.

PR #64, opened 2021-03-19
Updated Yarn CLI specs

shoeblabs:
Spec Update

**Additional info:**
- Added description in all of the root subcommands.
- Few sub commands details are pending.


specs/yarn.js
```diff
@@ -1,3 +1,52 @@
...
+        if (out.trim() == "") {
+            return []
+        }
+
+        try {
+            let package = JSON.parse(out)
+            let dependencyScripts = package["dependencies"]
+            let devDependencyScripts = package["devDependencies"]
+            if (dependencyScripts || devDependencyScripts) {
+                return [...Object.keys(dependencyScripts), ...Object.keys(devDependencyScripts)]
```
rolandtshen, line 15: What gets returned here needs to be formatted as a [suggestion object](https://withfig.com/docs/autocomplete/api#suggestion-object). I think having these keys as the name, and adding a little 📦 icon would be perfect.

rolandtshen: Everything else looks great, ready to merge after this!

falky97 (approved):
Hey - thanks so much for adding this! Great additions to the spec, especially the generator.
Two notes: 
1. please always add error logging to any try catch statement
2. we can't assume that a package.json has a dependencies or devDependencies key... When I tested it this wasn't the case and your generator didn't work at all because we weren't accounting for it. Next time try and test on a few more types before merging! 
```
      const dependencyScripts = packageContent["dependencies"] || {};
      const devDependencyScripts = packageContent["devDependencies"] || {};
```

Regardless, still great for your first generator

shoeblabs:
> Hey - thanks so much for adding this! Great additions to the spec, especially the generator.
> Two notes:
> 
> 1. please always add error logging to any try catch statement
> 2. we can't assume that a package.json has a dependencies or devDependencies key... When I tested it this wasn't the case and your generator didn't work at all because we weren't accounting for it. Next time try and test on a few more types before merging!
> 
> ```
>       const dependencyScripts = packageContent["dependencies"] || {};
>       const devDependencyScripts = packageContent["devDependencies"] || {};
> ```
> 
> Regardless, still great for your first generator

Thanks for the suggestion 👍

PR #66, opened 2021-03-20
feat: improve dev workflow and add watch option

cstrnt:
Feature

The dev workflow (especially with typescript) is kinda clunky

There is a `watch` command which watches the `.ts` specs and transpiles them on save which greatly increases the developer experience.

**Additional info:**


.gitignore
```diff
@@ -7,3 +7,4 @@ build
 npm-debug.log
 .env
 .idea/
+.vscode
```
QuiiBz, line 10: We should add a newline to the end of the file to follow the [POSIX standards](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).
```suggestion
.vscode

```

package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "test": "tsc --noEmit",
     "build": "ts-node-script compiler/compiler.ts",
-    "autocomplete": "cp ./specs/*.js ~/.fig/autocomplete/"
+    "autocomplete": "cp ./specs/*.js ~/.fig/autocomplete/",
```
QuiiBz, line 8: Maybe change this script name to a more comprehensive one? We don't really know what this script does before reading the command. Something like `sync` could be better, @mattschrage what do you think of it?

cstrnt: That's actually the command which we proposed during the meeting.image

But I totally agree with you! The name should be more verbose

PR #68, opened 2021-03-20
Update `git commit` and  `git config`

Jubeki:
This adds all specs from the `git commit` and `git config` command.

I will be doing some more specs for git in the next weeks, but I have exams next week so this will have to wait a bit.

For generating the specs I am using something self written which can be used:
https://github.com/Jubeki/fig-git-fetcher

### DIFF
--- a/specs/git.js
+++ b/specs/git.js
@@ -498,7 +498,156 @@ var completionSpec = {
                             }
                         ],
                     },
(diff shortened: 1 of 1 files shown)


rolandtshen:
Thanks for filling in more options! Looks good

PR #71, opened 2021-03-23
[GITHUB]: Add Completion spec

j-f1:
While most useful as a simple `github` invocation, the GitHub Desktop CLI also allows opening arbitrary local repositories or cloning remote ones. I added some common Git providers as autocomplete options for the clone subcommand. GitHub `user/repo` slugs are also allowed, and it might be cool to guess the user’s username/org memberships to provide them as autocomplete options but that might not be worth the effort.


specs/github.js
```diff
@@ -0,0 +1,61 @@
...
+        {
+          name: ["--branch", "-b"],
+          description: "The branch to checkout after cloning",
+          args: { name: "branch" },
+        },
+      ],
+    },
+    {
+      name: "open",
+      insertValue: "open '{cursor}'",
```
rolandtshen, line 43: Did you mean for this to be `open '{cursor}'`, or `open {cursor}`? Having the {cursor} in quotes would insert `open '|'` into the terminal, with the | being the cursor, as opposed to `open |`.

j-f1: I put it with quotes so that pasting a filename with spaces would work but it seems terminals (at least macOS’s Terminal) automatically insert escape characters so this is unnecessary.

mattschrage:
Hey @j-f1 - sorry for the slow response on this. We've been porting over the completion specs to Typescript and I think your changes were caught in the middle.

If you just move `./specs/github.js` to `/dev/github.ts` and change `var completionSpec = ` to `export const completionSpec: Fig.Spec =` we can get this merged in!

dev/github.ts
```diff
@@ -0,0 +1,61 @@
...
+    {
+      name: "clone",
+      description: "Clone a repository",
+      insertValue: "clone ",
+      args: {
+        name: "url|slug",
+        suggestions: [
+          { name: "https://github.com/", icon: "fig://icon?type=git" },
+          { name: "https://gitlab.com/", icon: "fig://icon?type=git" },
+          { name: "https://bitbucket.com/", icon: "fig://icon?type=git" },
```
cstrnt, line 30: This is totally fine, but you could also add the icons of the websites itself. You can just use the url of the favicon for example.

j-f1: My concern is that e.g. the GitHub icon will not look good on a dark background. Maybe if there was a 🔗 or 🌐 icon that would work better?

cstrnt: [
Screenshot 2021-04-13 at 17 25 27
](url)

 I honestly think it looks pretty awesome with this one https://github.com/fluidicon.png

cstrnt (approved):
LGTM

mattschrage:
🎉

PR #76, opened 2021-03-29
feat: add custom eslint rules

cstrnt:
Adds custom eslint rules tailored for fig specs

There are no mechanisms to check the content of the specs

The eslint rules check whether the specs are well suited for use

**Additional info:**


scripts/eslint-plugin-fig/index.js
```diff
@@ -0,0 +1,60 @@
+const rmRfRegexp = /rm \-(?:f(?:(?: \-)?r)?|r(?: \-f|f)?)/;
```
QuiiBz, line 1: I would prefer an uppercase variable name since it's a global constant.

scripts/eslint-plugin-fig/index.js
```diff
@@ -0,0 +1,60 @@
...
+                            if (value.type === 'Literal' && value.value.endsWith('=')) {
+                                context.report({
+                                    node,
+                                    message: 'The name property must not include `=`',
+                                    fix: function (fixer) {
+                                        const [, end] = value.range;
+                                        return fixer.replaceTextRange([end - 2, end - 1], '');
+                                    },
+                                });
+                            }
```
QuiiBz, line 44: Hard to read with all theses `value` variables, maybe rename the first one to add clarity:
```suggestion
                            const currentNode = node.value;
                            if (currentNode.type === 'Literal' && currentNode.value.endsWith('=')) {
                                context.report({
                                    node,
                                    message: 'The name property must not include `=`',
                                    fix: function (fixer) {
                                        const [, end] = currentNode.range;
                                        return fixer.replaceTextRange([end - 2, end - 1], '');
                                    },
                                });
                            }
```

scripts/eslint-plugin-fig/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
QuiiBz, line 1: No need to set `keywords` and `scripts`, but we should fill `description` and `author`, and add `private: true` to make sure the package isn't published by accident.

QuiiBz (changes requested):

QuiiBz (approved):

PR #85, opened 2021-04-01
feat: add hyphen subcommand to git and cd spec

cstrnt:
New Spec

The `-` subcommand for `git checkout`and `cd` is not implemented

It implements the subcommand for both specs 

**Additional info:**


falky97:
Hey Tim - so the `-` is an **argument** not a **subcommand**.

So for both `cd` and `git checkout`, we will need to add the following to the arg object:
```
args: {
    ...,
    suggestions: [{name: "-", description: "switch to the last used folder" }]
}
```

I believe the prop name is `suggestions` however it could also be `additionalSuggestions`. You'll have to test


Slides 5 and 6 of this hack-night deck give you a good summary of how it works: https://docs.google.com/presentation/d/1B69F5mg0YPu7z6Y6VQkNgdWgPXCg8fVVHWcgu4BYEL4/edit#slide=id.gc60f6c760e_0_15

cstrnt:
I totally messed that one up (and it actually works this way as well). Will change that and keep the actual way in mind

falky97:
Looks good! Here is something I just wrote up about the difference between subcommands, options, and arguments. Definitely needs a look over, but high level stuff is there: https://forum.withfig.com/t/cli-tools-the-difference-between-a-command-subcommand-option-and-argument/56

PR #89, opened 2021-04-02
feat: add yarn icon for scripts from package.json

cstrnt:
Feature

When running `yarn <scriptname>` or `yarn run <scriptname>` the yarn icon doesn't appear like it does when invoking `npm run <scriptname`

The icon is displayed

**Additional info:**

### DIFF
--- a/dev/yarn.ts
+++ b/dev/yarn.ts
@@ -1,30 +1,30 @@
+const yarnGenerators = {
+  getScripts: {
+    script: "cat package.json",
+    postProcess: function (output) {
+      if (output.trim() == "") {
+        return [];
+      }
+      try {
+        const packageContent = JSON.parse(output);
+        const scripts = packageContent["scripts"];
+        if (scripts) {
+          return Object.keys(scripts).map((scriptName) => ({
+            name: scriptName,
+            icon: "https://yarnpkg.com/favicon-32x32.png",
+          }));
+        }
+      } catch (e) {}
+
+      return [];
+    },
+  } as Fig.Generator,
+};
+
 export const completionSpec: Fig.Spec = {
   name: "yarn",
   description: "Manage packages and run scripts",
-  args: [
-    {
-      generators: {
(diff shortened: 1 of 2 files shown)


dev/yarn.ts
```diff
@@ -1,30 +1,30 @@
...
+            name: scriptName,
+            icon: "https://yarnpkg.com/favicon-32x32.png",
+          }));
+        }
+      } catch (e) {}
+
+      return [];
+    },
+  } as Fig.Generator,
+};
```
QuiiBz, line 22: We shouldn't use type assertion, rather infer the type at the declaration level:
```suggestion
const yarnGenerators: Record<string, Fig.Generator> = {
  getScripts: {
    script: "cat package.json",
    postProcess: function (output) {
      if (output.trim() == "") {
        return [];
      }
      try {
        const packageContent = JSON.parse(output);
        const scripts = packageContent["scripts"];
        if (scripts) {
          return Object.keys(scripts).map((scriptName) => ({
            name: scriptName,
            icon: "https://yarnpkg.com/favicon-32x32.png",
          }));
        }
      } catch (e) {}

      return [];
    },
  },
};
```

QuiiBz (changes requested):

dev/yarn.ts
cstrnt: I did this before, but doing that causes autocompletion when typing `yarnGenerators.` not to work. When you use `string` as the key of a record it means it can be any `string` and it will not show you (atleast in vscode) completions for `getScripts`

QuiiBz: Nevermind, you're right. We could hard write the `getScripts` property like this to get the completion: `const yarnGenerators: { getScripts: Fig.Generator } = {`, but it's not very clean.

QuiiBz (approved):

mattschrage (approved):
lgtm!

PR #90, opened 2021-04-03
update fig and git

rolandtshen:

**Additional info:**


mattschrage (approved):
Looks good. Going forward though, try to only change 1 completion spec per PR. :)

Issue #87, opened 2021-04-02
React-Native Basic Commands Support


alexandruenache19:
I would love to have the autocomplete functionality for React-Native Basic Commands like react-native run-ios, react-native start, etc

alexnaiman:
Can I work on this?

mattschrage:
@alexnaiman I don't think anyone else has tackled `react` CLI tooling yet. We'd love your help :)

alexnaiman:
@mattschrage I think that this issue can be closed now

PR #91, opened 2021-04-03
feat: Added spec for react native cli and npx

alexnaiman:
Added spec for standalone `react-native`'s cli and also for when using it from `npx`(`npx react-native [command]`)
See #87 

**What is the current behaviour? (You can also link to an open issue here)**
No specs for `react-native`

**Additional info:**

Some of the generators created, fail sometimes because some of the commands used inside the `script` timeout. Unfortunately, I couldn't figure out exactly a way to avoid that

### DIFF
--- a/dev/npx.ts
+++ b/dev/npx.ts
@@ -1,0 +1,133 @@
+export const completionSpec: Fig.Spec = {
+  name: "npx",
+  description: "Execute binaries from npm packages",
+  subcommands: [
+    {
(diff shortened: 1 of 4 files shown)


mattschrage (approved):
There is is a 5 second timeout by default. I think we might make this configureable in the spec to give autocomplete maintainers more control. cc @falky97 

Otherwise, this looks great! Thanks for contributing :)

PR #92, opened 2021-04-06
Fix search path in brew install subcommand generator

vicentube:
Bug fix
Search path for formulas and casks in brew install subcommand generator is wrong for homebrew installations in new M1 macs. On M1 macs homebrew installs in /opt/homebrew while on Intel macs is in /usr/local/Homebrew 
By using the command "brew --repository", the right homebrew installation path can be used for any mac.
**Additional info:**

### DIFF
--- a/dev/brew.ts
+++ b/dev/brew.ts
@@ -55,7 +55,7 @@ export const completionSpec: Fig.Spec = {
         description: "Formula or cask to install",
         generators: {
           script:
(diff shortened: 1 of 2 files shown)


falky97:
There seems to be a slight performance hit with this of ~20ms. Considering brew is already a little slow, I think we are okay with this in order to support m1. 
Great catch, thanks @vicentube !!

image

falky97:
@vicentube I have also added you as a member of the @withfig organization so you can add the Fig logo to your list of orgs on your github profile 😁 
https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/publicizing-or-hiding-organization-membership#:~:text=Changing%20the%20visibility%20of%20your,the%20icon%20for%20your%20organization.

PR #93, opened 2021-04-06
feat: add lerna command

cstrnt:
Feature

There is no completion spec for [lerna](https://github.com/lerna/lerna)

It adds the completion spec 🥳 

**Additional info:**

### DIFF
--- a/dev/lerna.ts
+++ b/dev/lerna.ts
@@ -40,8 +40,7 @@ const getAllScriptsFromPackages: Fig.Generator = {
 
     return Array.from(new Set(scripts)).map((scriptName) => ({
       name: scriptName,
-      icon:
-        "https://img.pngio.com/publishing-to-npm-from-kentcdodds-on-eggheadio-npm-png-800_800.png",
+      icon: "fig://icon?type=npm",
     }));
   },
 };
@@ -76,8 +75,7 @@ const npmClientOption: Fig.Option = {
     suggestions: [
       {
         name: "npm",
-        icon:
-          "https://img.pngio.com/publishing-to-npm-from-kentcdodds-on-eggheadio-npm-png-800_800.png",
+        icon: "fig://icon?type=npm",
       },
       {
         name: "yarn",
@@ -467,8 +465,7 @@ export const completion: Fig.Spec = {
         isOptional: true,
         suggestions: SEMVER_NOUNS.map((versionName) => ({
           name: versionName,
-          icon:
-            "https://img.pngio.com/publishing-to-npm-from-kentcdodds-on-eggheadio-npm-png-800_800.png",
+          icon: "fig://icon?type=npm",
         })),
       },
       options: [
--- a/dev/lerna.ts
+++ b/dev/lerna.ts
@@ -1,0 +1,787 @@
+const SPLIT_CHAR = "END";
+
+const getPackages: Fig.Generator = {
+  script: "lerna ls",
+  postProcess: (output) =>
+    output.split("\n").map((packageName) => ({
+      name: packageName,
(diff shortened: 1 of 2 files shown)


dev/lerna.ts
```diff
@@ -0,0 +1,787 @@
+const SPLIT_CHAR = "END";
+
+const getPackages: Fig.Generator = {
+  script: "lerna ls",
+  postProcess: (output) =>
+    output.split("\n").map((packageName) => ({
```
QuiiBz, line 6: Can be simplified by using the same `name` variable name:
```suggestion
    output.split("\n").map(name => ({
```

dev/lerna.ts
```diff
@@ -0,0 +1,787 @@
...
+      name: packageName,
```
QuiiBz, line 7: So this would become just:
```suggestion
      name,
```

dev/lerna.ts
cstrnt: I agree with you that it makes it less code but I'm always struggling whether verbose code is better or *cleaner* code

QuiiBz: In this case, I feel like it's both cleaner and better code since it's easy to understand and easier to read.

dev/lerna.ts
```diff
@@ -0,0 +1,787 @@
...
+    let scripts: string[] = [];
+    packages.forEach((packageContent) => {
+      const parsed = JSON.parse(packageContent);
+      scripts = scripts.concat(Object.keys(parsed["scripts"]));
+    });
+
+    return Array.from(new Set(scripts)).map((scriptName) => ({
+      name: scriptName,
+      icon:
+        "https://img.pngio.com/publishing-to-npm-from-kentcdodds-on-eggheadio-npm-png-800_800.png",
```
mattschrage, line 44: I know this isn't documented, but there is actually an `npm` icon builtin. `fig://icon?type=npm`

I really need to make a list.

dev/lerna.ts
```diff
@@ -0,0 +1,787 @@
...
+const npmClientOption: Fig.Option = {
+  name: "--npm-client",
+  description: "The exececutable to run the script with",
+  args: {
+    name: "client",
+    suggestions: [
+      {
+        name: "npm",
+        icon:
+          "https://img.pngio.com/publishing-to-npm-from-kentcdodds-on-eggheadio-npm-png-800_800.png",
```
mattschrage, line 80: Same as above.

specs/lerna.js
```diff
@@ -0,0 +1,717 @@
...
+        for (var s, i = 1, n = arguments.length; i < n; i++) {
+            s = arguments[i];
+            for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
+                t[p] = s[p];
+        }
+        return t;
+    };
+    return __assign.apply(this, arguments);
+};
+var __spreadArray = (this && this.__spreadArray) || function (to, from) {
```
mattschrage, line 12: Are these utility functions computed at compile time or runtime?

Also - does it make sense to include this in the dev environment by default so every spec doesn't need to reimplement?

mattschrage (changes requested):
Just update the npm icon in a few places + a few clarifying questions!

dev/lerna.ts
cstrnt: That would be awesome!

specs/lerna.js
cstrnt: I don't think it is possible to add them to the dev environment as we are not bundling them, they just exist next to each other. 
And they are interpreted at run time because it's JS I think

PR #111, opened 2021-04-09
feat: remove inquirer and create cli in bash

cstrnt:
Feature

The `create-boilerplate` command uses `inquirer` which bloats the repo

The `create-boilerplate` command is written in bash and inquirer is removed as a dependency

**Additional info:**


falky97:
@cstrnt So this is good however I made a few tweaks to make this nicer:
1. I use a heredoc for the boilerplate code. It's a really nice, easy syntax. See comment in the script + this link: https://linuxhint.com/bash-heredoc-tutorial/
2. When getting the user's path in an npm script, **always use $INIT_CWD NOT $(pwd)**. Init_cwd is set to a project's root directory whereas pwd is the user's shell's current working directory. If the user is one directory deep, the command will fail when using pwd. More about init_cwd: https://allalmohamedlamine.medium.com/how-to-reference-project-root-directory-in-an-npm-script-6802adfa12e2
3. I updated some of the copy. When writing copy, think about if you were the user, what would you want? What would look nice? What's the best experience here.

PR #118, opened 2021-04-10
[FIG] feat: only transpile dirty specs in dev mode

cstrnt:
Feature

All files are transpiled in dev mode, this doesn't scale well

Only the files which changes will be transpiled

**Additional info:**

This PR is based on #117, because the file structure changes


mattschrage (approved):
🎉

PR #120, opened 2021-04-12
Kubectl resource generators

DannyAziz:
Feature
No generators for individual resources
Generators for resources & applied to subcommands
**Additional info:**


falky97 (approved):
Looks good! I'm not a kubectl savage but I trust you are getting this right.

One note: perhaps look at putting all of your shared generators and/or args into an object called something like `sharedArgs`. Think it will make things a little cleaner

PR #122, opened 2021-04-13
[FIG] feat: add new eslint rules

cstrnt:
Feature

The `name` prop isn't validated for subcommands and options

There are two new eslint rules which check and validate the `name` prop

**Additional info:**
There are a lot of failing specs now

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -174,8 +174,7 @@ export const completionSpec: Fig.Spec = {
       description: "Pass a config parameter to the command",
     },
     {
-      name: "--exec-path[=<path>]",
-      insertValue: "--exec-path",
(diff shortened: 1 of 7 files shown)


falky97 (changes requested):
Everything other than the `no-invalid-options` is good to merge. 

For the moment, can we take the `no-invalid-option.js`, put it into a separate PR. It turns out that multi-character options with just one dash can be valid. This is popular in golang. e.g -help does not mean -h -e -l -p. We shouldn’t be saying that is invalid. So i want to add a thing to the completion spec that tells the linter and the parser whether this spec takes multi-char single dash options

Finally, before we do this merge, can we make sure all the completion specs are fixed so they are valid.

PR #126, opened 2021-04-14
[NPM]: feat: npm install autosuggests from npms.io

DannyAziz:
Feature
Npm install returns packages that are already in the package.json
Uses`npm search` to search as you are typing
**Additional info:**
Screen Recording 2021-04-14 at 21 48 38

### DIFF
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -1,3 +1,26 @@
+const searchGenerator = {
+  script: function (context) {
+    if (context[context.length - 1] === "") return "";
+    const searchTerm = context[context.length - 1];
+    return `npm search ${searchTerm}`;
+  },
+  postProcess: function (out) {
+    const allLines: Array<Record<string, string>> = out
+      .split("\n")
+      .filter(
+        (line) =>
+          !line.includes(
+            '"NAME | DESCRIPTION | AUTHOR | DATE | VERSION | KEYWORDS '
+          )
+      )
+      .map((item) => item.split("|")[0].trim());
+    return allLines;
+  },
+  trigger: function () {
+    return true;
+  },
+};
+
 export const completionSpec: Fig.Spec = {
   name: "npm",
   description: "Node package manager",
@@ -5,46 +28,17 @@ export const completionSpec: Fig.Spec = {
     {
       name: "install",
       description: "",
-      args: [
-        {
-          generators: {
-            script: "cat package.json",
-            postProcess: function (out) {
-              if (out.trim() === "") {
-                return [];
-              }
-              try {
-                const packageConten = JSON.parse(out);
-                const dependencies = packageConten["dependencies"];
-                if (dependencies) {
-                  const dps = Object.keys(dependencies);
-                  return dps.map((pkg) => {
-                    const scope = pkg.indexOf("/") + 1;
-                    if (scope !== -1) {
-                      pkg = pkg.substring(scope);
-                    }
-                    const version = pkg.indexOf("@");
-                    if (version !== -1) {
-                      pkg = pkg.substring(version);
-                    }
-                    return {
-                      name: pkg,
-                      icon: `fig://icon?type=file`,
-                      description: "dependency file",
-                    };
-                  });
-                }
-              } catch (e) {}
-              return [];
-            },
-          },
-        },
-      ],
+      args: {
+        name: "package",
+        generators: searchGenerator,
+        debounce: true,
+        variadic: true,
+      },
       options: [
         {
           name: ["-P", "--save-prod"],
           description:
-            "Package will appear in your `dependencies`. This is thedefault unless `-D` or `-O` are present.",
+            "Package will appear in your `dependencies`. This is the default unless `-D` or `-O` are present.",
         },
         {
           name: ["-D", "--save-dev"],
@@ -140,7 +134,16 @@ export const completionSpec: Fig.Spec = {
     { name: "help", description: "search npm help documentation" },
     { name: "help-search", description: "get help on npm" },
     { name: "hook", description: "manage registry hooks" },
-    { name: "i", description: "install local package" },
+    {
+      name: "i",
+      description: "install local package",
+      args: {
+        name: "package",
+        generators: searchGenerator,
+        debounce: true,
+        variadic: true,
+      },
+    },
     {
       name: "install-ci-test",
       description: "install a project with a clean slate and run tests",
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -6,7 +6,11 @@ const searchGenerator: Fig.Generator = {
   },
   postProcess: function (out) {
     return JSON.parse(out).results.map(
-      (item) => item.package.name
+      (item) =>
+        ({
+          name: item.package.name,
+          description: item.package.description,
+        } as Fig.Suggestion)
     ) as Fig.Suggestion[];
   },
   trigger: function () {
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -1,20 +1,13 @@
-const searchGenerator = {
+const searchGenerator: Fig.Generator = {
   script: function (context) {
     if (context[context.length - 1] === "") return "";
     const searchTerm = context[context.length - 1];
-    return `npm search ${searchTerm}`;
+    return `curl -s -H "Accept: application/json" "https://api.npms.io/v2/search?q=${searchTerm}&size=250"`;
   },
   postProcess: function (out) {
-    const allLines: Array<Record<string, string>> = out
-      .split("\n")
-      .filter(
-        (line) =>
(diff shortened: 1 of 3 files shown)


QuiiBz:
Really cool idea! I don't know why but it's a lot slower for me, the `npm search` command takes ages. Any idea why?

dev/npm.ts
```diff
@@ -1,50 +1,44 @@
...
+          )
+      )
+      .map((item) => item.split("|")[0].trim());
+    return allLines;
+  },
+  trigger: function () {
+    return true;
+  },
+};
+
```
QuiiBz, line 23: We should add the `Fig.Generator` type to prevent mistakes, so we will need to cast the return of the `postProcess` function to `Fig.Suggestion[]`:
```suggestion
const searchGenerator: Fig.Generator = {
  script: function (context) {
    if (context[context.length - 1] === "") return "";
    const searchTerm = context[context.length - 1];
    return `npm search ${searchTerm}`;
  },
  postProcess: function (out) {
    return out
      .split("\n")
      .filter(
        (line) =>
          !line.includes(
            '"NAME | DESCRIPTION | AUTHOR | DATE | VERSION | KEYWORDS '
          )
      )
      .map((item) => item.split("|")[0].trim()) as Fig.Suggestion[];
    return allLines;
  },
  trigger: function () {
    return true;
  },
};
```

cstrnt:
Hey @DannyAziz you can use `npm search --json` which makes the string splitting stuff much easier, because you can just parse the json and look for the`name` prop :)

DannyAziz:
Hey @cstrnt `--json` seems to fail for me every time I use it with an error:

```
npm ERR! cb() never called!

npm ERR! This is an error with npm itself. Please report this error at:
npm ERR!     <https://github.com/npm/cli/issues>

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/danny/.npm/_logs/2021-04-15T10_15_41_562Z-debug.log
```

cstrnt:
@DannyAziz what npm version do you use? (`npm -v`) ?

QuiiBz:
Having the same error with Node v15.9.0 and NPM v7.5.3

cstrnt:
https://github.com/npm/cli/pull/2845
It got fixed in version 7.6.2 and works with npm ~6.x

falky97:
Hey guys - so this generator is good, however, it would be best to switch over to running a `curl` request in our script hitting this API: https://api-docs.npms.io/. 

Why?
We want generators that are 1. fast and 2. work on everyone's device. We can't assume that an end user has a specific version of a CLI installed (at least not until we version specs). Therefore, we always need a solution that works for everyone.

In general, I'd agree with Tim that using `--json` is nicer for parsing, however as it's not supported everywhere, we shouldn't do it. 

https://api-docs.npms.io/ is fast and works for everyone and so we should switch to it. Not gonna merge this PR until that's done

DannyAziz:
@falky97 Bump

falky97:
Code looks good, either @mattschrage or I will test later today / tomorrow
Two concerns:
1. npm install should ideally rank commands in the local package.json higher up should it not? I guess this is annoying to mix in a slow generator with a fast local json parse of the package.json... But we should think a little more about how to do this
2. Why is the check PR / danger test not working? (haven't had a chance to look into it)

mattschrage:
1. Does anyone really install local packages individually? Seems like the usual workflow is running `npm install` in this case. Totally could be wrong.

2. The `danger.js` workflow failing is not an issue with this PR.  We just need to add a Github token for danger to use.

falky97:
Yeah @matt probably fair re npm local installs. The only times I have done it for local files was when I was updating the version number. Main workflow is installing foreign packages

cstrnt:
# Overview
 
## dev/npm.ts:
### Info:
**Script:**
`cat package.json`
**postProcess(function):**
```typescript
 function (out) {
              if (out.trim() === "") {
                return [];
              }
              try {
                const packageConten = JSON.parse(out);
                const dependencies = packageConten["dependencies"];
                if (dependencies) {
                  const dps = Object.keys(dependencies);
                  return dps.map((pkg) => {
                    const scope = pkg.indexOf("/") + 1;
                    if (scope !== -1) {
                      pkg = pkg.substring(scope);
                    }
                    const version = pkg.indexOf("@");
                    if (version !== -1) {
                      pkg = pkg.substring(version);
                    }
                    return {
                      name: pkg,
                      icon: `fig://icon?type=file`,
                      description: "dependency file",
                    };
                  });
                }
              } catch (e) {}
              return [];
            }
```

**Script:**
`until [[ -f package.json ]] || [[ $PWD = '/' ]]; do cd ..; done; cat package.json`
**postProcess(function):**
```typescript
 function (out) {
              if (out.trim() == "") {
                return [];
              }

              try {
                const packageContent = JSON.parse(out);
                const scripts = packageContent["scripts"];
                const figCompletions = packageContent["fig"];

                if (scripts) {
                  const keys = Object.keys(scripts).map((key) => {
                    return Object.assign(
                      {},
                      { icon: "fig://icon?type=npm" },
                      figCompletions[key],
                      { name: key, insertValue: key }
                    ); // ensure that name and insertValue are defined by "scripts" dict
                  });
                  return keys;
                }
              } catch (e) {}

              return [];
            }
```

**Script:**
`cat package.json`
**postProcess(function):**
```typescript
 function (out) {
              if (out.trim() === "") {
                return [];
              }
              try {
                const packageContent = JSON.parse(out);
                const dependencies = packageContent["dependencies"];
                if (dependencies) {
                  const dps = Object.keys(dependencies);
                  return dps.map((pkg) => {
                    const scope = pkg.indexOf("/") + 1;
                    if (scope !== -1) {
                      pkg = pkg.substring(scope);
                    }
                    const version = pkg.indexOf("@");
                    if (version !== -1) {
                      pkg = pkg.substring(version);
                    }
                    return {
                      name: pkg,
                      icon: `fig://icon?type=file`,
                      description: "dependency file",
                    };
                  });
                }
              } catch (e) {}
              return [];
            }
```

**Script:**
`cat package.json`
**postProcess(function):**
```typescript
 function (out) {
              if (out.trim() === "") {
                return [];
              }
              try {
                const packageContent = JSON.parse(out);
                const dependencies = packageContent["dependencies"];
                if (dependencies) {
                  const dps = Object.keys(dependencies);
                  return dps.map((pkg) => {
                    const scope = pkg.indexOf("/") + 1;
                    if (scope !== -1) {
                      pkg = pkg.substring(scope);
                    }
                    const version = pkg.indexOf("@");
                    if (version !== -1) {
                      pkg = pkg.substring(version);
                    }
                    return {
                      name: pkg,
                      icon: `fig://icon?type=file`,
                      description: "dependency file",
                    };
                  });
                }
              } catch (e) {}
              return [];
            }
```

dev/npm.ts
```diff
@@ -1,50 +1,37 @@
+const searchGenerator: Fig.Generator = {
+  script: function (context) {
+    if (context[context.length - 1] === "") return "";
+    const searchTerm = context[context.length - 1];
+    return `curl -s -H "Accept: application/json" "https://api.npms.io/v2/search?q=${searchTerm}&size=250"`;
+  },
+  postProcess: function (out) {
+    return JSON.parse(out).results.map(
+      (item) => item.package.name
+    ) as Fig.Suggestion[];
```
cstrnt, line 10: The json also includes the description. I think this would be extremely nice if we could also include it

```suggestion
    return JSON.parse(out).results.map(
      (item) =>
        ({
          name: item.package.name,
          description: item.package.description,
        } as Fig.Suggestion)
    ) as Fig.Suggestion[];
```

dev/npm.ts
```diff
@@ -143,7 +130,16 @@ export const completionSpec: Fig.Spec = {
...
+    {
+      name: "i",
+      description: "install local package",
+      args: {
+        name: "package",
+        generators: searchGenerator,
+        debounce: true,
+        variadic: true,
+      },
+    },
```
cstrnt, line 142: Why is this a separate command? Shouldn't it be the same as install? The name prop can take an array as well!

cstrnt (changes requested):

dev/npm.ts
QuiiBz: The `Fig.Suggestion[]` should be inferred by typescript since we cast above in the map to a `Fig.Suggestion`, isn't it?

cstrnt: Unfortunately it isn't. Atleast I didn't get autocompletion in vscode

dev/npm.ts
falky97: Nope this is correct. Name prop of a subcommand cannot currently take an array. However, it could be interesting to add this as a feature.

falky97 (approved):

PR #133, opened 2021-04-16
Kubectl - Context & rollout generators

DannyAziz:
feature
No generators and missing args & options for `config` and `rollout` subcommands
- Adds missing args & opts for `config` and `rollout` subcommands
- Adds generators where appropriate for `config` and `rollout` subcommands
- Refactors all shared args into a `sharedArgs` const
**Additional info:**
Screen Recording 2021-04-16 at 10 37 12

### DIFF
--- a/dev/kubectl.ts
+++ b/dev/kubectl.ts
@@ -1662,6 +1662,9 @@ export const completionSpec: Fig.Spec = {
         {
           name: "deployment",
           description: "Create a deployment with the specified name.",
+          args: {
+            name: "NAME",
+          },
           options: [
             {
               name: ["--allow-missing-template-keys"],
--- a/dev/kubectl.ts
+++ b/dev/kubectl.ts
@@ -938,7 +938,16 @@ export const completionSpec: Fig.Spec = {
       name: "config",
       description:
         'Modify kubeconfig files using subcommands like "kubectl config set current-context my-context"',
-      options: [],
+      options: [
+        {
+          name: "--kubeconfig",
+          insertValue: "--kubeconfig=",
+          args: {
+            name: "path",
+            template: "filepaths",
+          },
+        },
+      ],
       subcommands: [
         {
           name: "current-context",
--- a/dev/kubectl.ts
+++ b/dev/kubectl.ts
@@ -41,6 +41,13 @@ const sharedArgs = {
       splitOn: "\n",
     },
   },
+  listDeployments: {
+    name: "Deployments",
+    generators: {
+      script: "kubectl get deployments.apps -o name",
+      splitOn: "\n",
+    },
+  },
 };
 
 export const completionSpec: Fig.Spec = {
@@ -3735,6 +3742,7 @@ export const completionSpec: Fig.Spec = {
         {
           name: "history",
           description: "View previous rollout revisions and configurations.",
+          args: sharedArgs.listDeployments,
           options: [
             {
               name: ["--allow-missing-template-keys"],
@@ -3784,6 +3792,7 @@ export const completionSpec: Fig.Spec = {
         {
           name: "pause",
           description: "Mark the provided resource as paused",
+          args: sharedArgs.listDeployments,
           options: [
             {
               name: ["--allow-missing-template-keys"],
@@ -3913,6 +3922,7 @@ export const completionSpec: Fig.Spec = {
         {
           name: "status",
           description: "Show the status of the rollout.",
+          args: sharedArgs.listDeployments,
           options: [
             {
               name: ["-f", "--filename"],
@@ -3947,6 +3957,51 @@ export const completionSpec: Fig.Spec = {
             {
               name: ["-w", "--watch"],
               description: "Watch the status of the rollout until it's done.",
+              args: {},
+            },
+          ],
+          subcommands: [],
+        },
+        {
+          name: "undo",
+          description: "Rollback to a previous rollout.",
+          args: sharedArgs.listDeployments,
+          options: [
+            {
+              name: ["--to_revision=revision"],
+              insertValue: "--to_revision=",
+              args: {},
+            },
+            {
+              name: "--dry-run=strategy",
+              insertValue: "--dry-run=",
+              args: {
+                name: "Strategy",
+                suggestions: ["none", "client", "server"],
+              },
+            },
+            {
+              name: ["-f", "--filename"],
+              description:
+                "Filename, directory, or URL to files identifying the resource to get from a server.",
+              args: {},
+            },
+            {
+              name: ["-k", "--kustomize"],
+              description:
+                "Process the kustomization directory. This flag can't be used together with -f or -R.",
+              args: {},
+            },
+            {
+              name: ["-R", "--recursive"],
+              description:
+                "Process the directory used in -f, --filename recursively. Useful when you want to manage related manifests organized within the same directory.",
+              args: {},
+            },
+            {
+              name: ["--timeout"],
+              description:
+                "The length of time to wait before ending watch, zero means never. Any other values should contain a corresponding time unit (e.g. 1s, 2m, 3h).",
               args: {},
             },
           ],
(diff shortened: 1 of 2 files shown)


dev/kubectl.ts
```diff
@@ -1,28 +1,52 @@
...
+        return `kubectl get ${resourceType} -o custom-columns=:.metadata.name`;
+      },
+      splitOn: "\n",
+    },
+  },
+  listKubeConfContexts: {
+    name: "Context",
+    generators: {
+      script: function (context) {
+        console.log(context);
```
cstrnt, line 32: Please remove this

dev/kubectl.ts
```diff
@@ -1,28 +1,52 @@
...
+        if (context.includes("--kubeconfig")) {
+          const index = context.indexOf("--kubeconfig");
```
cstrnt, line 34: You could define the const before the if and then use it in the if statement like this `if(index !== -1)`

dev/kubectl.ts
```diff
@@ -1,28 +1,52 @@
...
-  generators: {
-    script: "kubectl get pods --field-selector=status.phase=Running -o name",
-    splitOn: "\n",
+const sharedArgs = {
+  resourcesArg: {
+    name: "Resource Type",
+    generators: {
+      script: "kubectl api-resources -o name",
+      splitOn: "\n",
+    },
```
cstrnt, line 9: You could type all those args using `{ /** content here */ } as Fig.Arg` to make it more type safe!

cstrnt (changes requested):
This is just some high level code review, I have no clue about k8s 🤣

dev/kubectl.ts
mattschrage: 👍

dev/kubectl.ts
```diff
@@ -1003,12 +1036,34 @@ export const completionSpec: Fig.Spec = {
...
           description: "Sets a context entry in kubeconfig",
+          args: sharedArgs.listKubeConfContexts,
           options: [
             {
               name: ["--current"],
               description: "Modify the current context",
               args: {},
             },
+            {
+              name: ["--cluster=cluster_nickname"],
```
mattschrage, line 1047: As I mentioned in Slack:
> `name` is used for parsing, so it needs to match the flag directly. (In this case either --username or --password) If you want to change how the suggestion appears, set the displayName property.

Should look something like this.
```
name: "--cluster",
displayName: "--cluster=cluster_nickname"
```

mattschrage (changes requested):
Make sure that the `name` field for options exactly match the flag. 

❌ `name: "--flag=parameter"`  
✅ `name: "--flag"`

If you want to update the display text, use `displayName`

dev/kubectl.ts
DannyAziz: Playing around with it, I kinda prefer how fig displays this if I put `cluster_nickname` as the name of the arg instead
```
{
  name: ["--cluster"],
  arg: {
   name: 'cluster_nickname'
  }
}
```


https://user-images.githubusercontent.com/3679064/116064430-e9e2a680-a67d-11eb-933c-532c02d9ea48.mov

Whereas this is how it looks when using `displayName` instead of the `arg`

https://user-images.githubusercontent.com/3679064/116064572-0c74bf80-a67e-11eb-85f7-bce2839c828b.mov


I guess, we could do both?

DannyAziz:
# Update
feature

- Missing Args
- Missing Options
- Missing Subcommands
- Options and Subcommands under the wrong commands

- `sharedArgs` const for arguments that are shared across commands
- `sharedOpts` const for options that are shared across commands
- All options have the correct `insertValue` and have `args` if they need them (For example, many options are flags where an `arg` is not necessary)
- Generators for general resources and specific generators for pods, clusters, and containers

**Additional info:**
- `isOptional` is used pretty heavily here, atm this doesn't affect parsing but I've written this spec in mind that `isOptional` will affect parsing in the future (After a convo with Matt & Brendan on Slack)
- Still some pending TODO's but the current changes are 90% of the work and the pending TODO's will need minimal review IMO

dev/kubectl.ts
```diff
@@ -1,31 +1,250 @@
 // TODO: Handle if not connected to a k8s cluster
-const resourcesArg = {
-  name: "Resource Type",
-  generators: {
-    script: "kubectl api-resources -o name",
-    splitOn: "\n",
+// TODO: Handle if no resources found
+const scripts = {
```
cstrnt, line 3: Could you please add a comment indicating that those scripts are used for "internal" stuff? I think it's easy to confuse them with the `script` property

cstrnt (commented):
Such a huge PR, congratulations Danny

dev/kubectl.ts
```diff
@@ -1,31 +1,250 @@
...
+    generators: {
+      script: function (context) {
+        const lastInput = context[context.length - 1];
+        if (lastInput.includes("/")) {
+          return scripts.typeWithName(lastInput.replace("/", ""));
+        }
+        return scripts.types;
+      },
+      splitOn: "\n",
+      trigger: function () {
```
falky97, line 94: I think change this to trigger: "/". This will cause the generator to re-run only when a slash is added, not on every single keystroke. More: https://withfig.com/docs/autocomplete/api#generator-object

dev/kubectl.ts
```diff
@@ -1,31 +1,250 @@
 // TODO: Handle if not connected to a k8s cluster
-const resourcesArg = {
-  name: "Resource Type",
-  generators: {
-    script: "kubectl api-resources -o name",
-    splitOn: "\n",
+// TODO: Handle if no resources found
```
falky97, line 2: Possible to do this before we merge in? Should be as simple as the post process function checking if the output had an error message and then return an empty array

dev/kubectl.ts
```diff
@@ -1,31 +1,250 @@
...
+const scripts = {
+  types: "kubectl api-resources -o name",
+  typeWithName: function (type) {
+    return `kubectl get ${type} -o custom-columns=:.metadata.name | sed '/^[[:space:]]*$/d' | sed -e 's~^~pods/~;'`;
```
falky97, line 6: Rather than user sed, is it possible to put the parsing logic in javascript? We generally prefer to rely on the shell as little as we can as there is always the potential someone has some weird version of a CLI tool like sed. Whereas javascript is always going to be the same

dev/kubectl.ts
```diff
@@ -1,31 +1,250 @@
...
+            i === "pod" ||
+            i.includes("pod/")
+        );
+        const podName = context[podIndex].includes("/")
+          ? context[podIndex]
+          : `${context[podIndex]} + ${context[podIndex + 1]}`;
+        return `kubectl get ${podName} -o json`;
+      },
+      postProcess: function (out) {
+        return JSON.parse(out).spec.containers.map((item) => ({
```
falky97, line 130: I don't know if this is relevant here, but in general, want to add `icon: "fig://icon?type=png"` to the argument suggestions you return? This will give a nice kubernetes icon

falky97: API documented here: https://withfig.com/docs/autocomplete/reference/icon-api

falky97 (changes requested):
Made a few comments throughout. Let's get on a call and go through this too

dev/kubectl.ts
mattschrage: Yeah, I agree. Better to just use the `name` property in the Argument Object.

dev/kubectl.ts
DannyAziz: Here's where I use this:
```
  typeOrTypeSlashName: {
    name: "TYPE | TYPE/NAME",
    generators: {
      script: function (context) {
        const lastInput = context[context.length - 1];
        if (lastInput.includes("/")) {
          return scripts.typeWithName(lastInput.replace("/", ""));
        }
        return scripts.types;
      },
      splitOn: "\n",
      trigger: "/",
    },
  },
 ```
 
I'm not sure `postProcess` would help here as you would need access to the user's input to replicate the same behaviour as above which would return `pods/foobar` if the user has a `/` in their input

Perhaps a custom generator?

PR #135, opened 2021-04-21
[HEROKU] feat: add autogenerated heroku spec

cstrnt:
Feature

The current heroku cli spec is lacking a lot of subcommands

This is an autogenerated spec created from the source code

**Additional info:**

### DIFF
--- a/dev/heroku.ts
+++ b/dev/heroku.ts
@@ -38,7 +38,7 @@ export const spec: Fig.Spec = {
           args: {},
         },
         {
-          name: "--app",
+          name: ["-a", "--app"],
           description: "app to run command against",
           args: {
             name: "app",
@@ -77,7 +77,7 @@ export const spec: Fig.Spec = {
           description: "watch add-on creation status and exit when complete",
         },
         {
-          name: "--app",
+          name: ["-a", "--app"],
           description: "app to run command against",
           args: {
             name: "app",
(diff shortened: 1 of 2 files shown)


mattschrage (commented):
Is there anyway to get the short flags as well. Eg.  `-a` for `--app`?

cstrnt:
Sure, I'll try to figure out how it's implemented in the code and I'll try to implement it as well

QuiiBz:
What a huge CLI, congrats on the huge work!

dev/heroku.ts
```diff
@@ -1,122 +1,7079 @@
...
+      name: "domains:remove",
+      description: "remove a domain from an app",
+      args: [
+        {
+          name: "hostname",
+        },
+      ],
+      options: [
+        {
+          name: ["-h", "--help"],
```
mattschrage, line 561: Seems like the `--help` flags all take arguments? Is this expected?

mattschrage (approved):
This looks great @cstrnt! Thanks for tackling `heroku` - it's a beast of a CLI and I use it all the time.

dev/heroku.ts
cstrnt: Nah, you're right. Just fixed this

QuiiBz (approved):

PR #139, opened 2021-04-23
[FASTLY] Create completion spec

kailan:
The spec is generated by a custom tool. The source and documentation is available at https://github.com/kailan/figly


dev/fastly.ts
```diff
@@ -0,0 +1,9600 @@
...
+                {
+                  name: "type",
+                  isOptional: true,
+                },
+              ],
+            },
+            {
+              name: "--comment",
+              description: "Human-readable comment",
+              args: [],
```
mattschrage, line 73: Does `--comment` take an argument? If it does, can you have the `figly` script add an ArgumentObject to the `args` array?

mattschrage (changes requested):
Mostly looks great! There is one small tweak you can make to the the [figly](https://github.com/kailan/figly) script.

It seems like right now only optional arguments get added to the `args` array? If you can add non-optional arguments as well then the completion spec will be a perfect match!

dev/fastly.ts
kailan: Unfortunately, it seems the CLI's JSON output doesn't differentiate between flags with and without arguments. The best I can do is add the args to flags that are required or have defaults set, as they would always have values in those cases.

Manual overrides could be added for these flags, but there are a lot of commands and options and this would take some time.

cstrnt:
@kailan sorry for fiddling around in your PR, our commenting system wasn't working properly :(

kailan:
No worries! I hope the missing args is not a blocker for merging. It's something I can look at resolving in the future but it would require more extensive work.

dev/fastly.ts
mattschrage: Makes sense! Let's get this merged in 👍

mattschrage (approved):

PR #142, opened 2021-04-25
[YARN] add search and scripting

cstrnt:
Feature

https://withfig.com/docs/autocomplete/scripting and the search for NPM packages doesn't work

They both work :)

**Additional info:**

### DIFF
--- a/dev/yarn.ts
+++ b/dev/yarn.ts
@@ -1,32 +1,67 @@
-const yarnGenerators = {
-  getScripts: {
-    script: "cat package.json",
-    postProcess: function (output) {
-      if (output.trim() == "") {
-        return [];
-      }
-      try {
-        const packageContent = JSON.parse(output);
-        const scripts = packageContent["scripts"];
-        if (scripts) {
-          return Object.keys(scripts).map((scriptName) => ({
-            name: scriptName,
-            icon: "https://yarnpkg.com/favicon-32x32.png",
-          }));
-        }
-      } catch (e) {}
+type SearchResult = {
+  package: {
+    name: string;
+    description: string;
+  };
+  searchScore: number;
+};
 
+const searchGenerator: Fig.Generator = {
+  script: function (context) {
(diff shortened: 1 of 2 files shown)


dev/yarn.ts
```diff
@@ -1,32 +1,67 @@
...
+  // splitOn: "\n",
+  postProcess: function (out) {
+    if (out.trim() == "") {
+      return [];
+    }
+
+    try {
+      var packageContent = JSON.parse(out);
+      var scripts = packageContent["scripts"];
+      var figCompletions = packageContent["fig"];
```
QuiiBz, line 40: Should be `const` instead of `var`

dev/yarn.ts
```diff
@@ -1,32 +1,67 @@
...
+      if (scripts) {
+        const keys = Object.keys(scripts).map((key) => {
+          return Object.assign(
+            {},
+            { icon: "fig://icon?type=npm" },
+            (figCompletions || {})[key], // need the || {} otherwise it errors
+            { name: key, insertValue: key }
+          ); // ensure that name and insertValue are defined by "scripts" dict
+        });
+        return keys;
```
QuiiBz, line 51: Couldn't you add the `Fig.Suggestion` to add autocompletion?

QuiiBz (changes requested):

dev/yarn.ts
```diff
@@ -1,32 +1,71 @@
...
+    }
+
+    try {
+      const packageContent = JSON.parse(out);
+      const scripts = packageContent["scripts"];
+      const figCompletions = packageContent["fig"] || {};
+
+      if (scripts) {
+        return Object.keys(scripts).map((key) => {
+          const icon = "fig://icon?type=npm";
```
mattschrage, line 44: Should be `fig://icon?type=yarn`

mattschrage (approved):
Small tweak with the icon, otherwise looks good to merge!

PR #147, opened 2021-04-26
[GO] add completion spec

cstrnt:
Feature

There is no completion spec for `go`

There will be one 🥳 

**Additional info:**

### DIFF
--- a/dev/go.ts
+++ b/dev/go.ts
@@ -174,6 +174,17 @@ const globalOptions: Fig.Option[] = [
     description: "module download mode to use: readonly, vendor, or mod.",
     args: {
       name: "mode",
+      suggestions: [
+        {
+          name: "readonly",
+        },
+        {
+          name: "vendor",
+        },
+        {
+          name: "mod",
+        },
+      ],
     },
   },
   {
@@ -214,6 +225,7 @@ const globalOptions: Fig.Option[] = [
 const packagesArg: Fig.Arg = {
   name: "packages",
   variadic: true,
+  isOptional: true,
 };
 
 export const completionSpec: Fig.Spec = {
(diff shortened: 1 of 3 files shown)


dev/go.ts
```diff
@@ -0,0 +1,783 @@
...
+          name: "-t",
+          description:
+            "modules needed to build tests of packages specified on the command line.",
+        },
+        {
+          name: "-u",
+          description: "update to newer minor or patch releases when available",
+        },
+        {
+          name: "-u=patch",
```
cstrnt, line 387: FIX ME

specs/go.js
```diff
@@ -0,0 +1,720 @@
...
+                },
+                {
+                    name: "-deps",
+                    description: "terate over not just the named packages but also all their dependencies",
+                },
+                {
+                    name: "-f",
+                    insertValue: "-f '{cursor}'",
+                    description: "specify an alternate format for te list",
+                    args: {
```
mattschrage, line 393: Typo.

mattschrage (commented):
Everything seems fine to me. Once @falky97 takes a look, I think it's ready to merge in.

dev/go.ts
cstrnt: Actually not, it's a feature!

dev/go.ts
```diff
@@ -0,0 +1,783 @@
...
+          name: "-n",
+          description: "print the commands that would be executed",
+        },
+        {
+          name: "-x",
+          description: "print the commands as they are exectued",
+        },
+        {
+          name: "-mod",
+          description: "which module download modeto use",
```
cstrnt, line 343: typo

dev/go.ts
falky97: Fig's tokenizer will treat this as [-u, patch]. So we should add an optional argument to `-u`

specs/go.js
```diff
@@ -0,0 +1,720 @@
...
+        },
+    },
+    {
+        name: "-trimpath",
+        description: "remove all file system paths from the resulting executable.",
+    },
+]);
+var packagesArg = {
+    name: "packages",
+    variadic: true,
```
falky97, line 194: `isOptional: true`

dev/go.ts
```diff
@@ -0,0 +1,783 @@
...
+      options: [
+        ...globalOptions,
+        {
+          name: "-exec",
+          description: "invoke the binary using xprog",
+          args: {},
+        },
+      ],
+      args: {
+        name: "package",
```
falky97, line 683: Look into this more, but likely would add the prop: `isScript: true`

QuiiBz (approved):

falky97:
Going to merge this in. I know we went over this together. Given how poor go's documentation is I anticipate there are going to be some problems here. Here's hoping we have some go aficionados come in and give us a helping hand :)

PR #148, opened 2021-04-27
Add subcommands, arguments, and an environment generator to Conda

Alexanderchen929:
Conda update

Only missing get_all_packages functionality

### DIFF
--- a/dev/conda.ts
+++ b/dev/conda.ts
@@ -3,7 +3,7 @@ const getInstalledPackages: Fig.Generator = {
   postProcess: function (out) {
     const lines = out.split("\n");
     const installedPackages = [];
-    for (let i = 3; i < lines.length; i++) {
+    for (let i = 2; i < lines.length; i++) {
       installedPackages.push({
         name: lines[i],
         icon: "🐍",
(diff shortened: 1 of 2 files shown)


dev/conda.ts
```diff
@@ -13,25 +13,58 @@ const getInstalledPackages: Fig.Generator = {
...
+//     // for (let i = 1; i < lines.length; i++) {
+//     //   allPackages.push({
+//     //     name: lines[i],
+//     //     icon: "🐍",
+//     //   });
+//     // }
+//     return allPackages;
+//   },
+// };
+
```
cstrnt, line 36: Is there any reason why those are commented out and not deleted?

dev/conda.ts
```diff
@@ -13,25 +13,58 @@ const getInstalledPackages: Fig.Generator = {
...
+const getCondaEnvironments: Fig.Generator = {
+  script: "conda env list",
   postProcess: function (out) {
     const lines = out.split("\n");
-    const allPackages = [];
-    for (let i = 1; i < lines.length; i++) {
-      allPackages.push({
-        name: lines[i],
+    const installedPackages = [];
```
cstrnt, line 41: You could type this array as `Fig.Suggestion[]` to help with future development

dev/conda.ts
```diff
@@ -652,19 +712,27 @@ export const completionSpec: Fig.Spec = {
...
         },
         {
           name: ["--pkg-build"],
           description: "Package build number of the created package.",
+          args: {},
+        },
+        {
+          name: ["-n", "--name"],
+          description: "Name of environment.",
+          args: {},
```
cstrnt, line 730: This name flag (and others) are re-used all over this file. I think it would be awesome if we create a top-level const which has those shared options in order to reduce code duplication

cstrnt (changes requested):

dev/python.ts
```diff
@@ -20,11 +20,137 @@ export const completionSpec: Fig.Spec = {
...
+      description:
+        "Execute the Python code in command. command can be one or more statements separated by newlines, with significant leading whitespace as in normal module code.",
       args: {
         name: "command",
         isCommand: true,
       },
     },
+    {
+      name: ["-m"],
+      insertValue: "-m ",
```
cstrnt, line 32: This is not necessary  imo

cstrnt: Or do you want use something like this `-m '{cursor}'` ?

cstrnt (dismissed):

dev/conda.ts
Alexanderchen929: Hi! Brendan mentioned that we might use a fig server as an endpoint for search for all package managers. Since this issue is currently unresolved, I've just left the code as commented until we decide on a solution.

falky97 (approved):
looks good!

PR #150, opened 2021-04-27
[FASTLY] Add fastly completions with args

kailan:
generated with updates from https://github.com/kailan/figly/commit/0e9c2f2a8513e7f1e909fea6f6bc739200be071d


mattschrage (changes requested):
One final tweak!

mattschrage (approved):
This looks great! 🎉

dev/fastly.ts
```diff
@@ -0,0 +1,12449 @@
...
+        },
+      ],
+      options: [],
+    },
+  ],
+  options: [
+    {
+      name: "--help",
+      description: "Show context-sensitive help.",
+      args: [
```
falky97, line 12424: Hey - the spec looks good. I think the only thing that strikes me a little strange is why do--verbose and --help  have arguments? I am guessing they should not. Is this potentially a problem with the underlying parser? Would you be able to check this for us?

dev/fastly.ts
```diff
@@ -0,0 +1,12449 @@
...
+                },
+              ],
+            },
+          ],
+        },
+      ],
+      options: [],
+    },
+  ],
+  options: [
```
falky97, line 12420: Also, reading the docs, these seem like global flags. Adding them at the top level doesn't automatically add them to all the relevant subcommands unfortunately. One option is your parser adds these options to every single command. Another option is we set a setting in the completino spec called something like "globalFlags" and we add them in from Fig's end for you. I presume --token is pretty important though so maybe want to add it until we can do the latter?

PR #153, opened 2021-04-28
[NG] Initial Angular spec

NRaf:

Initial Angular spec with a few commands partially implemented.

Currently, there is no Angular spec.

Definitions for creating a Angular project as well as generating libraries, applications, components, and classes.

**Additional info:**

### DIFF
--- a/dev/ng.ts
+++ b/dev/ng.ts
@@ -1,0 +1,135 @@
+interface ProjectDetails {
+  projectType: string;
+}
+
+const projectsGenerator = {
+  script: "ng config projects",
+  postProcess: function (out) {
+    try {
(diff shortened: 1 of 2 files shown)


cstrnt (approved):
Looks good to me

dev/ng.ts
```diff
@@ -0,0 +1,135 @@
...
+        ([projectName, details]: [string, ProjectDetails]) => ({
+          name: projectName,
+          description: details.projectType,
+        })
+      );
+    } catch (e) {}
+    return [];
+  },
+};
+const projectsOption = {
```
cstrnt, line 21:
```suggestion
const projectsOption: Fig.Option = {
```

PR #156, opened 2021-04-29
Run prettier directly, rather than using ESLint

mattschrage:
Pulls prettier out of ESLint and runs the command separately.
prettier is included as a plugin of ESLint

**Additional info:**
ESLint with the prettier plugin was really slow on large files (eg. it took 2h to lint all the `aws` completion specs). Running prettier as a standalone command dramatically [increases performance](https://github.com/prettier/prettier-eslint-cli/issues/89#issuecomment-703478321).

There may be things we can do to improve compatibility: for instance, installing `eslint-config-prettier`. See this [document](https://prettier.io/docs/en/install.html) for more details.


mattschrage:
Also: the `aws` branch is based of of this because otherwise the pre-commit hooks would time out. (I pushed gcloud using `--no-verify`, so once this is merged in I'll redo that PR as well)

cstrnt:
Any particular reason why the scripts folder and the .github folder are ignored? I would love them to be pretty as well ;)

QuiiBz:
Do you plan to remove the comments in prettier's config?

cstrnt:
While I think the PR is great, I honestly think that we don't need to `.prettierignore` file. I think all files should be formatted in a nice way. Don't you think so?

cstrnt:
Works like a charm with your AWS Branch 😎

mattschrage:
I deleted the `.prettierignore` file. Everything can be pretty @cstrnt.

PR #160, opened 2021-04-29
[NEXT] Add spec for Next.js CLI

QuiiBz:
Adds the spec for Next.js's CLI

There isn't any spec for Next.js's CLI

This PR adds the Next.js CLI

**Additional info:**

### DIFF
--- a/dev/next.ts
+++ b/dev/next.ts
@@ -43,13 +43,17 @@ export const completionSpec: Fig.Spec = {
           description: "Represent the directory of the Next.js application",
           template: "folders",
         },
+      ],
+      options: [
         {
           name: "-p",
           description: "A port number on which to start the application",
+          args: {},
         },
         {
           name: "-H",
           description: "Hostname on which to start the application",
+          args: {},
         },
       ],
     },
@@ -62,6 +66,8 @@ export const completionSpec: Fig.Spec = {
           description: "Represent the directory of the Next.js application",
           template: "folders",
         },
+      ],
+      options: [
         {
           name: "-p",
           description: "A port number on which to start the application",
--- a/dev/next.ts
+++ b/dev/next.ts
@@ -46,12 +46,12 @@ export const completionSpec: Fig.Spec = {
       ],
       options: [
         {
-          name: "-p",
+          name: ["-p", "--port"],
           description: "A port number on which to start the application",
           args: {},
         },
         {
-          name: "-H",
+          name: ["-H", , "--hostname"],
           description: "Hostname on which to start the application",
           args: {},
         },
@@ -69,11 +69,11 @@ export const completionSpec: Fig.Spec = {
       ],
       options: [
         {
-          name: "-p",
+          name: ["-p", "--port"],
           description: "A port number on which to start the application",
         },
         {
-          name: "-H",
+          name: ["-H", , "--hostname"],
           description: "Hostname on which to start the application",
         },
       ],
--- a/dev/next.ts
+++ b/dev/next.ts
@@ -1,5 +1,18 @@
 const icon = "https://nextjs.org/static/favicon/favicon-16x16.png";
 
+const hostPortOptions: Fig.Option[] = [
+  {
+    name: ["-p", "--port"],
+    description: "A port number on which to start the application",
+    args: {},
+  },
+  {
+    name: ["-H", , "--hostname"],
+    description: "Hostname on which to start the application",
+    args: {},
+  },
+];
+
 export const completionSpec: Fig.Spec = {
   name: "next",
   description: "Next.js CLI to start, build and export your application",
@@ -49,18 +62,7 @@ export const completionSpec: Fig.Spec = {
           isOptional: true,
         },
       ],
-      options: [
-        {
-          name: ["-p", "--port"],
-          description: "A port number on which to start the application",
-          args: {},
-        },
-        {
-          name: ["-H", , "--hostname"],
-          description: "Hostname on which to start the application",
-          args: {},
-        },
-      ],
+      options: hostPortOptions,
     },
     {
       name: "start",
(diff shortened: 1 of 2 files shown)


specs/next.js
```diff
@@ -0,0 +1,114 @@
...
+                    template: "folders",
+                },
+                {
+                    name: "-p",
+                    description: "A port number on which to start the application",
+                },
+                {
+                    name: "-H",
+                    description: "Hostname on which to start the application",
+                },
```
cstrnt, line 53: hose are not args, those are options :)

specs/next.js
```diff
@@ -0,0 +1,114 @@
...
```
cstrnt, line 72: see above

cstrnt (changes requested):

dev/next.ts
```diff
@@ -0,0 +1,113 @@
...
+      name: "start",
+      description: "Start the application in production mode",
+      args: [
+        {
+          name: "dir",
+          description: "Represent the directory of the Next.js application",
+          template: "folders",
+        },
+        {
+          name: "-p",
```
mattschrage, line 66: `-H` and `-p` should take arguments.

specs/next.js
QuiiBz: Sorry for the mistake, my brain was struggling a bit 😅

QuiiBz:
Is this Danger error caused by my changes?

cstrnt:
> Is this Danger error caused by my changes?

I don't think there are any more errors, right? But either way, it's not your fault 😉

QuiiBz:
The error disappeared, so that’s fine!

dev/next.ts
```diff
@@ -0,0 +1,128 @@
...
+    {
+      name: ["-v", "--version"],
+      description: "Output the version number",
+    },
+  ],
+  subcommands: [
+    {
+      name: "build",
+      description: "Create an optimized production build of your application",
+      icon: "https://nextjs.org/static/favicon/favicon-16x16.png",
```
cstrnt, line 19: I think it would be great if you define this icon property as a top level const so we would only need to update it once when we want to change it. I also opened #174 because I think it would be great if the icon is inherited

dev/next.ts
```diff
@@ -0,0 +1,128 @@
...
+        {
+          name: "--debug",
+          description: "Enable more verbose build output",
+        },
+      ],
+    },
+    {
+      name: "dev",
+      description: "Start the application in development mode",
+      icon: "https://nextjs.org/static/favicon/favicon-16x16.png",
```
cstrnt, line 42: see above

dev/next.ts
```diff
@@ -0,0 +1,128 @@
...
+          name: ["-H", , "--hostname"],
+          description: "Hostname on which to start the application",
+          args: {},
+        },
+      ],
+    },
+    {
+      name: "start",
+      description: "Start the application in production mode",
+      icon: "https://nextjs.org/static/favicon/favicon-16x16.png",
```
cstrnt, line 67: see above

dev/next.ts
```diff
@@ -0,0 +1,128 @@
...
+        {
+          name: ["-H", , "--hostname"],
+          description: "Hostname on which to start the application",
+        },
+      ],
+    },
+    {
+      name: "export",
+      description: "Exports the application for production deployment",
+      icon: "https://nextjs.org/static/favicon/favicon-16x16.png",
```
cstrnt, line 90: see above

dev/next.ts
```diff
@@ -0,0 +1,128 @@
...
+        {
+          name: "-s",
+          description: "Do not print any messages to console",
+        },
+      ],
+    },
+    {
+      name: "telemetry",
+      description: "Allows you to control Next.js' telemetry collection",
+      icon: "https://nextjs.org/static/favicon/favicon-16x16.png",
```
cstrnt, line 109: see above

cstrnt (changes requested):
Just a tiny little wish from my side :wink

dev/next.ts
QuiiBz: Originally (see https://github.com/withfig/autocomplete/pull/160/commits/1a09891e0ed2f1b634b3d32b8300c169ff2fb698#diff-16de0e0b934e0e5053f31155e3160ab60c2565f6bf37d867ea1ab708a85de232R2-R5), the `icon` was set to the top-level, but as you said the sub-commands don't inherit it, so I had to put them in each.

Let's see if this can be improved in #174

cstrnt: I mean I would just prefer having a const called `nextIcon` at the top of the file and then referencing that const over and over

QuiiBz: Never mind, I didn't understand what you said. Let me fix this!

QuiiBz:
Can we merge this @cstrnt?

cstrnt:
@QuiiBz I think it's good to merge, but you could for example extract the shared options like `-p` and `--host` into a top-level const in order to reduce duplication, but that's just my personal preference

QuiiBz:
Good advice! Just updated it.

dev/next.ts
```diff
@@ -0,0 +1,108 @@
...
+      name: "build",
+      description: "Create an optimized production build of your application",
+      icon,
+      args: [
+        {
+          name: "dir",
+          description: "Represent the directory of the Next.js application",
+          template: "folders",
+          isOptional: true,
+        },
```
cstrnt, line 47: This should be the `dirArgument` as well

cstrnt (changes requested):

dev/next.ts
QuiiBz: Updated

cstrnt (approved):
Awesome 😄 Thank you very much, Tom :)

PR #161, opened 2021-04-29
[NX] add Spec for Nx CLI

ronderksen:
Feature

Fig currently has no spec for use with the Nx CLI.

Introduces autocomplete support for the Nx CLI. The CLI has quite a few commands, which can be hard to remember. 

**Additional info:**
There's still a few commands missing. I will add them in the next few days.

### DIFF
--- a/dev/nx.ts
+++ b/dev/nx.ts
@@ -1,95 +1,81 @@
-const nxGenerators: Record<string, Fig.Generator> = {
+type PostProcessFn = (out: string, context: string[]) => Fig.Suggestion[];
+
+type PostProcessWorkspaceFn = (
+  filterFn: (projectName: string, index: number, array: string[]) => boolean
+) => PostProcessFn;
+
+interface NxGenerators {
+  apps: Fig.Generator;
+  e2eApps: Fig.Generator;
+  appsAndLibs: Fig.Generator;
+  localSchematics: Fig.Generator;
+  localGenerators: Fig.Generator;
+  installedPlugins: Fig.Generator;
+  pluginsSchematics: Fig.Generator;
+}
+
+const processWorkspaceJson: PostProcessWorkspaceFn = (filterFn) => (out) => {
+  try {
+    const workspace = JSON.parse(out);
+    return Object.keys(workspace.projects)
+      .filter(filterFn)
+      .map((suggestion) => ({
+        name: suggestion,
+        type: "option",
+      }));
+  } catch (err) {
+    return [];
+  }
+};
+
+const processGenerators: PostProcessFn = function (out) {
+  return out
+    .split("\n")
+    .map((line) => line.split(" ").pop())
+    .map((suggestion) => ({
+      name: suggestion,
+      type: "option",
+    }));
+};
+
+const oneDayCache: Fig.Cache = {
+  ttl: 60 * 60 * 24,
+};
+
+const nxGenerators: NxGenerators = {
   apps: {
     script: "cat workspace.json",
-    postProcess: function (out) {
-      try {
-        const workspace = JSON.parse(out);
-        return Object.keys(workspace.projects)
-          .filter(
-            (projectName) =>
-              workspace.projects[projectName].projectType === "application"
-          )
-          .filter((projectName) => !projectName.endsWith("-e2e"))
-          .map((suggestion) => ({
-            name: suggestion,
-            type: "option",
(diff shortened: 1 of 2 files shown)


dev/nx.ts
```diff
@@ -0,0 +1,492 @@
+const nxGenerators: Record<string, Fig.Generator> = {
```
cstrnt, line 1: using Record<string, Fig.Generator> causes TypeScript to also allow nxGenerators.nonExistingProperty. Even though it looks hacky, I would prefer not typing the object itself but rather each value as` Fig.Generator`

dev/nx.ts
```diff
@@ -0,0 +1,492 @@
...
+    },
+    postProcess: function (out) {
+      return out
+        .split("\n")
+        .map((line) => line.split(" ").pop())
+        .map((suggestion) => ({
+          name: suggestion,
+          type: "option",
+        }));
+    },
```
cstrnt, line 84: This is basically the same as the generator for `localSchematics`.  I would prefer referencing the cache and postProcess stuff from the `localSchematics` generator to reduce code doublication

dev/nx.ts
```diff
@@ -0,0 +1,492 @@
...
+  },
+  installedPlugins: {
+    script: () => {
+      return "nx list";
+    },
```
cstrnt, line 89:
```suggestion
    script: "nx list" ,
```

dev/nx.ts
```diff
@@ -0,0 +1,492 @@
...
+    cache: {
+      ttl: 60 * 60 * 24,
+    },
```
cstrnt, line 92: The cache property is re-used. I would prefer to create a top-level const with the type `Fig.Cache`

dev/nx.ts
```diff
@@ -0,0 +1,492 @@
...
+      if (argument.indexOf(":") > -1) {
+        return `nx list ${argument.split(":")[0]}`;
+      } else {
+        return "nx list";
+      }
+    },
+    trigger: ":",
+    cache: {
+      ttl: 60 * 60 * 24,
+    },
```
cstrnt, line 119: see above

cstrnt (changes requested):
What a huge PR! Thank you very much 😊

dev/nx.ts
```diff
@@ -0,0 +1,476 @@
...
+      .map((suggestion) => ({
+        name: suggestion,
+        type: "option",
+      }));
+  } catch (err) {
+    return [];
+  }
+};
+
+const processGenerators: PostProcessFn = function (out) {
```
QuiiBz, line 31: Shouldn't this be a `() => {...}` instead of a `function`, since you mostly used them?

dev/nx.ts
```diff
@@ -0,0 +1,476 @@
...
+  },
+  localGenerators: {
+    script: "ls tools/generators",
+    cache: oneDayCache,
+    postProcess: processGenerators,
+  },
+  installedPlugins: {
+    script: "nx list",
+    cache: oneDayCache,
+    postProcess: function (out) {
```
QuiiBz, line 79: Same here, I would prefer not to mix the two and instead choose one.

QuiiBz (approved):

dev/nx.ts
```diff
@@ -0,0 +1,476 @@
...
+    }));
+};
+
+const oneDayCache: Fig.Cache = {
+  ttl: 60 * 60 * 24,
+};
+
+const nxGenerators: NxGenerators = {
+  apps: {
+    script: "cat workspace.json",
```
mattschrage, line 47: I am not super familiar with the `nx` tool - is it possible to run `nx` commands from child directories of the root folder that contains the `workspace.json` file?

If this is the case, you might be able to adapt the same script that the `npm` CLI uses to find the `package.json` file:

```
until [[ -f package.json ]] || [[ $PWD = '/' ]]; do cd ..; done; cat package.json
```

mattschrage (commented):
This looks great! Thanks for contributing @ronderksen :)

I had one small question, but I think this is about ready to merge in.

dev/nx.ts
ronderksen: No, it only works from the root folder.

mattschrage:
@ronderksen sorry for the delay on this! Everything looks good to me. Going to merge now!

PR #162, opened 2021-04-29
[Firebase]: add project alias generator

dereklowlind:
Added a generator to the firebase.js spec. Not fully working. The goal was to add a generator like for git getting the local branches but getting the user's firebase projects. The command "firebase use <project alias>" is the target for the generator. Getting the user's projects is done using the "firebase projects:list" command which calls the firebase servers and takes almost 10 seconds to complete on my internet. The generator is currently able to grab only the first project in the list and display it as an option. I am not sure if it is because the "firebase projects:list" command does not return the complete list or because my code is having an issue (can I console log in the generator function?). 

Due to the server lag this may or may not be useful addition. It is slow but from a workflow standpoint but it is faster than entering the "firebase projects:list" command then copying the project's alias (which is my current workflow).


falky97 (changes requested):
Hey Derek - this generator looks fine! Do you think it's a your internet problem or is it always this slow? You could add the option to cache the spec?

We will honestly probably approve this, however, all Fig specs must be done by editing the typescript file (in this case `dev/firebase.ts`). Want to make literally the exact same changes you already did, then run `npm run build` to compile, and then push?

dereklowlind:
Can you try running the "firebase projects:list" command. My internet is pretty good ~300mbps, it might also be something backend at firebase with auth and stuff.

Any ideas on why it only picks up only the first of the ten firebase projects I have?

Yep I can do the changes to the ts file instead.

dev/firebase.ts
```diff
@@ -1,3 +1,17 @@
+const firestoreGenerators: Record<string, Fig.Generator> = {
+  projectAliases: {
+    script: "firebase projects:list", // this calls to a firebase server and is therefore slow
+    postProcess: (out) => {
+      var getAliasRegex = /^│ (\w.*?)│/gm;
+      var aliasesRaw = Array.from(out.matchAll(getAliasRegex));
```
cstrnt, line 6: I would prefer using `const` here :)

dev/firebase.ts
```diff
@@ -1,3 +1,17 @@
+const firestoreGenerators: Record<string, Fig.Generator> = {
```
cstrnt, line 1: using `Record<string, Fig.Generator>` causes TypeScript to also allow ` firestoreGenerators.nonExistingProperty`. Even though it looks hacky, I would prefer not typing the object itself but rather each value `as Fig.Generator` cc @QuiiBz

cstrnt (changes requested):

dev/firebase.ts
QuiiBz: Agree, but in this case since there is only one generator, you can simply do this:
```typescript
const projectAliasesGenerator: Fig.Generator = {
  script: "firebase projects:list", // this calls to a firebase server and is therefore slow
  postProcess: (out) => {
      ...
  },
};
```

PR #163, opened 2021-04-30
[LN]: add command

dereklowlind:
Created a template for the ln command as I noticed there was not one yet

### DIFF
--- a/dev/ln.ts
+++ b/dev/ln.ts
@@ -1,0 +1,36 @@
+export const completionSpec: Fig.Spec = {
+  name: "ln",
+  description: "Create (hard) symbolic links to files",
+  args: {},
+  options: [
+    {
+      name: "-s",
+      description: "Create a symbolic link",
+    },
+    {
+      name: "-v",
+      description: "verbose",
+    },
+    {
+      name: ["-F"],
+      description: "If link name already exists replace it",
+    },
+    {
+      name: "-h",
+      description: "Don't follow symbolic links",
+    },
+    {
+      name: "-f",
+      description:
+        "If link name already exists unlink the old one before creating the new one",
+    },
+    {
+      name: "-i",
+      description: "Prompt if proposed link already exists",
+    },
+    {
+      name: "-n",
+      description: "Same as -h don't follow symbolic links",
+    },
+  ],
+};
--- a/dev/ln.ts
+++ b/dev/ln.ts
@@ -1,27 +1,22 @@
+const source_dest_args: Fig.Arg[] = [
+  {
+    name: "source_file",
+    // template: ["filepaths", "folders"], // is there an absolute path version of this? ln requires absolute path
+  },
+  {
+    name: "link_name or link_dirname",
+  },
+];
+
 export const completionSpec: Fig.Spec = {
   name: "ln",
   description: "Create (default hard) symbolic links to files",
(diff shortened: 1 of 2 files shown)


QuiiBz:
Hi, thanks for contributing for the first time to Fig! Did you planned to add the `args` for ln? It takes two args: `source_file` and `link_name`, or a variadic option `linkname_dir`.

dev/ln.ts
```diff
@@ -0,0 +1,36 @@
...
+    {
+      name: "-s",
+      description: "Create a symbolic link",
+    },
+    {
+      name: "-v",
+      description: "verbose",
+    },
+    {
+      name: ["-F"],
```
QuiiBz, line 15: You can remove the `[]` since there is only one name.

QuiiBz (changes requested):

dev/ln.ts
dereklowlind: Hi @QuiiBz I removed those brackets and added the args.

dev/ln.ts
```diff
@@ -0,0 +1,76 @@
+export const completionSpec: Fig.Spec = {
+  name: "ln",
+  description: "Create (default hard) symbolic links to files",
+  args: [
+    {
+      name: "source_file",
+      // template: ["filepaths", "folders"], // is there an absolute path version of this? ln requires absolute path
```
QuiiBz, line 7: @mattschrage Do you have any plan on adding this template? It should combine `filepaths` and `folders` templates.

dev/ln.ts
```diff
@@ -0,0 +1,76 @@
...
+      name: "-s",
+      description: "Create a symbolic link",
+      args: [
+        {
+          name: "source_file",
+        },
+        {
+          name: "link_name or link_dirname",
+        },
+      ],
```
cstrnt, line 24: Those args are reused over and over. I would prefer having a top-level const of the type `Fig.Args[]` and you can just re-use it

cstrnt (changes requested):

dev/ln.ts
dereklowlind: @cstrnt  thank you for the feedback, changes are implemented and pushed

dev/ln.ts
```diff
@@ -0,0 +1,50 @@
+const source_dest_args: Fig.Arg[] = [
+  {
+    name: "source_file",
+    // template: ["filepaths", "folders"], // is there an absolute path version of this? ln requires absolute path
```
QuiiBz, line 4: We can stay with `["filepaths", "folders"],` since `ln` [does not](https://linuxize.com/post/how-to-create-symbolic-links-in-linux-using-the-ln-command/#creating-symlink-to-a-file) require absolute paths

QuiiBz (changes requested):

dev/ln.ts
```diff
@@ -0,0 +1,50 @@
+const source_dest_args: Fig.Arg[] = [
```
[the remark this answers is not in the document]
dereklowlind: Added the change and pushed

dev/ln.ts
dereklowlind: Added the change and pushed

QuiiBz (approved):

PR #167, opened 2021-05-01
[YARN]: Load specs of cli that exists in current package

QuiiBz:
This PR adds a new feature to the `yarn` spec

Currently, if you want to run a cli from yarn using `yarn <cli> ...`, you won't get any completion

Fig loads the specs of the cli that are in the current `package.json` and provide completion for them

**Additional info:**
Example:
The `next` package exists in the current `package.json`, so the spec for it will be loaded.
Screenshot 2021-05-01 at 11 31 10
Screenshot 2021-05-01 at 11 32 03

### DIFF
--- a/dev/yarn.ts
+++ b/dev/yarn.ts
@@ -121,6 +121,30 @@ const configList: Fig.Generator = {
 export const completionSpec: Fig.Spec = {
   name: "yarn",
   description: "Manage packages and run scripts",
+  generateSpec: async (context, executeShellCommand) => {
+    const { script, postProcess } = packageList;
+    const packages = postProcess(
+      await executeShellCommand(script as string)
+    ).map(({ name }) => name as string);
+
+    const cli = ["vue", "nuxt", "expo", "jest", "next"];
+    const subcommands: Fig.Subcommand[] = [];
+
+    packages
+      .filter((name) => cli.includes(name))
+      .forEach((name) => {
+        subcommands.push({
+          name,
+          loadSpec: name,
+          icon: "fig://icon?type=package",
+        });
+      });
+
+    return {
+      name: "yarn",
+      subcommands,
+    } as Fig.Spec;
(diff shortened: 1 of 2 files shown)


dev/yarn.ts
```diff
@@ -121,6 +121,30 @@ const configList: Fig.Generator = {
 export const completionSpec: Fig.Spec = {
   name: "yarn",
   description: "Manage packages and run scripts",
+  generateSpec: async (context, executeShellCommand) => {
```
cstrnt, line 124:
```suggestion
  generateSpec: async (_context, executeShellCommand) => {
```

TypeScript suggests prefixing unused params with `_`

dev/yarn.ts
```diff
@@ -121,6 +121,30 @@ const configList: Fig.Generator = {
...
+
+    packages
+      .filter((name) => cli.includes(name))
+      .forEach((name) => {
+        subcommands.push({
+          name,
+          loadSpec: name,
+          icon: "fig://icon?type=package",
+        });
+      });
```
cstrnt, line 141: Instead of creating a temp array for the subcommands you can simply return packages.filter.map

```ts
 const subcommands = packages
      .filter((name) => cli.includes(name))
      .map((name) => ({
          name,
          loadSpec: name,
          icon: "fig://icon?type=package",
      }));
```

cstrnt:
Apparently Danger doesn't like github code suggestions

dev/yarn.ts
QuiiBz: Good idea

QuiiBz:
> Apparently Danger doesn't like github code suggestions

Yup that's strange. I just pushed to fix https://github.com/withfig/autocomplete/pull/167#discussion_r624667638, is everything good?

cstrnt (approved):
Awesome!

PR #168, opened 2021-05-01
[PHPUNIT-WATCHER] Add watcher generator

binaryk:
Plugin for [*phpunit-watcher*](https://github.com/spatie/phpunit-watcher) autocompletion test name

GIF: https://cln.sh/YtMpEI

image


specs/phpunit-watch.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+    ],
+    options: [
+        {
+            name: ["--version"],
+            description: "The current Homebrew version",
+        },
+    ],
+};
+
```
QuiiBz, line 46: Please remove this generated spec, since the right one is https://github.com/withfig/autocomplete/blob/adb02aafe07fc383f50bbe5bf5e7b98f4e764795/specs/phpunit-watcher.js

QuiiBz (changes requested):

specs/phpunit-watch.js
binaryk: Addressed, thanks!

dev/phpunit-watcher.ts
```diff
@@ -0,0 +1,46 @@
+var tests = {
```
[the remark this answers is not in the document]
binaryk: Addressed, thanks!

dev/phpunit-watcher.ts
```diff
@@ -1,18 +1,20 @@
 const tests: Record<string, Fig.Generator> = {
```
QuiiBz, line 1: This should only be a `Fig.Generator`, not a `Record<string, Fig.Generator>`. `Fig.Generator` type contains both the `script` and `postProcess` fields.

QuiiBz (changes requested):

dev/phpunit-watcher.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    return out
+      .split("\n")
+      .filter((line) => line.indexOf("::") > 0)
+      .map((line) => {
+        return {
+          name: line.substring(line.indexOf("::") + 2, line.length),
+          icon: "fig://icon?type=php",
+          description: line.substring(line.indexOf("::") + 2, line.length),
```
cstrnt, line 15: Does it make sense that the name and the description are the same?

binaryk: Yes.

dev/phpunit-watcher.ts
binaryk: Addressed @QuiiBz . Thank you!

QuiiBz (approved):

QuiiBz:
Thanks for your first contribution!

PR #169, opened 2021-05-01
[EXA] Add spec for exa

alexnaiman:
Added new spec for exa 
There isn't any spec for exa
This PR adds the exa
**Additional info:**

### DIFF
--- a/dev/exa.ts
+++ b/dev/exa.ts
@@ -1,6 +1,3 @@
-// To learn more about FIg's autocomplete standard visit: https://withfig.com/docs/autocomplete/building-a-spec#building-your-first-autocomplete-spec
-
-// The below is a dummy example for git. Make sure to change the file name!
 export const completion: Fig.Spec = {
   name: "exa",
   description: "a modern replacement for ls",
--- a/dev/exa.ts
+++ b/dev/exa.ts
@@ -193,7 +193,7 @@ export const completion: Fig.Spec = {
     {
       description:
         "which timestamp field to list (modified, accessed, created)",
-      name: ["-t", "--time FIELD"],
+      name: ["-t", "--time"],
       args: {
         name: "FIELD",
         suggestions: [
--- a/dev/exa.ts
+++ b/dev/exa.ts
@@ -1,0 +1,217 @@
+// To learn more about FIg's autocomplete standard visit: https://withfig.com/docs/autocomplete/building-a-spec#building-your-first-autocomplete-spec
+
+// The below is a dummy example for git. Make sure to change the file name!
+export const completion: Fig.Spec = {
+  name: "exa",
+  description: "a modern replacement for ls",
+  args: {
+    variadic: true,
+    name: "files...",
+    template: "filepaths",
+  },
+  options: [
+    {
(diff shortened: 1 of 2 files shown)


dev/exa.ts
```diff
@@ -0,0 +1,217 @@
...
+      description: "show hidden and 'dot' files",
+      name: ["-a", "--all"],
+    },
+    {
+      description: "list directories like regular files",
+      name: ["-d", "--list-dirs"],
+    },
+    {
+      description: "limit the depth of recursion",
+      name: ["-L", "--level DEPTH"],
```
cstrnt, line 82: I think DEPTH should not be included in the name because it is in fact an argument, right?

dev/exa.ts
```diff
@@ -0,0 +1,217 @@
...
+      description: "which field to sort by",
+      name: ["-s", "--sort"],
+      args: {
+        name: "sort field",
+        suggestions: [
+          {
+            name: "name",
+          },
+          {
+            name: "Name",
```
cstrnt, line 98: Does is make a difference for exa?

dev/exa.ts
```diff
@@ -0,0 +1,217 @@
...
+      description: "list directories before other files",
+      name: ["--group-directories-first"],
+    },
+    {
+      description: "list only directories",
+      name: ["-D", "--only-dirs"],
+    },
+    {
+      description: "glob patterns (pipe-separated) of files to ignore",
+      name: ["-I", "--ignore-glob"],
```
cstrnt, line 152: I think here is an `arg` required

dev/exa.ts
```diff
@@ -0,0 +1,217 @@
...
+      name: ["-m", "--modified"],
+    },
+    {
+      description: "show number of file system blocks",
+      name: ["-S", "--blocks"],
+    },
+    {
+      description:
+        "which timestamp field to list (modified, accessed, created)",
+      name: ["-t", "--time FIELD"],
```
cstrnt, line 193: Missing arg

cstrnt (changes requested):

dev/exa.ts
alexnaiman: My bad, I generated the spec using a script and some things I had to manually correct

dev/exa.ts
alexnaiman: Yes, it actually does 😄 

>  Fields starting with a capital letter will sort uppercase before  lowercase:  'A'  then 'B' then 'a' then 'b'.  Fields starting with a lowercase letter will mix them: 'A' then 'a' then 'B' then 'b'.

dev/exa.ts
alexnaiman: Added right below

dev/exa.ts
cstrnt: No worries 😄

cstrnt (approved):
Awesome!

QuiiBz (approved):

mattschrage (approved):
This is awesome! Thanks for contributing @alexnaiman 🎉

PR #170, opened 2021-05-01
[ADB] Added adb spec

alexnaiman:
Adds the spec for adb CLI
There isn't any spec for adb CLI
This PR adds the adb CLI
**Additional info:**

### DIFF
--- a/dev/adb.ts
+++ b/dev/adb.ts
@@ -1,4 +1,4 @@
-const installOptions = [
+const installOptions: Fig.Option[] = [
   {
     name: "-l",
     description: "forward-lock the app",
@@ -72,7 +72,7 @@ const installOptions = [
   },
 ];
 
-const compressionOptions = [
+const compressionOptions: Fig.Option[] = [
   {
     description:
       "enable compression with a specified algorithm (any, none, brotli)",
@@ -98,7 +98,7 @@ const compressionOptions = [
   },
 ];
 
-const forwardConnectionSuggestions = [
+const forwardConnectionSuggestions: Fig.Suggestion[] = [
   {
     name: "tcp",
     insertValue: "tcp:",
@@ -129,7 +129,7 @@ const forwardConnectionSuggestions = [
   },
 ];
 
-const reverseConnectionSuggestions = [
+const reverseConnectionSuggestions: Fig.Suggestion[] = [
   {
     name: "tcp",
     insertValue: "tcp:",
@@ -379,6 +379,7 @@ export const completion: Fig.Spec = {
         "generate adb public/private key; private key stored in FILE",
       args: {
         name: "FILE",
+        template: "filepaths",
       },
     },
     {
@@ -543,6 +544,7 @@ export const completion: Fig.Spec = {
       args: {
         isOptional: true,
         name: "COMMANDS ...",
+        variadic: true,
       },
     },
     {
--- a/dev/adb.ts
+++ b/dev/adb.ts
@@ -1,3 +1,153 @@
+const installOptions = [
+  {
+    name: "-l",
+    description: "forward-lock the app",
+  },
+  {
+    name: "-r",
+    description: "replace existing application",
+  },
+  {
+    name: "-t",
+    description: "allow test packages",
+  },
+  {
+    name: "-d",
+    description: "allow version code downgrade (debuggable packages only)",
+  },
+  {
+    name: "-s",
+    description: "install on SD card instead of internal storage",
+  },
+  {
+    name: "-g",
+    description: "grant all runtime permissions",
+  },
+  {
+    description: "override platform's default ABI",
+    name: "--abi",
+    args: {
+      name: "ABI",
+    },
+  },
+  {
+    description: "cause the app to be installed as an ephemeral install app",
+    name: "--instant",
+  },
+  {
+    description:
+      "always push APK to device and invoke Package Manager as separate steps",
+    name: "--no-streaming",
+  },
+  {
+    description: "force streaming APK directly into Package Manager",
+    name: "--streaming",
+  },
+  {
+    description: "use fast deploy",
+    name: "--fastdeploy",
+  },
+  {
+    description: "prevent use of fast deploy",
+    name: "--no-fastdeploy",
+  },
+  {
+    description: "force update of deployment agent when using fast deploy",
+    name: "--force-agent",
+  },
+  {
+    description:
+      "update deployment agent when local version is newer and using fast deploy",
+    name: "--date-check-agent",
+  },
+  {
+    description:
+      "update deployment agent when local version has different version code and using fast deploy",
+    name: "--version-check-agent",
+  },
+  {
+    description:
+      "locate agent files from local source build (instead of SDK location)",
+    name: "--local-agent",
+  },
+];
+
+const compressionOptions = [
+  {
+    description:
+      "enable compression with a specified algorithm (any, none, brotli)",
+    name: "-z",
+    args: {
+      name: "ALGORITHM",
+      suggestions: [
+        {
+          name: "any",
+        },
+        {
+          name: "none",
+        },
+        {
+          name: "brotli",
+        },
+      ],
+    },
+  },
+  {
+    description: "disable compression",
+    name: "-Z",
+  },
+];
+
+const forwardConnectionSuggestions = [
+  {
+    name: "tcp",
+    insertValue: "tcp:",
+  },
+  {
+    name: "localabstract",
+    insertValue: "localabstract:",
+  },
+  {
+    name: "localreserved",
+    insertValue: "localreserved:",
+  },
+  {
+    name: "localfilesystem",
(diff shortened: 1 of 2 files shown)


dev/adb.ts
```diff
@@ -0,0 +1,1011 @@
+export const completion: Fig.Spec = {
+  name: "adb",
+  description: "Android Debug Bridge",
+  subcommands: [
+    {
+      name: "devices",
+      description: " list connected devices ",
+      options: [
+        {
+          name: ["-l"],
```
QuiiBz, line 10: You can safely remove the array and simply write `"-l"` since it has no other names. Same for below.

dev/adb.ts
```diff
@@ -0,0 +1,1011 @@
...
+      description: "restart adbd listening on USB",
+    },
+    {
+      name: "sideload",
+      description: "sideload the given full OTA package",
+      args: {
+        name: "OTAPACKAGE",
+      },
+    },
+
```
QuiiBz, line 65: Please remove this empty line.

dev/adb.ts
```diff
@@ -0,0 +1,1011 @@
...
+          description: "kick connection from device side to force reconnect",
+          name: "device",
+        },
+        {
+          description: "reset offline/unauthorized devices to force reconnect`",
+          name: "offline",
+        },
+      ],
+    },
+
```
QuiiBz, line 88: Same, please remove the empty line.

alexnaiman:
@QuiiBz done with the changes

specs/adb.js
```diff
@@ -0,0 +1,968 @@
...
+                },
+            ],
+            args: [
+                {
+                    name: "LOCAL",
+                    variadic: true,
+                    template: "filepaths",
+                },
+                {
+                    name: "REMOTE",
```
cstrnt, line 644: I'm just wondering how this plays together with fig cc @mattschrage @falky97

cstrnt: Because there are two args but one of them is variadic

specs/adb.js
```diff
@@ -0,0 +1,968 @@
...
+                            name: "localfilesystem:",
+                            insertValue: "localfilesystem:",
+                        },
+                        {
+                            name: "dev:",
+                            insertValue: "dev:",
+                        },
+                        {
+                            name: "jdwp:",
+                            insertValue: "jdwp:",
```
cstrnt, line 838: I think all of those should have the colon (`:`) removed from the name because they are already in the insertvalue

specs/adb.js
```diff
@@ -0,0 +1,968 @@
...
+                    name: "--no-rebind",
+                    description: "reversal fails if the specified socket is already bound through a previous reverse command",
+                },
+            ],
+            args: [
+                {
+                    name: "REMOTE -> port|domain|device|pid",
+                    suggestions: [
+                        {
+                            name: "tcp:",
```
cstrnt, line 877: See above

cstrnt (changes requested):
Looks good to me! But there is quite some repetition in the code. Maybe you could create some shared options / args to it easier :)

specs/adb.js
alexnaiman: Yeah, I didn't see that. It just ignores the second param. It always suggest the "LOCAL" argument 

Screen Shot 2021-05-01 at 11 54 51 PM

alexnaiman: This is the last requested change that I don't know how to tackle and I also don't know if it can be tackled in the way the command works.

mattschrage: Sorry for the slow response here. I don’t think we have good answer for this right now.

The workaround would probably be to have the first argument be non-variadic and have the second suggest both - with something like this: `name: “file | remote”`

The second would be variadic and have the same  generator as the first.

mattschrage (approved):
This is massive! Nice work - android devs everywhere will be grateful :)

Two small things:
- As @cstrnt, we don't really have a great solution for a variadic argument followed by *other* argument. There are workarounds but really this should be addressed at the core autocomplete logic rather than in individual completion specs. I opened an issue for this #177.

- There are some places where the argument is a shell command. If you annotate these arguments with `isCommand: true` then Fig will continue to provide autocompletion!

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
+const installOptions = [
```
cstrnt, line 1: Hey alex, any chance you could type this so we can catch errors in the future? 
```suggestion
const installOptions:Fig.Option[] = [
```

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+    name: "--version-check-agent",
+  },
+  {
+    description:
+      "locate agent files from local source build (instead of SDK location)",
+    name: "--local-agent",
+  },
+];
+
+const compressionOptions = [
```
cstrnt, line 75: see above

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+      ],
+    },
+  },
+  {
+    description: "disable compression",
+    name: "-Z",
+  },
+];
+
+const forwardConnectionSuggestions = [
```
cstrnt, line 101: See above, but type this as `Fig.Suggestion[]`

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+    name: "jdwp",
+    insertValue: "jdwp:",
+  },
+  {
+    name: "acceptfd",
+    insertValue: "acceptfd:",
+  },
+];
+
+const reverseConnectionSuggestions = [
```
cstrnt, line 132: see above

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+      name: "wait-for-any-disconnect",
+      description: "wait for any in state=disconnect",
+    },
+
+    {
+      name: "keygen",
+      description:
+        "generate adb public/private key; private key stored in FILE",
+      args: {
+        name: "FILE",
```
cstrnt, line 381: I think you could add `template: filepaths` here to get nice autocompletion for the file here 😊

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+        },
+      ],
+    },
+    {
+      name: "bugreport",
+      description: "write bugreport to given PATH [default=bugreport.zip];",
+      args: [
+        {
+          name: "PATH",
+          isOptional: true,
```
cstrnt, line 424: see above

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+      ],
+    },
+    {
+      name: "shell",
+      description:
+        "run remote shell command (interactive shell if no command given)",
+      options: [
+        {
+          name: "-e",
+          description: "choose escape character, or `none` default '~'",
```
cstrnt, line 520: I think this takes an optional argument, right?

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+          description: "-tt: force pty allocation",
+        },
+        {
+          name: "-x",
+          description: "disable remote exit codes and stdout/stderr separation",
+        },
+      ],
+      args: {
+        isOptional: true,
+        name: "COMMANDS ...",
```
cstrnt, line 545: should this be variadic?

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+      name: "-t",
+      args: {
+        name: "ID",
+      },
+    },
+    {
+      description: "name of adb server host [default=localhost]",
+      name: "-H",
+      args: {
+        name: "host name",
```
cstrnt, line 761: should be optional, right?

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+      },
+    },
+    {
+      description: "port of adb server [default=5037]",
+      name: "-P",
+      args: {
+        name: "port",
```
cstrnt, line 768: see above

dev/adb.ts
```diff
@@ -0,0 +1,780 @@
...
+      },
+    },
+    {
+      description:
+        "listen on given socket for adb server [default=tcp:localhost:5037]",
```
cstrnt, line 773: see above

cstrnt (changes requested):
Hey Alex, thanks for this huge PR. I just left some comments which I think might improve this spec 😄

dev/adb.ts
mattschrage: This is definitely correct! However, not necessary to add before we merge. We are just trying to ensure the CLI skeleton is correct - we can always add/update generators later!

mattschrage:
I think we can merge this in @cstrnt. Seems like the skeleton is correct.

alexnaiman:
@mattschrage @cstrnt Hello! thanks for all the feedback. I was really busy these past weeks and I didn't get to do all the changes. I will try at the beginning of the next week to do all the changes you requested, and in case you merged this PR all the fixes will be in a new one.

PR #171, opened 2021-05-01
[AWS] [SECRETSMANAGER] add several generators

nagypeterjob:

* [x] cancel-rotate-secret
* [x] create-secret
* [x] delete-resource-policy
* [x] delete-secret
* [x] describe-secret
* [x] get-random-password
* [x] get-resource-policy
* [x] get-secret-value
* [x] list-secret-version-ids
* [x] list-secrets
* [x] put-resource-policy
* [x] put-secret-value
* [x] remove-regions-from-replication
* [x] replicate-secret-to-regions
* [x] restore-secret
* [x] rotate-secret
* [x] stop-replication-to-replica
* [x] tag-resource
* [x] untag-resource
* [x] update-secret
* [x] update-secret-version-stage
* [x] validate-resource-policy

**Additional info:**


falky97:
This checklist is an excellent format for what needs to be done.

QuiiBz:
Hi! @nagypeterjob already made the `delete-secret` generator in #166, which one should be kept?

nagypeterjob:
Hi @QuiiBz 👋🏻
I will close #166 in favour of this PR.

QuiiBz:
Alright!

mattschrage (changes requested):
Overall looks great. A couple things to note:
1. There is a much easier approach for suggesting `json` files.
2. For really large CLI tools, it is best to define all generators at the top of the file. Seems like you do this for the most part.
3. How did you decide the `priority` level to grant various options?

If you just replace the `inputJSONGenerator` and `getBlobsGenerator` we can merge this in!

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,348 @@
...
+        console.error(error);
+      }
+      return [];
+    },
+    cache: {
+      ttl: 30000,
+    },
+  },
+  // List all json files in current directory
+  inputJSONGenerator: {
```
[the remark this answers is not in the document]
nagypeterjob: The function name was unlucky, now I see it could lead to confusion. I have added comments & renamed the function to be more clearer of why I do it this way. 
Context: https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-parameters-file.html
Some options need the `file://` or `fileb://` prefix when providing a filepath. Basically this functionality is  the reason why @falky97 created that example trigger script. 

Alternatively, I could just use the `filepaths` template, and just append a prefix on insert time using the `insertValue` property, but I feel this is the right way. 

I hope I cleared up some confusion.

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,348 @@
...
+        return true;
+      } else return false;
+    },
+
+    filterTerm: (token) => {
+      if (!token.startsWith(_prefix_file)) return token;
+      return token.slice(token.lastIndexOf("/") + 1);
+    },
+  },
+  getBlobsGenerator: {
```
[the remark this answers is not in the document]
nagypeterjob: see above comment

dev/aws/secretsmanager.ts
```diff
@@ -941,6 +1365,29 @@ export const completionSpec: Fig.Spec = {
             "A list of tag key names to remove from the secret. You don't specify the value. Both the key and its associated value are removed. This parameter to the API requires a JSON text string argument. For information on how to format a JSON parameter for the various command line tool environments, see Using JSON for Parameters in the AWS CLI User Guide.",
           args: {
             name: "list",
+            generators: {
+              custom: async function (context, executeShellCommand) {
+                try {
```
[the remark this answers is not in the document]
nagypeterjob: noted, I moved up the generator

nagypeterjob:
To address the priority stuff: I have gave high priority to options which are not optional, or have some kind of dependency between them.
e.g: `aws secretmanager cancel-rotate-secret` won't do anything without `--secret-id`, so I have added it a 100 priority to get to the top.

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+  "eu-central-1",
+  "us-east-1",
+  "us-east-2",
+  "us-west-1",
+  "us-west-2",
+];
+
+const appendFolderPath = (tokens: string[], prefix: string): string => {
+  var baseLSCommand = "\\ls -1ApL ";
+  var whatHasUserTyped = tokens[tokens.length - 1];
```
cstrnt, line 27: I would prefer using `const` over `var` :)

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+  let folderPath = "";
+  const lastSlashIndex = whatHasUserTyped.lastIndexOf("/");
+
+  if (lastSlashIndex > -1) {
+    if (whatHasUserTyped.startsWith("~/"))
+      folderPath = whatHasUserTyped.slice(0, lastSlashIndex + 1);
+    else if (whatHasUserTyped.startsWith("/")) {
+      if (lastSlashIndex === 0) folderPath = "/";
+      else folderPath = whatHasUserTyped.slice(0, lastSlashIndex + 1);
+    } else folderPath = whatHasUserTyped.slice(0, lastSlashIndex + 1);
```
cstrnt, line 44: Could you please add some curly brackets here, this is hard to read imho

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+      },
+    ];
+  }
+  const sortFnStrings = (a, b) => {
+    return a.localeCompare(b);
+  };
+
+  const alphabeticalSortFilesAndFolders = (arr) => {
+    var dots_arr = [];
+    var other_arr = [];
```
cstrnt, line 65: `const` pls

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+    return [
+      ...other_arr.sort(sortFnStrings),
+      "../",
+      ...dots_arr.sort(sortFnStrings),
+    ];
+  };
+
+  var temp_array = alphabeticalSortFilesAndFolders(out.split("\n"));
+
+  var final_array = [];
```
cstrnt, line 82: const :)

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+        return list.map((item) => ({
+          name: item["Name"],
+        }));
+      } catch (error) {
+        console.error(error);
+      }
+      return [];
+    },
+    cache: {
+      ttl: 30000,
```
cstrnt, line 138: This is reused 7 times in this spec. I think it would make sense to create a top-level const for this which makes feature changes easier

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+
+const filterWithPrefix = (token: string, prefix: string): string => {
+  if (!token.startsWith(prefix)) return token;
+  return token.slice(token.lastIndexOf("/") + 1);
+};
+
+const _prefixFile = "file://";
+const _prefixFileb = "fileb://";
+
+const generators: Record<string, Fig.Generator> = {
```
cstrnt, line 121: The problem with using `Record<string, Fig.Generator` is that TypeScript doesn't warn you if you access a property that doesn't exist! If I use `generators.cstrntIsADeveloper` TypeScript thinks it's a valid property because `cstrntIsADeveloper` is a string! Therefore I would recommend either splitting the generators up to individual variables or using a string union like this `Record<'secretIdsGenerator' | 'kmsKeyIdGenerator' | ..., Fig.Generator>`

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+  },
+  getReplicaRegionsGenerator: {
+    script: "aws kms list-keys --page-size 100",
+    postProcess: function (out, context) {
+      try {
+        const list = JSON.parse(out)["Keys"];
+        return list.flatMap((secret) => {
+          return awsRegions.flatMap((region) => {
+            return {
+              name: "Region=" + region + ",KmsKeyId=" + secret.KeyId,
```
cstrnt, line 208: I would prefer using a Template String here 
`Region=${region},KmsKeyId=${secret.KeyId}`
Makes it more readable imho

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+  getVersionIdGenerator: {
+    custom: async function (context, executeShellCommand) {
+      try {
+        // secret-id value
+        const idx = context.indexOf("--secret-id");
+        if (idx < 0) {
+          return [];
+        }
+        const secretId = context[idx + 1];
+        var out = await executeShellCommand(
```
cstrnt, line 247: Const :) (I know I'm annoying, sorry)

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+  getVersionStageGenerator: {
+    custom: async function (context, executeShellCommand) {
+      try {
+        // secret-id value
+        const idx = context.indexOf("--secret-id");
+        if (idx < 0) {
+          return [];
+        }
+        const secretId = context[idx + 1];
+        var out = await executeShellCommand(
```
cstrnt, line 270: const 😊

dev/aws/secretsmanager.ts
```diff
@@ -1,3 +1,312 @@
...
+  listTagKeys: {
+    custom: async function (context, executeShellCommand) {
+      try {
+        // secret-id value
+        const idx = context.indexOf("--secret-id");
+        if (idx < 0) {
+          return [];
+        }
+        const secretId = context[idx + 1];
+        var out = await executeShellCommand(
```
cstrnt, line 294: const :D

cstrnt (changes requested):
Here is my Code Review! I have no experience with AWS and trust you that it is right 😛

cstrnt (approved):
Apart from the `Record<string, Fig.Generator` this looks good to be merged imo

QuiiBz (approved):
Let's merge this amazing PR!

PR #173, opened 2021-05-02
[EXPO] Fix `expo web` aliase to `expo start:web`

### LINKED ISSUE #159: `expo web` command not recognized
`expo web` is an unrecognized alias for `expo start:web`

Matt tells me it should be a simple fix: just finding the `start:web` subcommand and then convert the `name` property from a string to an array ("start:web"  →  ["start:web", "web"])!

QuiiBz:
Fixes #159 

`expo web` is not recognized

`expo web` is recognized

**Additional info:**
Screenshot 2021-05-02 at 08 37 44
I also removed all empty arrays of `subcommands` since that wasn't needed.


dev/expo.ts
```diff
@@ -1,3 +1,89 @@
...
+    args: {},
+  },
+  {
+    name: ["--config"],
+    description: "Specify a path to app.json or app.config.js\n",
+    args: {},
+  },
+  {
+    name: ["-h", "--help"],
+    description: "output usage information\n  \n  \n",
```
cstrnt, line 82: I think we don't need the `\n`s here

dev/expo.ts
```diff
@@ -1,3 +1,89 @@
...
+  {
+    name: ["-w", "--web"],
+    description: "Opens your app in a web browser\n",
+    args: {},
+  },
+  {
+    name: ["-m", "--host"],
+    description:
+      'lan (default), tunnel, localhost. Type of host to use. "tunnel" allows you to view your link on other networks\n',
+    args: {},
```
cstrnt, line 53: Wouldn't it make sense to add `tunnel` `lan` and `localhost` as suggestions here?

cstrnt (changes requested):

QuiiBz:
I copy/pasted the existing spec, didn’t noticed that there were some issues in it. Let me fix them

cstrnt:
Thank you very much, Tom 😊

dev/expo.ts
QuiiBz: Fixed thanks to the power of regex  😄

dev/expo.ts
QuiiBz: Done, looks cool!
Screenshot 2021-05-02 at 12 51 52

mattschrage (dismissed):
Love the addition of the `host` suggestions and thanks for stripping the newline literals!

You can add an alias to a subcommand just by include an array of names in the `name` prop.

So `name: "web"` can become `name: ["start:web", "web"]` 

This just saves you from needing to extract all of the shared options and is a little cleaner! We might need to document this a bit better 😅

QuiiBz:
> Love the addition of the `host` suggestions and thanks for stripping the newline literals!
> 
> You can add an alias to a subcommand just by include an array of names in the `name` prop.
> 
> So `name: "web"` can become `name: ["start:web", "web"]`
> 
> This just saves you from needing to extract all of the shared options and is a little cleaner! We might need to document this a bit better 😅

I see! But currently, the `Subcommand` interface only accepts a string. We will need to update this first in order to change this spec:
Screenshot 2021-05-04 at 09 33 58
Based on. what you said, the type of `name` should be changed to: `SingleOrArray<string>`

cstrnt:
@mattschrage this is new, right? Because I remember bumping into that problem and it didn't work back then.

@QuiiBz I think so!

mattschrage:
> @mattschrage this is new, right? Because I remember bumping into that problem and it didn't work back then.

I just tested it and you're totally right - having an array for the `name` field in a Subcommand isn't currently supported. cc @falky97

falky97:
^^ Yes this is fixed as matt outlined above

QuiiBz:
Sweet! Just updated, ready to merge in.

cstrnt (approved):

cstrnt:
Well done, Tom

PR #181, opened 2021-05-05
[CUT] Add spec for cut

colwin:
Adds specs for the cut command
No specs exist for the cut command
cut command autocompletion is enabled
**Additional info:**
Screenshot 2021-05-05 at 1 23 59 PM

### DIFF
--- a/dev/cut.ts
+++ b/dev/cut.ts
@@ -1,0 +1,76 @@
+export const completionSpec: Fig.Spec = {
+  name: "cut",
+  description: "cut out selected portions of each line of a file",
+  args: {
+    template: "filepaths",
+    isOptional: true,
+    variadic: true,
+  },
+  subcommands: [
+    {
+      name: ["-b"],
+      description: "byte positions as a comma or - separated list of numbers",
+      args: [
+        {
+          name: "list",
+          description: "specifies byte positions",
+        },
+        {
+          name: "-n",
+          description: "do not split multi-byte characters",
+          isOptional: true,
+        },
+        {
+          name: "file",
+          template: "filepaths",
+          isOptional: true,
+          variadic: true,
+        },
+      ],
+    },
+    {
+      name: ["-c"],
+      description: "column positions as a comma or - separated list of numbers",
+      args: [
+        {
+          name: "list",
+          description: "specifies column positions",
+        },
+        {
+          name: "file",
+          template: "filepaths",
+          isOptional: true,
+          variadic: true,
+        },
+      ],
+    },
+    {
+      name: ["-f"],
+      description: "field positions as a comma or - separated list of numbers",
+      args: [
+        {
+          name: "list",
+          description: "specifies column positions",
+        },
+        {
+          name: "-d",
+          description:
+            "use delim as the field delimiter character instead of the tab character.",
+          isOptional: true,
+        },
+        {
+          name: "-s",
+          description:
+            "suppress lines with no field delimiter characters.  unless specified, lines with no delimiters are passed through unmodified.",
+          isOptional: true,
+        },
+        {
+          name: "file",
+          template: "filepaths",
+          isOptional: true,
+          variadic: true,
+        },
+      ],
+    },
+  ],
+};
--- a/dev/cut.ts
+++ b/dev/cut.ts
@@ -20,12 +20,6 @@ export const completionSpec: Fig.Spec = {
           description: "do not split multi-byte characters",
           isOptional: true,
         },
-        {
-          name: "file",
-          template: "filepaths",
-          isOptional: true,
-          variadic: true,
-        },
       ],
     },
     {
@@ -36,16 +30,10 @@ export const completionSpec: Fig.Spec = {
           name: "list",
           description: "specifies column positions",
         },
-        {
-          name: "file",
-          template: "filepaths",
-          isOptional: true,
-          variadic: true,
-        },
       ],
     },
     {
-      name: ["-f"],
+      name: "-f",
       description: "field positions as a comma or - separated list of numbers",
       args: [
         {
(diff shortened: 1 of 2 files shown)


dev/cut.ts
```diff
@@ -0,0 +1,76 @@
...
+  description: "cut out selected portions of each line of a file",
+  args: {
+    template: "filepaths",
+    isOptional: true,
+    variadic: true,
+  },
+  subcommands: [
+    {
+      name: ["-b"],
+      description: "byte positions as a comma or - separated list of numbers",
```
QuiiBz, line 12: This shouldn't be a `subcommands`, but an `option`. See the [docs](https://withfig.com/docs/autocomplete/guides/building-first-spec#3-adding-options-and-args) for a good example, or check an existing spec to see how it's implemented.

dev/cut.ts
```diff
@@ -0,0 +1,76 @@
...
+          name: "file",
+          template: "filepaths",
+          isOptional: true,
+          variadic: true,
+        },
+      ],
+    },
+    {
+      name: ["-c"],
+      description: "column positions as a comma or - separated list of numbers",
```
QuiiBz, line 33: Same as above. Also, you can safely remove the brackets around the same, since you have only one value:
```suggestion
 {
      name: "-c",
      description: "column positions as a comma or - separated list of numbers",
```

dev/cut.ts
```diff
@@ -0,0 +1,76 @@
...
+          name: "file",
+          template: "filepaths",
+          isOptional: true,
+          variadic: true,
+        },
+      ],
+    },
+    {
+      name: ["-f"],
+      description: "field positions as a comma or - separated list of numbers",
```
QuiiBz, line 49: Same as above

QuiiBz (changes requested):

colwin:
Got it. I’ll update the PR in the morning. I couldn’t find an example where an option had multiple arguments. But I’ll try making that a list under the option and update the PR. 

Thank you for taking the time to review and provide feedback @QuiiBz

QuiiBz:
The `Fig.Option` object provides an `args` option, which can take a single `Fig.Arg` object or an array of them (docs: https://withfig.com/docs/autocomplete/api#option-object). Feel free to join the [Slack community](Community) if you need more help!

dev/cut.ts
colwin: Updated this to be an option. Updated name to not have [] around it anymore

dev/cut.ts
colwin: Updated this to be an option. Updated name to not have [] around it anymore

dev/cut.ts
colwin: Updated this to be an option. Updated name to not have [] around it anymore

colwin:
@QuiiBz  - Have updated the PR to address your review comments! Please let me know if there is anything else I need to change to bring it up to spec for a merge.

QuiiBz (approved):
I'm not very familiar with the `cut` CLI, but this looks good to me. Let's wait for another approbation just to make sure everything is correct.

falky97:
Needs to be changed... I will update and add info to docs

colwin:
@falky97 - Updated to address your feedback. Can you please take another look and approve if its good to merge?

dev/cut.ts
```diff
@@ -0,0 +1,58 @@
...
+      args: [
+        {
+          name: "list",
+          description: "specifies byte positions",
+        },
+        {
+          name: "-n",
+          description: "do not split multi-byte characters",
+          isOptional: true,
+        },
```
cstrnt, line 22: Hey there, I know this is kinda tricky, but as a rule of thumb: everything that starts with a `-` or `--` is an option

dev/cut.ts
```diff
@@ -0,0 +1,58 @@
...
+          description:
+            "use delim as the field delimiter character instead of the tab character.",
+          isOptional: true,
+        },
+        {
+          name: "-s",
+          description:
+            "suppress lines with no field delimiter characters.  unless specified, lines with no delimiters are passed through unmodified.",
+          isOptional: true,
+        },
```
cstrnt, line 54: See above

dev/cut.ts
```diff
@@ -0,0 +1,58 @@
...
+      args: [
+        {
+          name: "list",
+          description: "specifies column positions",
+        },
+        {
+          name: "-d",
+          description:
+            "use delim as the field delimiter character instead of the tab character.",
+          isOptional: true,
```
cstrnt, line 47: This should be an option and it also takes an argument called `delim`
Screenshot 2021-05-11 at 19 36 12

cstrnt (changes requested):
Hey there, I know this is a really tricky spec and kinda hard to mirror the behaviour into fig :)

dev/cut.ts
colwin: I've converted them all to options instead of args

dev/cut.ts
colwin: converted to options

dev/cut.ts
colwin: converted to options

colwin:
@QuiiBz @falky97 Pushed a new commit fixing your review comments. The - arguments are now all options to cut

dev/cut.ts
```diff
@@ -0,0 +1,62 @@
...
+  },
+  options: [
+    {
+      name: "-b",
+      description: "byte positions as a comma or - separated list of numbers",
+      args: [
+        {
+          name: "list",
+          description: "specifies byte positions",
+        },
```
cstrnt, line 17: This is reused three times, I would prefer having a top-level const called 
`const listArg: Fig.Arg = { // Code goes here }` which is then referenced

cstrnt (dismissed):
Hey there :) Just a little change which would make this Spec even better :)

mattschrage (approved):
Looks good to me!

dev/cut.ts
mattschrage: This is nice to have but not essential for us to merge in!

PR #186, opened 2021-05-07
[JEST] Improve Jest spec

QuiiBz:
Feature

Options of Jest's spec doesn't have any template

Adds `filepath` template to needed options to get autocompletion on them

**Additional info:**

### DIFF
--- a/dev/jest.ts
+++ b/dev/jest.ts
@@ -69,14 +69,14 @@ export const completionSpec: Fig.Spec = {
         "The path to a Jest config file specifying how to find and execute tests",
       args: {
         name: "path",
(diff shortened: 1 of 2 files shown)


cstrnt (approved):
While this PR makes the spec already waaay better, there are still some options that still lack `templates` or the variadic flag. Do you think we should tackle this in this PR or create another one

QuiiBz:
I tried adding the `variadic` option, but it wasn't providing autocompletion with `templates`. Could you try yourself this?I was maybe missing something

QuiiBz:
I will create another PR to further improve this spec.

Issue #176, opened 2021-05-03
[Bug] autocomplete specs does not show with sudo


dipakparmar:
### Completion Spec

What completion spec does this bug refer to? Or is it Fig itself?

**_Fig itself_**

### Description

Describe the issue that you're seeing.

**_No autocomplete_**

### Steps to reproduce

Clear steps describing how to reproduce the issue. **Please** link to a reproduction, this makes your issue _much_ easier to diagnose (seriously).

**_Use <kbd>sudo</kbd> before any command_**

### Expected result

What should happen?

**_Should ignore sudo or change permission to run the script and then ignore sudo and run autocomplete specs with sudo permission_**

### Actual result

What happened.

**_No autocomplete spec_**

### System Information

*Fig Version:*  1.0.40 (B188)

*OS:* macOS Big Sur 11.3

*Device:* macBook

### Potential Solution

(optional) What could be a solution to solve this bug?

**_Should ignore sudo or change permission to run the script and then ignore sudo and run autocomplete specs with sudo permission_**

mattschrage:
Thanks for reporting @dipakparmar! So the issue here is that Fig is looking at the tokenized edit buffer `["sudo", ...]`, trying to load a completion spec for `sudo` and then not finding one.

I think the fix here is pretty simple. Create a `sudo` completion spec and make sure to add an argument that has the `isCommand` property set.
```
args: {
  name: "command",
  description: "command to run with elevated permissions"
  isCommand: true
}
```

CastaChick:
Hi, May I try to work on this issue?

dipakparmar:
@mattschrage this seems fixed on the latest update `v1.0.41` 

Thank you! 😀

PR #189, opened 2021-05-08
[Sudo] feat: Add sudo spec

CastaChick:
Add spec for `sudo` 

#176 

Show autocompletes after `sudo`

スクリーンショット 2021-05-08 17 22 19

**Additional info:**

### DIFF
--- a/dev/sudo.ts
+++ b/dev/sudo.ts
@@ -1,8 +1,29 @@
 export const completion: Fig.Spec = {
   name: "sudo",
-  description: "",
+  description: "execute a command as the superuser or another user",
   subcommands: [],
-  options: [],
+  options: [
+    {
+      name: ["-g"],
+      description: "run command as the specified group name or ID",
+      args: {
+        name: "group",
+        description: "group name or ID",
+      },
+    },
+    {
+      name: ["-h"],
+      description: "display help message and exit",
+    },
+    {
+      name: ["-u"],
+      description: "run command as specified user name or ID",
+      args: {
+        name: "user",
+        description: "user name or ID",
+      },
+    },
+  ],
   // Only uncomment if sudo takes an argument
   args: {
     name: "command",
--- a/dev/sudo.ts
+++ b/dev/sudo.ts
@@ -1,0 +1,12 @@
+export const completion: Fig.Spec = {
+  name: "sudo",
+  description: "",
+  subcommands: [],
+  options: [],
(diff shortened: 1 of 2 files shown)


CastaChick:
This is my first contribution to Fig ✨
If I should fix something, I'm willing to fix it 😃

dev/sudo.ts
```diff
@@ -0,0 +1,33 @@
+export const completion: Fig.Spec = {
+  name: "sudo",
+  description: "execute a command as the superuser or another user",
+  subcommands: [],
```
cstrnt, line 4: this can safely be removed 😊

dev/sudo.ts
```diff
@@ -0,0 +1,33 @@
...
+  options: [
+    {
+      name: ["-g"],
```
cstrnt, line 7: I think it would be awesome if you could also add the long variant (--group) here

dev/sudo.ts
```diff
@@ -0,0 +1,33 @@
...
+      description: "run command as the specified group name or ID",
+      args: {
+        name: "group",
+        description: "group name or ID",
+      },
+    },
+    {
+      name: ["-h"],
```
cstrnt, line 15:
```suggestion
      name: ["-h", "--help"],
```

dev/sudo.ts
cstrnt, line 7:
```suggestion
      name: ["-g", "--group"],
```

dev/sudo.ts
```diff
@@ -0,0 +1,33 @@
...
+      description: "display help message and exit",
+    },
+    {
+      name: ["-u"],
```
cstrnt, line 19:
```suggestion
      name: ["-u", "--other-user"],
```

cstrnt (changes requested):
Hey there, thank you very much for contributing! This is a really nice PR which will make our autocompletion even better 😊 I left some comments which are just some minor improvements.

dev/sudo.ts
CastaChick: Is it suitable `--user` instead of `--other-user`?
I fixed it by adding `--user`😃

CastaChick:
@cstrnt Thanks for your review!
I've fixed proposed changes 👍

cstrnt (approved):
Thank you very much for your contribution!

QuiiBz (approved):
Thanks for your first contribution!

PR #194, opened 2021-05-11
[BREW] Add analytics subcommand

xxczaki:

I added the `brew analytics <on|off>` command to Fig autocomplete specs. People like to turn the analytics off for privacy reasons.

Fig `brew` autocomplete does not list this subcommand.

Fig will show the autocomplete for the following:

- `brew analytics`
- `brew analytics on`
- `brew analytics off`

**Additional info:**

N/A

### DIFF
--- a/specs/brew.js
+++ b/specs/brew.js
@@ -228,12 +228,10 @@ var completionSpec = {
             subcommands: [
                 {
                     name: "on",
-                    insertValue: "on",
                     description: "Turns on analytics",
                 },
                 {
                     name: "off",
-                    insertValue: "off",
                     description: "Turns off analytics",
                 },
             ],
--- a/specs/brew.js
+++ b/specs/brew.js
@@ -222,6 +222,22 @@ var completionSpec = {
                 },
             ],
         },
+        {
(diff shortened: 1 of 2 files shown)


specs/brew.js
```diff
@@ -222,6 +222,22 @@ var completionSpec = {
                 },
             ],
         },
+        {
+            name: "analytics",
+            description: "Manages analytics preferences",
+            subcommands: [
+                {
+                    name: "on",
+                    insertValue: "on",
```
QuiiBz, line 231: No need for this when it's the same value as `name`

specs/brew.js
```diff
@@ -222,6 +222,22 @@ var completionSpec = {
...
+                    description: "Turns on analytics",
+                },
+                {
+                    name: "off",
+                    insertValue: "off",
```
QuiiBz, line 236: Same

QuiiBz (changes requested):
Could you add the analytics subcommand to the `brew.ts` file instead of the js one? Then, compile it to js with `npm run build`.

PR #195, opened 2021-05-11
Update fastly spec to 0.28.0

kailan:
0.28.0 includes a refactor of some of our `compute` commands. Options from `compute init` have moved to `compute deploy` and `compute publish`.


cstrnt (dismissed):
Just some small tweaks which are mostly personal preference, but I think it should be a small tweak in your parser 😎

PR #200, opened 2021-05-15
[Poetry] Add spec for poetry

CastaChick:
feature

no spec for poetry

add new spec for poetry

**Additional info:**
poetry is package manager for Python

### DIFF
--- a/dev/poetry.ts
+++ b/dev/poetry.ts
@@ -519,6 +519,83 @@ export const completion: Fig.Spec = {
         },
       ],
     },
+    {
+      name: "shell",
+      description: "Spawns a shell within the virtual environment",
+      options: [...globalOptions],
+    },
+    {
+      name: "show",
+      description: "Shows information about packages",
+      options: [
+        ...globalOptions,
+        {
+          name: ["--no-dev"],
+          description: "Do not list the development dependencies",
+        },
+        {
+          name: ["-t", "--tree"],
+          description: "List the dependencies as a tree",
+        },
+        {
+          name: ["-l", "--latest"],
+          description: "Show the latest version",
+        },
+        {
+          name: ["-o", "--outdated"],
+          description:
+            "Show the latest version but only for packages that are outdated",
+        },
+        {
+          name: ["-a", "--all"],
+          description: "Show all packages",
+        },
+      ],
+      args: {
+        name: "dependency",
+        description: "Package to inspect",
+      },
+    },
+    {
+      name: "update",
+      description:
+        "Update the dependencies as according to the pyproject.toml file",
+      options: [
+        ...globalOptions,
+        {
+          name: ["--no-dev"],
+          description: "Do not update the development dependencies",
+        },
+        {
+          name: ["--dry-run"],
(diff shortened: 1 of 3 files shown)


specs/fig.js
```diff
@@ -102,7 +102,7 @@ var completionSpec = {
                     },
                 },
                 {
-                    name: "autocomplete.immediatelyExecute",
+                    name: "autocomplete.immediatelyExecuteAfterSpace",
```
QuiiBz, line 105: Why did you update this?

dev/poetry.ts
```diff
@@ -0,0 +1,603 @@
...
+
+const noAnsi = {
+  name: ["--no-ansi"],
+  description: "Disable ANSI output",
+};
+
+const noInteraction = {
+  name: ["-n", "--no-interaction"],
+  description: "Do not ask any interactive question",
+};
```
QuiiBz, line 44: Could you add the `Fig.Option` type to these options?

dev/poetry.ts
```diff
@@ -0,0 +1,603 @@
...
+        {
+          name: ["-E", "--extras"],
+          description: "Extras to activate for the dependency",
+          args: {
+            name: "extras",
+            description: "Extras",
+          },
+        },
+        {
+          name: ["--optional"],
```
QuiiBz, line 88: No need to specify an array, you can simply write it as a string since there is only one name. Same for underneath

QuiiBz (changes requested):
Thanks for this big PR, I left a few improvements.

specs/fig.js
CastaChick: When I create my feature branch, `specs/fig.js` and `dev/fig.ts` had different value.
It fixed by https://github.com/withfig/autocomplete/commit/a39c26e6f38a05b4b6df13ea78dcd38abd784d0b.

Could you show me how can I do it?

CastaChick:
@QuiiBz Thanks for your review😃
I fixed requested changes!

specs/fig.js
QuiiBz: You can update your branch to the latest Fig's master branch to get these changes.

CastaChick: I've fetched and merged the latest Fig's master branch.
However, the diff is still displayed. 
I tried to revert the commit, but it does not work well :(
Must I revert or squash some commits and force-push?

QuiiBz: You just need to merge `master` into `feat/add_poetry_spec` and push.

CastaChick: > You just need to merge master into feat/add_poetry_spec and push.
I did so, but it seems not reflected to this PR.

I tried reopening another PR from the same branch, the changed files are two.
Is it an issue of Github? Shall I reopen another PR?

QuiiBz: Strange, looks like it's an issue of GitHub. Could you open a new PR and close this one please? With a # to the new one.

CastaChick: I created another issue #201 ☺️

PR #201, opened 2021-05-15
[Poetry] Add spec for Poetry

CastaChick:
feature

no spec for poetry

add new spec for poetry

**Additional info:**
poetry is package manager for Python

This PR is the same of #200 because of an issue of Github.


QuiiBz (approved):
Thanks for this contribution! (Weird GitHub bug for the last PR, it's resolved now)

dev/poetry.ts
```diff
@@ -0,0 +1,603 @@
...
+        },
+        {
+          name: "value",
+          description: "Setting value",
+        },
+      ],
+    },
+    {
+      name: "debug",
+      description: "Debug various elements of Peotry",
```
cstrnt, line 194: typo

dev/poetry.ts
```diff
@@ -0,0 +1,603 @@
...
+            suggestions: ["requirements.txt"],
+          },
+        },
+        {
+          name: ["-o", "--output"],
+          description: "The name of the output file",
+          args: {
+            name: "filename",
+            description: "Output file name",
+            suggestions: ["requirements.txt"],
```
cstrnt, line 263: I think rather than having this suggestion, I would use `template: filepaths` to suggest files

cstrnt (changes requested):
Thank you very much for your contribution

CastaChick:
Thank you for your review! I fixed them☺️

QuiiBz:
> Today, when I run `npm run build`, the output file is differ from the previous one. Is it OK?

Looks like it only affects spacing, it looks OK to me.

cstrnt (approved):
Well done :)

PR #205, opened 2021-05-18
feat: add no empty values rule

cstrnt:
Feature

Right now there are a lot of specs which have empty arrays for `arg`, `subcommands`, `option` which doesn't have any impact on fig and just bloats the code

A new eslint rule which warns the user and autofixes this as well by removing the empty arrays

**Additional info:**


scripts/eslint-plugin-fig-linter/rules/no-empty-array-values.js
```diff
@@ -0,0 +1,26 @@
+module.exports = {
+    meta: {
+        type: "problem",
+        fixable: "code",
+    },
+    create: function (context) {
+        return {
+            Property(node) {
+                if (node.key.name === "options" || node.key.name === "subcommand" || node.key.name === "args") {
```
QuiiBz, line 9: Shouldn’t `subcommand` be `subcommands`?

cstrnt: you are right, good finding Tom :)

scripts/eslint-plugin-fig-linter/rules/no-empty-array-values.js
```diff
@@ -0,0 +1,31 @@
...
+    fixable: "code",
+  },
+  create: function (context) {
+    return {
+      Property(node) {
+        if (
+          node.key.name === "options" ||
+          node.key.name === "subcommands" ||
+          node.key.name === "args"
+        ) {
```
QuiiBz, line 13: Can be simplified to the above if you want:
```suggestion
        if (["options", "subcommands", "args"].includes(node.key.name)) {
```

QuiiBz (approved):

Issue #81, opened 2021-03-30
[git] files already git added should appear lower on autocomplete list of files


veronicanutting:
Hey!

When i am adding files with git add, even when i already added them they still show up as top three. So i am having to use my arrow to get to new ones to add. I am on version .38

![Screen Shot 2021-03-30 at 6 17 06 PM](https://user-images.githubusercontent.com/31642442/113058595-b72fb600-9184-11eb-96c3-d21d8b873d9e.png)

falky97:
Agree, this isn't ideal - we will take a look!!

Brendan Falk
Co-founder & CEO @ Fig
***@***.***
(857) 800-1383

cstrnt:
Hello Veronica, 

thank you very much for reporting :) I just created a PR #208 
I'll let you know when it's merged :)

falky97:
Merged!

PR #208, opened 2021-05-19
[GIT] feat: prioritise  suggestions based on if they are already an argument

cstrnt:
Feature
See #81

Files which are already added with `git add` will have lower priority

**Additional info:**


dev/git.ts
```diff
@@ -130,13 +130,12 @@ const gitGenerators: Record<string, Fig.Generator> = {
 
   // Files for staging
   files_for_staging: {
-    script: "git status --short",
-    postProcess: function (out) {
+    custom: async (context, runFigCommand) => {
```
falky97, line 133: Keep it as a `script` and `postProcess`. postProcess takes context as the second argument. See https://fig.io/docs/autocomplete/schemas/rest#postprocess

custom generators are finicky. I am hesitant to push a custom generator live this close to launch.

falky97 (commented):
So long as you change it back to script and postprocess, this seems good. Please test it yourself though!!

dev/git.ts
cstrnt: Damn, stupid me! I was looking at the wrong docs
Screenshot 2021-05-23 at 12 23 07

QuiiBz (approved):
Tested and working fine on mac too, cool addition!

falky97:
Good stuff!

PR #211, opened 2021-05-19
[CODE] Fill out completion spec for code (VS Code)

### LINKED ISSUE #210: [CODE] Complete VS Code spec
### CLI Tool

The `code` spec is very minimal right now

Tyriar:
This implements the whole spec for `code` based on `code --help`

Fixes #210


dev/code.ts
```diff
@@ -1,40 +1,221 @@
...
+  {
+    name: ["-d", "--diff"],
+    description: "Compare two files with each other.",
+    args: [
+      {
+        template: "filepaths",
+      },
+      {
+        template: "filepaths",
+      },
```
QuiiBz, line 11: I think we should add `name` and `description` to all arguments, to get better feedback in the autocomplete box. Same for all other `args` objects.

QuiiBz (commented):
Thanks for your contribution!

dev/code.ts
```diff
@@ -1,40 +1,228 @@
...
         {
-          template: "filepaths",
+          name: "--category",
+          description: "Filters installed extensions by provided category",
         },
         {
-          template: "filepaths",
+          name: "--show-versions",
+          description:
+            "Show versions of installed extensions, when using --list-extensions.",
```
cstrnt, line 106: Hey there, this might sound weird, but those are `options`and not `args`. We don't offer the option to have conditional `options`

cstrnt: As a rule of thumb: If it starts with `-` or `--` it's an option

dev/code.ts
```diff
@@ -1,40 +1,228 @@
...
+    description: "Print version.",
+  },
+  {
+    name: "--verbose",
+    description: "Print verbose output (implies --wait).",
+  },
+  {
+    name: "--log",
+    description: "Log level to use. Default is 'info'.",
+    args: {
```
cstrnt, line 147: This should be option because it has a default value if omited

dev/code.ts
```diff
@@ -1,40 +1,228 @@
...
+      name: "port",
+    },
+  },
+  {
+    name: "--disable-gpu",
+    description: "Disable GPU hardware acceleration.",
+  },
+  {
+    name: "--max-memory",
+    description: "Max memory size for a window (in Mbytes).",
```
cstrnt, line 209: This should take an argument

dev/code.ts
```diff
@@ -1,40 +1,228 @@
...
+  {
+    name: "--list-extensions",
+    description: "List the installed extensions.",
+    args: {
+      isOptional: true,
+      suggestions: [
         {
-          template: "filepaths",
+          name: "--category",
+          description: "Filters installed extensions by provided category",
```
cstrnt, line 101: this should take an argument, right?

cstrnt (changes requested):
Hey there 👋 
thank you very much for you contribution :) I just left some comments, but apart from that that's a really nice PR for a really great IDE 😉

dev/code.ts
Tyriar: The wording is a little awkward, but I think this is correct as mandatory. It's calling out the default log level if you do not specify `--log`

dev/code.ts
Tyriar: How do I specify nested options? These 2 are only valid when --list-extensions is provided

dev/code.ts
Tyriar: Yep, pending nesting options question

dev/code.ts
Tyriar: If these don't exist I can just make it top level

Tyriar:
Fixed this up, also added suggestions for `--category`

dev/code.ts
cstrnt: There are no nested options as of right now :/

dev/code.ts
```diff
@@ -1,40 +1,253 @@
...
+      },
+    ],
+  },
+  {
+    name: ["-a", "--add"],
+    description: "Add folder(s) to the last active window",
+    args: {
+      name: "folder",
+      template: "folders",
+    },
```
QuiiBz, line 22: Shouldn't this be `variadic`? Since the description contains "folder(s)".

Tyriar: I added `variadic: true` but the file template is still showing up?

Screen Shot 2021-05-22 at 4 50 31 AM
Screen Shot 2021-05-22 at 4 50 37 AM

QuiiBz: [Variadic](https://fig.io/docs/autocomplete/schemas/arg#variadic) means that an argument repeats infinitely, so it should autocomplete folders infinitely. It does not work?

Tyriar: Yeah it didn't work, it may be a bug with fig? I committed the change regardless as it seems like it's correct.

cstrnt: That's really weird and seems to be a bug! We'll investigate!

cstrnt:
@Tyriar 
Hey there, if you like you could merge in the current master and remove the `code.js` from the git index (`git rm -f --cached specs/code.js`:) I think this PR is ready to get merged though, because it's a Bug with Fig itself and not your Spec!

QuiiBz (approved):
Thanks for this PR, ready to merge @cstrnt

cstrnt:
Opened #219

PR #214, opened 2021-05-19
[FASTLY] Update fastly spec to 0.30.0

kailan:
0.30.0 adds a `fastly compute pack` command for packaging compiled binaries into Fastly-ready packages.


kailan:
I think your linting rules have changed, hence the massive diff on the js file.

dev/fastly.ts
```diff
@@ -610,6 +610,23 @@ export const completionSpec: Fig.Spec = {
             },
           ],
         },
+        {
+          name: "pack",
+          description:
+            "Package a pre-compiled Wasm binary for a Fastly Compute@Edge service",
+          subcommands: [],
```
QuiiBz, line 617: You can safely remove `subcommands` since it's empty.

QuiiBz (changes requested):

dev/fastly.ts
cstrnt: #205 will fix this :)

cstrnt (approved):
Apart from Tom's comment this looks good to me :)

kailan:
Cool, I'll wait on #205 and then rebase this PR on the new changes.

mattschrage:
@kailan just merged #205 - you're good to go!

dev/fastly.ts
```diff
@@ -590,6 +590,23 @@ export const completionSpec: Fig.Spec = {
             },
           ],
         },
+        {
+          name: "pack",
+          description:
+            "Package a pre-compiled Wasm binary for a Fastly Compute@Edge service",
+
+          options: [
```
QuiiBz, line 598: Looks like an empty line as popped here, could you remove it?

kailan: Every other entry in the file is like this – this must have been changed in the recent linting update.

I didn't add the line break myself, it must have been done by the linting tool when I ran `--fix`.

QuiiBz: Strange, we will look into this. Let's merge this!

QuiiBz (approved):

PR #215, opened 2021-05-20
[GIT] add subcommands

### LINKED ISSUE #136: [GIT]: Git completion is missing --force-with-lease option.
See summary.

Alexanderchen929:
Exact subcommands are fetch, merge, pull, status, stage, rebase

Also closes #136


dev/git.ts
```diff
@@ -1,3 +1,5 @@
+import { isOptionalChain, transpileModule } from "typescript";
```
mattschrage, line 1: Is this `import` intentional?

mattschrage (approved):
Wow! Great work - didn't realize how many options were were missing!

dev/git.ts
cstrnt: Nope, those imports are unused. Maybe we should add `noUnusedLocals": true` to our `tsconfig.json`

dev/git.ts
```diff
@@ -766,27 +768,348 @@ export const completionSpec: Fig.Spec = {
...
+          args: {
+            name: "mode",
+            isOptional: true,
+            suggestions: ["rebase-cousins", "no-rebase-cousins"],
+          },
+        },
+
+        {
+          name: ["-x", "--exec"],
```
cstrnt, line 1053: In the docs they always add "" around the commands. Should we add `insertValue: "-x "{cursor}"` here?
Screenshot 2021-05-23 at 12 31 39

dev/git.ts
```diff
@@ -913,39 +1236,196 @@ export const completionSpec: Fig.Spec = {
       },
     },
     {
-      name: "status",
-      description: "Show the working tree status",
+      name: "stage",
+      description: "Add file contents to the staging area",
```
cstrnt, line 1240: Shouldn't we create a top-level const for the options and args of `git add` because they are the same?

cstrnt (changes requested):
Damn, git is such a big tool 😆  I think in general we have loads of code duplication in here and maybe we should think of a way of not repeating ourselves everytime we add something new

dev/git.ts
Alexanderchen929: agreed, think it was vscode trying to be helpful for my typos, will remove

dev/git.ts
Alexanderchen929: that sounds good! Thank you for this suggestion

dev/git.ts
Alexanderchen929: yes, was talking to brendan about this, will change around now. Thank you again!

Ashraf-Ali-aa:
Hi, I was wondering is it possible to add the following commands.
```
	git log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit
	git log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --name-status -p
```

cstrnt (approved):
Did some quick checks and it seems to be good! Well done, Alex 😊

PR #216, opened 2021-05-21
feat: add specs folder to gitignore

cstrnt:
Feature

Contributors need to check in their newly created `.js` files in as well

New PRs only have to include the `.ts` file

**Additional info:**
The files are still in git for compatibility with old versions of Fig


mattschrage (approved):
Merging it in now!

PR #217, opened 2021-05-22
[INTERNAL] Build all specs after install and before dev

mattschrage:
- Run `npm run build` on `postinstall`
- Run `npm run build` as part of `npm run dev`.

Together this ensure that the specs folder always contains the up to date specs. This is especially important for new users who have just cloned the repo and use our `create-example` script. If the user follows our instructions in the README.md, Fig will not work:

`npm run create-example` creates a new file.
The current version `npm run dev` starts the file watcher.

If the user tries to preview the new completion spec before making more changes and saving them, the compiled spec will not yet exist and completions will not appear!


package.json
```diff
@@ -3,7 +3,8 @@
   "version": "1.0.0",
   "description": "Fig Autocomplete Spec Linter",
   "scripts": {
-    "dev": "source scripts/dev.sh",
+    "dev": "\"$npm_execpath\" run build && source scripts/dev.sh",
+    "postinstall" : "\"$npm_execpath\" run build",
```
cstrnt, line 7: Is there any particular reason why we need to use `$npm_execpath run` over `npm run`?

falky97: So it works with `yarn`! Smart catch by @mattschrage

falky97 (approved):
Seems fine. Check wtf is going on with the formatting stuff with all the js files.

PR #225, opened 2021-05-25
[INTERNAL] feat: update no-empty-array-values rule

cstrnt:
Feature

The eslint rule disallows empty `options` and `subcommands` arrays at the root command

Those are now allowed and don't cause an error

**Additional info:**

### DIFF
--- a/scripts/eslint-plugin-fig-linter/rules/no-empty-array-values.js
+++ b/scripts/eslint-plugin-fig-linter/rules/no-empty-array-values.js
@@ -6,11 +6,17 @@ module.exports = {
   create: function (context) {
     return {
       Property(node) {
-        if (
-          node.key.name === "options" ||
-          node.key.name === "subcommands" ||
-          node.key.name === "args"
-        ) {
+        const keyName = node.key.name;
+        if (["options", "subcommands", "args"].includes(keyName)) {
+          // This rule should not be applied to subcommands and
+          // options on the root object
+          if (
+            node.parent.parent.parent.parent.type ===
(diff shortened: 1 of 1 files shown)


scripts/eslint-plugin-fig-linter/rules/no-empty-array-values.js
```diff
@@ -6,11 +6,17 @@ module.exports = {
...
-          node.key.name === "subcommands" ||
-          node.key.name === "args"
-        ) {
+        const keyName = node.key.name;
+        if (["options", "subcommands", "args"].includes(keyName)) {
+          // This rule should not be applied to subcommands and
+          // options on the root object
+          if (
+            node.parent.parent.parent.parent.type ===
+              "ExportNamedDeclaration" &&
```
QuiiBz, line 15: Wouldn't this throw an error if anyone makes a separate object with a property named `options`, `subcommands` or `args`? I think we can use optional chaining at this point.

cstrnt: you are right! Optional chaining doesn't work with plain js and eslint :/ Will need to add a compile/transpile step here. But that's a good find!

QuiiBz: I just started adding TypeScript to eslint-plugin-fig-linter, so it will also support optional chaining.

QuiiBz (approved):
Cool :) We will replace the "pyramid code" with optional chaining once the plugin will be switched to TypeScript.

PR #226, opened 2021-05-25
[INTERNAL] feat: add eslint rule to check if insertValue and name are equal

cstrnt:
Feature

Some specs contain equal `insertValue` and `name` props which doesn't have any effect

There is a new eslint rule which warns the user about that

**Additional info:**


mattschrage:
This is a great idea!

scripts/eslint-plugin-fig-linter/index.js
```diff
@@ -5,13 +5,14 @@ module.exports = {
...
     "no-invalid-name": require("./rules/no-invalid-name"),
     "no-empty-array-values": require("./rules/no-empty-array-values"),
+    "no-useless-insertvalue": require("./rules/no-useless-insertvalue"),
   },
   configs: {
     recommended: {
       plugins: ["fig-linter"],
       rules: {
         "fig-linter/no-malicious-script": "error",
-        "fig-linter/no-empty-array-values": "error",
```
hiradp, line 14: Sorry to chime in, but did you mean to add this rule or change `no-empty-array-values`?

cstrnt: Damn, really good find! I wanted to add the rule :D

QuiiBz: Hi Tim, could you fix this so this could get merged in? It's a really useful addition!

Issue #231, opened 2021-05-26
Add autocomplete support for rclone


rmorey:
### CLI Tool

`rclone` syncs files to and from cloud storage providers as well as mounting them, listing them in lots of different ways.

### Useful Links

- [Homepage](https://rclone.org)
- [GitHub](https://github.com/rclone/rclone)
- [CLI Documentation](https://rclone.org/docs/)
- [`--help` text](https://github.com/rclone/rclone/blob/master/docs/content/commands/rclone.md)

### Top Workflows

- `rclone config`
- `rclone listremotes`
- `rclone lsd remote:path`
- `rclone -P --exclude '.*{/**,}' sync /Volumes/Data/ remote:Data`
- `rclone mount remote: dir`

rmorey:
Working on this here: https://github.com/rmorey/autocomplete/blob/rmorey/rclone/dev/rclone.ts

rmorey:
What I've done so far is just transform the list of commands with descriptions from the help text into the spec format with `args: {}` for all. So this gets us simple autocomplete for every command, but nothing else - @cstrnt is this substantial enough for a PR?

cstrnt:
Hey @rmorey that's awesome! 
Thank you very much for starting! 😊 
While it is totally fine for a spec to just be a skeleton to get merged in, I think this needs some little tweaks.

`args: {}` tells Fig that the command takes an argument. One example for this would be `cd folderName` where `cd`is the name of the command itself and `folderName` the argument which is passed to the command itself. I just looked up `rclone` and saw that not all commands take an argument like that. It would be really cool if you could remove `args: {}` for the commands which don't take an argument. If this is done it would be awesome if you would create a PR :)

Again, thank you very much for your effort and your time! And also let me know if you need some help :)

cstrnt:
Also this page is a really good read 😆 
https://fig.io/docs/concepts/cli-skeleton

rmorey:
> `args: {}` tells Fig that the command takes an argument. ... It would be really cool if you could remove `args: {}` for the commands which don't take an argument. If this is done it would be awesome if you would create a PR :)

Ahhhh, I misread the comment in the boilerplate! I missed the conditional in "*If a subcommand or option takes an argument*, you must include the args prop, even if it's an empty object (like below)" I'll remove those and submit a PR.

> Also this page is a really good read 😆
> https://fig.io/docs/concepts/cli-skeleton

[...]

PR #233, opened 2021-05-26
Add skeleton for rclone

rmorey:

Feature: Basic support for [rclone](https://rclone.org) 

#231 

This adds basic autocomplete for all commands, as listed by `rclone --help`

### DIFF
--- a/dev/rclone.ts
+++ b/dev/rclone.ts
@@ -3,8 +3,253 @@ export const completion: Fig.Spec = {
   description: "The Swiss army knife of cloud storage",
   subcommands: [
     {
+      name: "about",
+      description: "Get quota information from the remote.",
+      args: {},
+    },
+    {
+      name: "authorize",
+      description: "Remote authorization.",
+      args: {},
+    },
+    {
+      name: "backend",
+      description: "Run a backend specific command.",
+      args: {},
+    },
+    {
+      name: "cat",
+      description: "Concatenates any files and sends them to stdout.",
+      args: {},
+    },
+    {
+      name: "check",
+      description: "Checks the files in the source and destination match.",
+      args: {},
+    },
+    {
+      name: "cleanup",
+      description: "Clean up the remote if possible.",
+      args: {},
+    },
+    {
+      name: "config",
+      description: "Enter an interactive configuration session.",
+      args: {},
+    },
+    {
+      name: "copy",
+      description: "Copy files from source to dest, skipping already copied.",
+      args: {},
+    },
+    {
+      name: "copyto",
+      description: "Copy files from source to dest, skipping already copied.",
+      args: {},
+    },
+    {
+      name: "copyurl",
+      description: "Copy url content to dest.",
+      args: {},
+    },
+    {
+      name: "cryptcheck",
+      description: "Cryptcheck checks the integrity of a crypted remote.",
+      args: {},
+    },
+    {
+      name: "cryptdecode",
+      description: "Cryptdecode returns unencrypted file names.",
+      args: {},
+    },
+    {
+      name: "dedupe",
+      description:
+        "Interactively find duplicate filenames and delete/rename them.",
+      args: {},
+    },
+    {
+      name: "delete",
+      description: "Remove the files in path.",
+      args: {},
+    },
+    {
+      name: "deletefile",
+      description: "Remove a single file from remote.",
+      args: {},
+    },
+    {
+      name: "genautocomplete",
+      description: "Output completion script for a given shell.",
+      args: {},
+    },
+    {
+      name: "gendocs",
+      description: "Output markdown docs for rclone to the directory supplied.",
+      args: {},
+    },
+    {
+      name: "hashsum",
+      description: "Produces a hashsum file for all the objects in the path.",
+      args: {},
+    },
+    {
+      name: "help",
+      description: "Show help for rclone commands, flags and backends.",
+      args: {},
+    },
+    {
+      name: "link",
+      description: "Generate public link to file/folder.",
+      args: {},
+    },
+    {
       name: "listremotes",
       description: "List all the remotes in the config file.",
+      args: {},
+    },
+    {
+      name: "ls",
+      description: "List the objects in the path with size and path.",
+      args: {},
+    },
+    {
+      name: "lsd",
+      description: "List all directories/containers/buckets in the path.",
+      args: {},
+    },
+    {
+      name: "lsf",
+      description:
+        "List directories and objects in remote:path formatted for parsing.",
+      args: {},
+    },
+    {
+      name: "lsjson",
+      description: "List directories and objects in the path in JSON format.",
+      args: {},
+    },
+    {
+      name: "lsl",
+      description:
+        "List the objects in path with modification time, size and path.",
+      args: {},
+    },
+    {
+      name: "md5sum",
+      description: "Produces an md5sum file for all the objects in the path.",
+      args: {},
+    },
+    {
+      name: "mkdir",
+      description: "Make the path if it doesn't already exist.",
+      args: {},
+    },
+    {
+      name: "mount",
+      description: "Mount the remote as file system on a mountpoint.",
+      args: {},
+    },
+    {
+      name: "move",
+      description: "Move files from source to dest.",
+      args: {},
+    },
+    {
+      name: "moveto",
+      description: "Move file or directory from source to dest.",
+      args: {},
+    },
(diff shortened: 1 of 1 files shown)


dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "authorize",
+      description: "Remote authorization.",
+      args: {},
+    },
+    {
+      name: "backend",
+      description: "Run a backend specific command.",
+      args: {},
```
cstrnt, line 18: I'm just looking at the documentation and it looks like the root command (e.g. `rclone backend`) itself doesn't take an argument, but rather a subcommand that then takes an argument. So I think this should not have `args: {}`as well

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "cat",
+      description: "Concatenates any files and sends them to stdout.",
+      args: {},
+    },
+    {
+      name: "check",
+      description: "Checks the files in the source and destination match.",
+      args: {},
```
cstrnt, line 28: This is a special case as well, because it takes two mandatory arguments! Could you please change it to
```ts
      args: [{}, {}],
```

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "config",
+      description: "Enter an interactive configuration session.",
+    },
+    {
+      name: "copy",
+      description: "Copy files from source to dest, skipping already copied.",
+      args: {},
```
cstrnt, line 42: Same as the comment for `check`

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "copyto",
+      description: "Copy files from source to dest, skipping already copied.",
+      args: {},
```
cstrnt, line 47: see above

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "copyurl",
+      description: "Copy url content to dest.",
+      args: {},
```
cstrnt, line 52: see above

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "cryptcheck",
+      description: "Cryptcheck checks the integrity of a crypted remote.",
+      args: {},
```
cstrnt, line 57: see above

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "cryptdecode",
+      description: "Cryptdecode returns unencrypted file names.",
+      args: {},
```
cstrnt, line 62: This is another tricky one:

Screenshot 2021-05-27 at 11 32 01

It also takes to arguments, but the second one is variadic (it means it can repeat itself unlimited times, for example `mkdir folder1 folder2 folder3...`


```ts
      args: [{}, {variadic: true}],
```

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "genautocomplete",
+      description: "Output completion script for a given shell.",
+      args: {},
+    },
+    {
+      name: "gendocs",
+      description: "Output markdown docs for rclone to the directory supplied.",
+      args: {},
```
cstrnt, line 88: While this is totally fine to just use `args: {}` here, you could add `template: "folders"` here and Fig will automatically suggest folders here :)

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "hashsum",
+      description: "Produces a hashsum file for all the objects in the path.",
+      args: {},
```
cstrnt, line 93: This is also fine, but you could do the following

```suggestion
      args: {
        suggestions: ["MD5", "SHA-1", "DropboxHash", "QuickXorHash"]
      },
```
to make it even more awesome 🎉

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "mkdir",
+      description: "Make the path if it doesn't already exist.",
+      args: {},
+    },
+    {
+      name: "mount",
+      description: "Mount the remote as file system on a mountpoint.",
+      args: {},
```
cstrnt, line 149: Same as the comment for check

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "move",
+      description: "Move files from source to dest.",
+      args: {},
```
cstrnt, line 154: see above

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "moveto",
+      description: "Move file or directory from source to dest.",
+      args: {},
```
cstrnt, line 159: Same as the comment for check

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+    },
+    {
+      name: "serve",
+      description: "Serve a remote over a protocol.",
+      args: {},
+    },
+    {
+      name: "settier",
+      description: "Changes storage class/tier of objects in remote.",
+      args: {},
```
cstrnt, line 213: this takes two arguments as well `args: [{}, {}]`

dev/rclone.ts
```diff
@@ -0,0 +1,262 @@
...
+      name: "size",
+      description:
+        "Prints the total size and number of objects in remote:path.",
+      args: {},
+    },
+    {
+      name: "sync",
+      description:
+        "Make source and dest identical, modifying destination only.",
+      args: {},
```
cstrnt, line 230: Same as the comment for check

cstrnt (changes requested):
Thank you very much for contributing 😊  I just left some comments which would make this spec even more awesome 🎉

rmorey:
Thanks for the feedback @cstrnt ! I will incorporate these notes

dev/rclone.ts
```diff
@@ -0,0 +1,572 @@
...
+    {
+      name: "copy",
+      description: "Copy files from source to dest, skipping already copied.",
+      args: [
+        {
+          name: "source:path",
+        },
+        {
+          name: "dest:path",
+        },
```
QuiiBz, line 185: Couldn't these args use a `filepaths` or `folders` template?

dev/rclone.ts
```diff
@@ -0,0 +1,572 @@
...
+    {
+      name: "copyto",
+      description: "Copy files from source to dest, skipping already copied.",
+      args: [
+        {
+          name: "source:path",
+        },
+        {
+          name: "dest:path",
+        },
```
QuiiBz, line 197: Same as above.

dev/rclone.ts
```diff
@@ -0,0 +1,572 @@
...
+    {
+      name: "copyurl",
+      description: "Copy url content to dest.",
+      args: [
+        {
+          name: "url",
+        },
+        {
+          name: "dest:path",
+        },
```
QuiiBz, line 209: Same as above.

dev/rclone.ts
QuiiBz: Not sure if `template` works along with `generators`. Could you try it, please?

dev/rclone.ts
```diff
@@ -0,0 +1,697 @@
...
+const hashes = [
+  "MD5",
+  "SHA-1",
+  "DropboxHash",
+  "QuickXorHash",
+  "Whirlpool",
+  "CRC-32",
+  "MailruHash",
+];
+const listOptions = [
```
cstrnt, line 21: Could you please add 
```suggestion
const listOptions: Fig.Option[] = [
```
here :)

cstrnt (changes requested):
Wow! This spec evolved so much! Thank you very much 😄

dev/rclone.ts
```diff
@@ -0,0 +1,697 @@
...
+    name: ["--recursive", "-R"],
+    description: "Recurse into the listing.",
+  },
+];
+const cryptedremote: Fig.Arg = {
+  name: "cryptedremote",
+  generators: remote.generators,
+  // TODO filter by crypted type
+};
+// TODO dynamically sugest path
```
QuiiBz, line 40: Are you going to tackle these TODOs? Just to know if we can merge this PR or wait for this to be implemented.

PR #239, opened 2021-05-27
[YARN] Add --latest flag to upgrade-interactive

nirtamir2:
**What kind of change does this PR introduce? ** feature

**What is the current behavior?**  No flag auto-completion to yarn upgrade-interactive

**What is the new behavior (if this is a feature change)?**  Add --latest flag to yarn upgrade-interactive command

**Additional info:** Haven't tested it. See https://classic.yarnpkg.com/en/docs/cli/upgrade-interactive/#toc-yarn-upgrade-interactive command documentation

### DIFF
--- a/dev/yarn.ts
+++ b/dev/yarn.ts
@@ -961,6 +961,12 @@ export const completionSpec: Fig.Spec = {
     {
       name: "upgrade-interactive",
       description: "Upgrades packages in interactive mode",
+      options: [
+        {
+          name: ["--latest"],
+          description: "Use the version tagged latest in the registry",
+        },
+      ],
     },
     {
       name: "version",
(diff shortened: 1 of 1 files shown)


dev/yarn.ts
```diff
@@ -961,6 +961,12 @@ export const completionSpec: Fig.Spec = {
     {
       name: "upgrade-interactive",
       description: "Upgrades packages in interactive mode",
+      options: [
+        {
+          name: ["--latest"],
```
QuiiBz, line 966: You can set `name` as a string instead of an array since there is only one value.

QuiiBz (changes requested):

dev/yarn.ts
nirtamir2: Fixed

cstrnt (approved):
Thank you very much for your contribution 😊

PR #241, opened 2021-05-27
[ZAPIER] Add Zapier spec

xavdid:

Add spec for the [Zapier CLI](https://github.com/zapier/zapier-platform). It was generated using [the oclif plugin](https://github.com/withfig/oclif-plugin).

We can probably make a fancier version that includes the args, but this is a good first step.

### DIFF
--- a/dev/zapier.ts
+++ b/dev/zapier.ts
@@ -1,3 +1,17 @@
+const formatFlag = {
+  name: ["-f", "--format"],
+  description:
+    'Change the way structured data is presented. If "json" or "raw", you can pipe the output of the command into other tools, such as jq.',
+  args: {
+    name: "format",
+    suggestions: ["plain", "json", "raw", "row", "table"],
+  },
+};
+const debugFlag = {
+  name: ["-d", "--debug"],
+  description: "Show extra debugging output.",
+};
+
 export const completionSpec: Fig.Spec = {
   name: "zapier-platform-cli",
   description:
@@ -14,14 +28,10 @@ export const completionSpec: Fig.Spec = {
             "Choose how much information to share. Anonymous mode drops the OS type and Zapier user id, but keeps command info. Identifying information is used only for debugging purposes.",
           args: {
             name: "mode",
-            isOptional: false,
             suggestions: ["enabled", "anonymous", "disabled"],
           },
         },
-        {
-          name: ["-d", "--debug"],
-          description: "Show extra debugging output.",
-        },
+        debugFlag,
       ],
     },
     {
@@ -33,10 +43,7 @@ export const completionSpec: Fig.Spec = {
           description:
             'Disable "smart" file inclusion. By default, Zapier only includes files that are required by `index.js`. If you (or your dependencies) require files dynamically (such as with `require(someVar)`), then you may see "Cannot find module" errors. Disabling this may make your `build.zip` too large. If that\'s the case, try using the `includeInBuild` option in your `.zapierapprc`. See the docs about `includeInBuild` for more info.',
         },
-        {
-          name: ["-d", "--debug"],
-          description: "Show extra debugging output.",
-        },
+        debugFlag,
       ],
     },
     {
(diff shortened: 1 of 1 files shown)


dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+      description:
+        "Show the status of the analytics that are collected. Also used to change what is collected.",
+      options: [
+        {
+          name: ["-m", "--mode"],
+          description:
+            "Choose how much information to share. Anonymous mode drops the OS type and Zapier user id, but keeps command info. Identifying information is used only for debugging purposes.",
+          args: {
+            name: "mode",
+            isOptional: false,
```
QuiiBz, line 17: No need to set `isOptional` to false, since Fig assumes by default that args are mandatory. Same for underneath

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+          name: "integrationId",
+          description:
+            'To get the integration/app ID, go to "https://zapier.com/app/developer", click on an integration, and copy the number directly after "/app/" in the URL.',
+          isOptional: false,
+        },
+        {
+          name: "path",
+          description:
+            "Relative to your current path - IE: `.` for current directory.",
+          isOptional: false,
```
QuiiBz, line 72: I think you can use the `folders` template here ([doc](https://fig.io/docs/reference/arg#template))

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+      description: "Describe the current integraiton.",
+      options: [
+        {
+          name: ["-f", "--format"],
+          description:
+            'Change the way structured data is presented. If "json" or "raw", you can pipe the output of the command into other tools, such as jq.',
+          args: {
+            name: "format",
+            isOptional: false,
+            suggestions: ["plain", "json", "raw", "row", "table"],
```
QuiiBz, line 139: You can extract this suggestion to a top-level const since it's repeated many times underneath.

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+          name: "version",
+          description:
+            "The version to set the environment for. Values are copied forward when a new version is created, but this command will only ever affect the specified version.",
+          isOptional: false,
+        },
+        {
+          name: "key-value pairs...",
+          description:
+            "The key-value pairs to set. Keys are case-insensitive. Each pair should be space separated and pairs should be separated by an `=`. For example: `A=123 B=456`",
+          isOptional: true,
```
QuiiBz, line 195: I believe this arg should be [variadic](https://fig.io/docs/reference/arg#variadic)?

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+      args: [
+        {
+          name: "version",
+          description: "The version to set the environment for.",
+          isOptional: false,
+        },
+        {
+          name: "keys...",
+          description: "The keys to unset. Keys are case-insensitive.",
+          isOptional: true,
```
QuiiBz, line 217: Same as above

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+          args: {
+            name: "mode",
+            isOptional: false,
+            suggestions: ["enabled", "anonymous", "disabled"],
+          },
+        },
+        {
+          name: ["-d", "--debug"],
+          description: "Show extra debugging output.",
+        },
```
QuiiBz, line 24: You can extract this arg to a top-level const to avoid duplication.

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+          name: ["-h", "--help"],
+          description: "show CLI help",
+        },
+        {
+          name: ["-o", "--output"],
+          description:
+            "filepath to export completion spec (do not specify file extension)",
+          args: {
+            name: "output",
+            isOptional: false,
```
QuiiBz, line 795: You can add the `filepaths` template here.

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+      name: ["plugins:install", "plugins:add"],
+      description: "installs a plugin into the CLI",
+      options: [
+        {
+          name: ["-h", "--help"],
+          description: "show CLI help",
+        },
+        {
+          name: ["-v", "--verbose"],
+          description: "",
```
QuiiBz, line 873: Please fill in this description.

dev/zapier.ts
```diff
@@ -0,0 +1,962 @@
...
+        },
+      ],
+      args: [
+        {
+          name: "path",
+          description: "path to plugin",
+          isOptional: false,
```
QuiiBz, line 905: I assume this arg could have a `folders` template?

QuiiBz (changes requested):
Thanks a lot for your first contribution here!

xavdid:
@QuiiBz thanks for the review! 

I put this together on a call with @mattschrage yesterday using the `oclif-plugin`. I'm happy to make the manual changes you suggested. But, it would also be great to be able to fully automate the process. In the future when we re-generate the spec, it would be great to not need to re-make these changes. Maybe a problem for another day.

That said, those plugin commands can be removed, they aren't meant to be there.

cstrnt (changes requested):

cstrnt:
Hey @xavdid I totally agree with you, that those changes shouldn't be done each time the spec is generated, but on the other hand it will be kinda hard to guess stuff like `template: "filePaths"` just from the naming of the argument. We humans are good at doing that, but oclif-plugins aren't 😆

xavdid:
Cool, i've made changes that _should_ address everything mentioned. Thanks y'all!

dev/zapier.ts
```diff
@@ -0,0 +1,558 @@
+const formatFlag = {
```
cstrnt, line 1: could you please type this variable 
```suggestion
const formatFlag: Fig.Option = {
```

dev/zapier.ts
```diff
@@ -0,0 +1,558 @@
...
+  name: ["-f", "--format"],
+  description:
+    'Change the way structured data is presented. If "json" or "raw", you can pipe the output of the command into other tools, such as jq.',
+  args: {
+    name: "format",
+    suggestions: ["plain", "json", "raw", "row", "table"],
+  },
+};
+const debugFlag = {
```
cstrnt, line 10: see above

cstrnt (changes requested):

xavdid:
@cstrnt done!

dev/zapier.ts
```diff
@@ -0,0 +1,558 @@
...
+      description:
+        "Add a starting trigger, create, search, or resource to your integration.",
+      options: [
+        {
+          name: ["-d", "--dest"],
+          description:
+            "Specify the new file's directory. Use this flag when you want to create a different folder structure such as `src/triggers` instead of the default `triggers`. Defaults to `[triggers|searches|creates]/{noun}`.",
+          args: {
+            name: "dest",
+          },
```
QuiiBz, line 365: I think this can have a template of `folders`

dev/zapier.ts
```diff
@@ -0,0 +1,558 @@
...
+      options: [debugFlag],
+    },
+    {
+      name: ["users:add", "users:invite"],
+      description: "Add a user to some or all versions of your integration.",
+      options: [
+        {
+          name: ["-f", "--force"],
+          description: "Skip confirmation. Useful for running programatically.",
+        },
```
QuiiBz, line 473: Maybe this should be dangerous?

dev/zapier.ts
```diff
@@ -0,0 +1,558 @@
...
+      options: [formatFlag, debugFlag],
+    },
+    {
+      name: ["users:remove", "users:delete"],
+      description: "Remove a user from all versions of your integration.",
+      options: [
+        {
+          name: ["-f", "--force"],
+          description:
+            "Skips confirmation. Useful for running programatically.",
```
QuiiBz, line 509: Same as above

QuiiBz (changes requested):
Hi! Could you fix those small issues, so we can get this merged in? Thanks a lot for this huge contribution!

xavdid:
Sure, good catch!

QuiiBz (approved):
Thanks a lot for your contribution!

QuiiBz:
@cstrnt could we merge this?

cstrnt:
Linting fails for some reason though! @xavdid could you please run `npm run lint:fix` and push again? 😊

xavdid:
I tried both `lint` and `lint:fix` and there were no changes locally 🤔 

```
% npm run lint

> linter@1.0.0 lint /Users/david/projects/autocomplete
> eslint '**/*.ts' && npx prettier --check '**/*.ts'

Checking formatting...
All matched files use Prettier code style!
```

node version is `14.16.1`.

PR #252, opened 2021-06-02
[GIT] Ignore warning, error message

seonggwonyoon:
Bug fix

https://github.com/withfig/fig/issues/107

**Additional info:**

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -1,4 +1,4 @@
-const filter = (out: string): string => {
+const filterWarnings = (out: string): string => {
   return out.startsWith("warning:") || out.startsWith("error:")
     ? out.split("\n").slice(1).join("\n")
     : out;
@@ -9,7 +9,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
   commits: {
     script: "git log --oneline",
     postProcess: function (out) {
-      out = filter(out);
+      out = filterWarnings(out);
 
       if (out.startsWith("fatal:")) {
         return [];
@@ -30,7 +30,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
   stashes: {
     script: "git stash list",
     postProcess: function (out) {
-      out = filter(out);
+      out = filterWarnings(out);
 
       if (out.startsWith("fatal:")) {
         return [];
@@ -55,7 +55,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
   treeish: {
     script: "git diff --cached --name-only",
     postProcess: function (out) {
-      out = filter(out);
+      out = filterWarnings(out);
 
       if (out.startsWith("fatal:")) {
         return [];
@@ -76,7 +76,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
   branches: {
     script: "git branch --no-color",
     postProcess: function (out) {
-      out = filter(out);
+      out = filterWarnings(out);
 
       if (out.startsWith("fatal:")) {
         return [];
@@ -147,7 +147,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
   files_for_staging: {
     script: "git status --short",
     postProcess: (output, context) => {
-      output = filter(output);
+      output = filterWarnings(output);
 
       if (output.startsWith("fatal:")) {
         return [];
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -7,7 +7,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
         return [];
       }
 
-      if (out.startsWith("warning:")) {
+      if (out.startsWith("warning:") || out.startsWith("error:")) {
         out = out.split("\n").slice(1).join("\n");
       }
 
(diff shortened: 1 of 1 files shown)


QuiiBz:
Hi, thanks a lot for your first contribution! Based on Matt's comment (https://github.com/withfig/fig/issues/107#issuecomment-852730125), shouldn't it filter out anything starting with `error` too, not only `warning`?

seonggwonyoon:
> Hi, thanks a lot for your first contribution! Based on Matt's comment ([withfig/fig#107 (comment)](https://github.com/withfig/fig/issues/107#issuecomment-852730125)), shouldn't it filter out anything starting with `error` too, not only `warning`?

Thank you @QuiiBz 
I missed `error` option! just added.

QuiiBz (commented):
I think we can simplify this and create a top-level function to avoid duplication. Something like this (not tested):

```ts
const filter = (out: string): string => {
  return (out.startsWith("warning:") || out.startsWith("error:"))
    ? out = out.split("\n").slice(1).join("\n")
    : out;
}
```

So we will only call `filter` this way:

```ts
out = filter(out);
```

seonggwonyoon:
> I think we can simplify this and create a top-level function to avoid duplication. Something like this (not tested):
> 
> ```ts
> const filter = (out: string): string => {
>   return (out.startsWith("warning:") || out.startsWith("error:"))
>     ? out.split("\n").slice(1).join("\n")
>     : out;
> }
> ```
> 
> So we will only call `filter` this way:
> 
> ```ts
> out = filter(out);
> ```

As you said, I applied the filter function!

QuiiBz (approved):
That's cleaner!

dev/git.ts
```diff
@@ -1,8 +1,16 @@
+const filter = (out: string): string => {
```
cstrnt, line 1: I honestly would prefer being more verbose here: Maybe `filterWarnings`? I think functions having the same name as builtins might confuse other contributors

cstrnt (changes requested):

dev/git.ts
seonggwonyoon: Fix `filter` to `filterWarnings`!

dev/git.ts
```diff
@@ -1,8 +1,16 @@
...
+    ? out.split("\n").slice(1).join("\n")
+    : out;
+};
+
 const gitGenerators: Record<string, Fig.Generator> = {
   // Commit history
   commits: {
     script: "git log --oneline",
     postProcess: function (out) {
+      out = filterWarnings(out);
```
cstrnt, line 12: Could you please create a const here. Re-assigning to a parameter is a code smell imho

dev/git.ts
```diff
@@ -45,6 +55,8 @@ const gitGenerators: Record<string, Fig.Generator> = {
   treeish: {
     script: "git diff --cached --name-only",
     postProcess: function (out) {
+      out = filterWarnings(out);
```
cstrnt, line 58: see above

dev/git.ts
```diff
@@ -64,9 +76,12 @@ const gitGenerators: Record<string, Fig.Generator> = {
   branches: {
     script: "git branch --no-color",
     postProcess: function (out) {
+      out = filterWarnings(out);
```
cstrnt, line 79: see above

dev/git.ts
```diff
@@ -132,9 +147,12 @@ const gitGenerators: Record<string, Fig.Generator> = {
   files_for_staging: {
     script: "git status --short",
     postProcess: (output, context) => {
+      output = filterWarnings(output);
```
cstrnt, line 150: see above

cstrnt (changes requested):

dev/git.ts
QuiiBz: Btw, this function’s name not only filters warnings but also errors. Shouldn’t we rename it to something more generic?

seonggwonyoon: @QuiiBz Yes, it filters out warnings and errors.
Let's think about changing it to a more generic name.

seonggwonyoon: @cstrnt I agree with the saying Re-assigning to a parameter is a code smell.
But I don't understand the phrase `create a const here' Can you explain a bit more?

seonggwonyoon: @cstrnt You mean this code?
```typescript
postProcess: function (out) {
      const ?? = filterWarnings(out);
      if (??.startsWith("fatal:")) {
        return [];
      }
```

cstrnt: hey @seonggwonyoon, sorry for not replying :/ That's exactly what I mean :)

seonggwonyoon: @cstrnt Create const is done!
@QuiiBz Fix filterWarnings to filterMessages

QuiiBz (approved):
LGTM

cstrnt:
Hey sorry to be annoying, but could you please run `npm run lint:fix` in order to to fix the styling :) After this is done, we will merge the PR 😊

seonggwonyoon:
> Hey sorry to be annoying, but could you please run `npm run lint:fix` in order to to fix the styling :) After this is done, we will merge the PR 😊

Fixed it. Thanks @cstrnt

PR #264, opened 2021-06-05
feat: add Gatsby CLI completion spec

indie-ridge-dev:

**Additional info:**

### DIFF
--- a/dev/gatsby.ts
+++ b/dev/gatsby.ts
@@ -1,3 +1,24 @@
+const sharedOptions: Fig.Option[] = [
+  {
+    name: ["-H", "--host"],
+    description: "Set host. Defaults to localhost",
+    args: {
+      name: "host",
+    },
+  },
+  {
+    name: ["-p", "--port"],
+    description: "Set port. Defaults to env.PORT or 8000",
+    args: {
+      name: "port",
+    },
+  },
+  {
+    name: ["-o", "--open"],
+    description: "Open the site in your (default) browser for you",
+  },
+];
+
 export const completion: Fig.Spec = {
   name: "gatsby",
   description: "Gatsby CLI",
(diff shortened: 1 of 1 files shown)


cstrnt (changes requested):
Hey there :)
Could you please take a look at all `args` here? I just left a comment for some, but I seems that you added `args` for everything, even though they don't take any args!


Here is a little explainer, what an argument is:

`cd /some/folder`

`cd` is the command itself and `/some/folder`is the argument which is passed. Another example from this spec is `gatsby develop -H 3000` where the `-H`is the option and `3000` is the argument from the option

dev/gatsby.ts
```diff
@@ -0,0 +1,253 @@
...
+
+        {
+          name: ["staticQueries"],
+          description: "staticQueries",
+          args: {
+            name: "staticQueries",
+          },
+        },
+      ],
+    },
```
QuiiBz, line 245: Based on the [documentation](https://www.gatsbyjs.com/docs/reference/gatsby-cli/#repl), I don't think that the REPL takes any options nor arguments. `gatsby repl` will open a REPL, where you can type the commands.

indie-ridge-dev: yes, you are right of course. I have just pushed a new version without the repl options and without the args for the sub commands.

dev/gatsby.ts
```diff
@@ -0,0 +1,253 @@
+export const completion: Fig.Spec = {
+  name: "gatsby",
+  description: "Gatsby CLI",
+  subcommands: [
+    {
+      name: "develop",
+      description: "Start the Gatsby development server",
+      args: {},
```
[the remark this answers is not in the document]
QuiiBz: Could you remove this argument, as Tim mentioned?

dev/gatsby.ts
```diff
@@ -0,0 +1,143 @@
...
+        {
+          name: ["-H", "--host"],
+          description: "Set host. Defaults to localhost",
+          args: {},
+        },
+        {
+          name: ["-p", "--port"],
+          description: "Set port. Defaults to env.PORT or 8000",
+          args: {},
+        },
```
QuiiBz, line 19: I would love to see `name` filled up in each argument

QuiiBz (changes requested):
Could you remove all `args: {}` where they are not needed, and add a `name` property to each one that really exists?

PR #272, opened 2021-06-08
Add aliases generator for git completion

chizmw:
This PR extends the existing `git` file to include an Argument generator which suggests git aliases (from config, not shell aliases).

Without this commands like `git st` become cumbersome with Fig as you have to <space> or <esc> away the menu to avoid the suggested vanilla command.

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -25,6 +25,17 @@ const gitGenerators: Record<string, Fig.Generator> = {
     },
   },
 
+  // user aliases
+  aliases: {
+    script: "git config --get-regexp '^alias' |cut -d. -f2",
+    postProcess: function (out) {
+      return out.split("\n").map((aliasLine) => {
+        const splitted = aliasLine.match(/^(\S+)\s(.*)/);
+        return { name: splitted[1], description: splitted[2] };
+      });
+    },
+  },
+
   // Saved stashes
   // TODO: maybe only print names of stashes
   stashes: {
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -25,6 +25,17 @@ const gitGenerators: Record<string, Fig.Generator> = {
     },
   },
 
+  // user aliases
+  aliases: {
+    script: "git config --get-regexp '^alias' |cut -d. -f2",
+    postProcess: function (out) {
+      return out.split("\n").map((aliasLine) => {
+        var splitted = aliasLine.match(/^(\S+)\s(.*)/);
+        return { name: splitted[1], description: splitted[2] };
```
QuiiBz, line 34: Could you use a const here?

QuiiBz (changes requested):
Thanks for your first contribution! Just a small change and we will be ready to merge.

chizmw:
I've made the change. For a cleaner git history would you like me to `rebase -i` this branch to one commit before having approve+merge?

QuiiBz:
> I've made the change. For a cleaner git history would you like me to `rebase -i` this branch to one commit before having approve+merge?

Yes, that would be cool!

chizmw:
Done! Thanks for your guidance.

QuiiBz (approved):
Thanks a lot for your first contribution! It's going to be really helpful

Issue #245, opened 2021-05-31
`compose` is not suggested with `docker` 20.10.6


deathemperor:
![image](https://user-images.githubusercontent.com/4255482/120172649-778d5500-c22d-11eb-82f8-12ff9db1fb48.png)

cstrnt:
Hey there @deathemperor :)

thank you very much for creating this issue :blush:!
Would you like to fix it as well? It should be a small fix and knowing docker itself makes it quite easy :)

Patil2099:
I would love to help in solving this issue. Can you guide me a little here @cstrnt?

cstrnt:
@Patil2099 of course :)
Probably the best resources to get started are the following:
- https://fig.io/docs/tutorials/contributing-to-public-specs
- https://fig.io/docs/concepts/cli-skeleton

Nutlope:
This seems like a good first issue and would love to submit a PR as a first time contributor. @Patil2099 just wanted to check if you're still interested in working on this since I don't want to steal it away from you if you are :)

Nutlope:
#273 addresses this issue

QuiiBz:
Merged in #273, thanks everyone!

PR #273, opened 2021-06-08
feat: add 'compose' to docker spec

Nutlope:
Feature
No 'compose' subcommand for 'docker' command
Add 'compose' subcommand to 'docker' command. No options/arguments were added due to a lack of familiarity with the most common options for this subcommand.
**Additional info:**
- [Docker compose docs](https://docs.docker.com/compose/cli-command/)
- Fulfills #245

### DIFF
--- a/dev/docker.ts
+++ b/dev/docker.ts
@@ -1959,6 +1959,11 @@ export const completionSpec: Fig.Spec = {
       // TODO Subcommands
     },
     {
+      name: "compose",
+      description: "Contains most of docker-compose features and flags",
+      // TODO Subcommands
+    },
+    {
       name: "container",
       description: "Manage containers",
     },
--- a/dev/docker.ts
+++ b/dev/docker.ts
@@ -1961,7 +1961,7 @@ export const completionSpec: Fig.Spec = {
     {
       name: "compose",
       description: "Contains most of docker-compose features and flags",
-      // TODO Subcommands
+      loadSpec: "docker-compose",
     },
     {
       name: "container",


QuiiBz:
Hi, thanks for your contribution. I saw that the spec for `docker-compose` already exists [here](https://github.com/withfig/autocomplete/blob/master/dev/docker-compose.ts). Let's not reinvent the wheel and use the `loadSpec` property instead. Here is a link to the documentation: https://fig.io/docs/reference/subcommand#loadspec

dev/docker.ts
```diff
@@ -1958,6 +1958,11 @@ export const completionSpec: Fig.Spec = {
       description: "Manage Docker configs",
       // TODO Subcommands
     },
+    {
+      name: "compose",
+      description: "Contains most of docker-compose features and flags",
+      // TODO Subcommands
+    },
```
QuiiBz, line 1965: Based on my previous comment, we can do this:
```suggestion
    {
      name: "compose",
      description: "Contains most of docker-compose features and flags",
      loadSpec: 'docker-compose',
    },
```

Nutlope:
> Hi, thanks for your contribution. I saw that the spec for `docker-compose` already exists [here](https://github.com/withfig/autocomplete/blob/master/dev/docker-compose.ts). Let's not reinvent the wheel and use the `loadSpec` property instead. Here is a link to the documentation: https://fig.io/docs/reference/subcommand#loadspec

I didn't know that, thanks for the heads up! Just committed your suggestion.

QuiiBz (approved):

PR #276, opened 2021-06-09
[DOCKER] Complete

### LINKED ISSUE #245: `compose` is not suggested with `docker` 20.10.6

DannyAziz:
Feature

Missing subcommands

Complete subcommands including all args and options

**Additional info:**
Also includes compose as does #273 
Fixes #245


dev/docker.ts
```diff
@@ -56,6 +57,113 @@ const dockerGenerators: Record<string, Fig.Generator> = {
       return true;
     },
   },
+  allDockerContexts: {
+    script: `docker context list --format '{{ json . }}'`,
+    postProcess: function (out) {
+      const allLines: Array<Record<string, string>> = out
+        .split("\n")
+        .map((line) => JSON.parse(line));
+      return allLines.map((i) => ({
```
QuiiBz, line 66: I would prefer having understandable variables names. Same for each generators underneath

dev/docker.ts
```diff
@@ -56,6 +57,113 @@ const dockerGenerators: Record<string, Fig.Generator> = {
...
+    script: `docker network list --format '{{ json . }}'`,
+    postProcess: function (out) {
+      const allLines: Array<Record<string, string>> = out
+        .split("\n")
+        .map((line) => JSON.parse(line));
+      return allLines.map((i) => ({
+        name: i.Name,
+        description: i.ID,
+      }));
+    },
```
QuiiBz, line 82: Looks like this code is duplicated few times - let's create a global function for this to avoid duplication

dev/docker.ts
```diff
@@ -76,1141 +184,1684 @@ const containerAndCommandArgs = [
...
+      "Inspect changes to files or directories on a container's filesystem",
+    args: containersArg,
+  },
+  exec: {
+    name: "exec",
+    description: "Run a command in a running container",
+    options: [
+      {
+        name: ["-it"],
+        insertValue: "-it ",
```
QuiiBz, line 1112: Not sure why there is an `insertValue` here, looks like it's not needed

dev/docker.ts
```diff
@@ -76,1141 +184,1684 @@ const containerAndCommandArgs = [
...
+      name: "containers",
+      suggestions: [
         {
-          args: {
-            name: "string",
-          },
-          description: 'Signal to stop a container (default "SIGTERM")',
-          name: ["--stop-signal"],
+          name: "$(docker ps -aq)",
+          insertValue: "$(docker ps -aq)",
```
QuiiBz, line 1290: No need to set an `insertValue` when it's the same as the `name`

dev/docker.ts
```diff
@@ -76,1141 +184,1684 @@ const containerAndCommandArgs = [
...
+          description: "All containers, running and exited",
         },
         {
-          args: {
-            name: "int",
-          },
-          description: "Timeout (in seconds) to stop a container",
-          name: ["--stop-timeout"],
+          name: "$(docker ps -q)",
+          insertValue: "$(docker ps -q)",
```
QuiiBz, line 1295: Same as above

dev/docker.ts
```diff
@@ -1425,584 +2014,3111 @@ export const completionSpec: Fig.Spec = {
...
+            },
+          ],
+          options: [
+            {
+              name: "--default-stack-orchestrator",
+              description:
+                "Default orchestrator for stack operations to use with this context (swarm|kubernetes|all)",
+              args: {
+                name: "string",
+              },
```
QuiiBz, line 2345: I would love to see suggestions here! (swarm|kubernetes|all)

dev/docker.ts
```diff
@@ -1425,584 +2014,3111 @@ export const completionSpec: Fig.Spec = {
...
-          },
+          name: "list",
+          description: "List available contexts",
+          options: [
+            {
+              name: "--format",
+              description:
+                "Format the output. Values: [pretty | json]. (Default: pretty)",
+              args: {
+                name: "string",
```
QuiiBz, line 2448: Same as above (pretty|json)

dev/docker.ts
```diff
@@ -1425,584 +2014,3111 @@ export const completionSpec: Fig.Spec = {
...
+          name: "update",
+          description: "Update a context",
+          args: contextsArg,
+          options: [
+            {
+              name: "--default-stack-orchestrator",
+              description:
+                "Default orchestrator for stack operations to use with this context (swarm|kubernetes|all)",
+              args: {
+                name: "string",
```
QuiiBz, line 2496: Same as above

dev/docker.ts
```diff
@@ -1425,584 +2014,3111 @@ export const completionSpec: Fig.Spec = {
...
+                name: "uint",
+              },
+            },
+            {
+              name: "--mode",
+              description:
+                'Service mode (replicated, global, replicated-job, or global-job) (default "replicated")',
+              args: {
+                name: "string",
+              },
```
QuiiBz, line 3550: Same as above

dev/docker/compose-build.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const completionSpec: Fig.Spec = {
+  name: "build",
+  description: "Build or rebuild services.",
+  args: [getServices],
+  options: [
+    {
+      name: ["--build-arg"],
+      description: "Set build-time variables for services.",
+      args: {},
```
QuiiBz, line 20: Could you add at least a name for this arg? Same for other args underneath

dev/docker/compose-build.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    {
+      name: ["--compress"],
+      description: "Compress the build context using gzip.",
+      args: {},
```
QuiiBz, line 25: No need to set args when the option doesn't take any - same of underneath

QuiiBz (changes requested):
Just a quick note: there is no need to set `args` when the option doesn't take any. I didn't comment this on each file but looks like this issue repeats in most of them inside `docker/`

dev/docker.ts
DannyAziz: Not part of anything I did, I just moved it around - But you're right, will remove it

dev/docker.ts
DannyAziz: Save as above

dev/docker.ts
DannyAziz: Save as above (Not part of anything I did, I just moved it around - But you're right, will remove it)

dev/docker/compose-build.ts
DannyAziz: Everything inside `docker/compose-*.ts` is just what was in `docker-compose.ts` before but extracted out

Anything missing was already missing in the spec, but will add everything in

DannyAziz:
@QuiiBz Actioned all of your feedback & fixed merge conflicts :)

QuiiBz (approved):
Perfect! Thanks a lot

Issue #247, opened 2021-06-01
Add warning when running `npm run dev` if Fig is not installed


mattschrage:
We should provide a warning message when attempting to develope Fig specs when the macOS application is not installed.

PR #280, opened 2021-06-10
Add "Fig Not Installed" Warning in Dev Script

Monkeyank:
Dev Script Update
#247 Dev Script doesn't include a warning about Fig not being installed
Adds warning!
**Additional info:**
In order to test  what happens when the app isn't installed, change line 24 from `whichapp Fig` to `whichapp blah`.

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -1,6 +1,28 @@
 #!/bin/bash
 
 clear
+
+whichapp() {
+    local appNameOrBundleId=$1 isAppName=0 bundleId
+    # Determine whether an app *name* or *bundle ID* was specified.
+    [[ $appNameOrBundleId =~ \.[aA][pP][pP]$ || $appNameOrBundleId =~ ^[^.]+$ ]] && isAppName=1
+    if (( isAppName )); then
+        # Translate to a bundle ID first.
+        bundleId=$(osascript -e "id of application \"$appNameOrBundleId\"" 2>/dev/null) || 
+        { 
+            echo
+            echo "******"
+            echo "******"
+            echo "WARNING: Fig is not installed" 1>&2;
+            echo "******"
+            echo "******"
+            echo
+        }
+    fi
+}
+
+whichapp Fig
+
 echo 
 echo "Welcome to $(tput bold)$(tput setaf 5)Fig Dev Mode$(tput sgr0)";
 echo 


falky97 (changes requested):
Haven't had a chance to test this, however, a few of things:

1. The Fig CLI doesn't exist error usually occurs when the user is on Linux. Can also occur when on macOS and Fig isn't installed
2. `osascript` doesn't exist on Linux so we shouldn't use it
3. We would like for Linux users to be able to contribute to specs. So I think the best fix here is to 1. Implement the checks below and 2. have a function that determined if Fig exists and updates an env variable if it does. Then every time a `fig` command is used, we should check if the CLI exists. Or alternatively we should have error messages just go to /dev/null




I think using the logic on line 11 of this file would work best: https://github.com/withfig/config/blob/main/fig.sh#L11 - (Check if Fig app is installed and check if Fig CLI exists)


Finally, this code is good and gets the job done, but it's not super readable. I think we'd prefer more basic code (like above) that's easy to understand. e.g. below you're using command substitution, arithmetic substitution, wildcards/globs... This works, but I'm sure someone less experienced with bash will be too scared to touch this and we don't want that. If we have to do something complex, at least add comments! But here, I don't think we need to do anything complex

scripts/dev.sh
```diff
@@ -1,6 +1,38 @@
 #!/bin/bash
 
 clear
+
+# Checks if CLI is installed
+if ! command -v Fig --help &> /dev/null; then
```
mattschrage, line 6: Any reason `Fig` is capitalized?

mattschrage (changes requested):
I think we should also check which operating system the user is currently running. This might be an easier check than seeing if the Applications folder exists. Can you add an early exit if not on macOS?

QuiiBz:
> I think we should also check which operating system the user is currently running.

This won't work by default in Windows - the user will need to install either [Cygwin](https://www.cygwin.com/) or [Git Bash](https://git-scm.com/download/win) (or any other tool to read bash).

Maybe we should rewrite this script in pure JS (using [ZX](https://github.com/google/zx) can be a nice idea) so it will work natively on all operating systems.

What's your thought on this @cstrnt?

mattschrage:
Here is a snippet from [StackOverflow](https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script) that let's you check which operating system is currently in use:
```bash
if [[ "$OSTYPE" == "linux-gnu"* ]]; then
        # ...
elif [[ "$OSTYPE" == "darwin"* ]]; then
        # Mac OSX
elif [[ "$OSTYPE" == "cygwin" ]]; then
        # POSIX compatibility layer and Linux environment emulation for Windows
elif [[ "$OSTYPE" == "msys" ]]; then
        # Lightweight shell and GNU utilities compiled for Windows (part of MinGW)
elif [[ "$OSTYPE" == "win32" ]]; then
        # I'm not sure this can happen.
elif [[ "$OSTYPE" == "freebsd"* ]]; then
        # ...
else
        # Unknown.
fi
```

scripts/dev.sh
```diff
@@ -1,6 +1,38 @@
 #!/bin/bash
 
 clear
+
+# Checks if CLI is installed
+if ! command -v Fig --help &> /dev/null; then          
+
+    # Checks if Applications directory exists
+    if ([[! -d /Applications ]] && [[! -d ~/Applications]])> /dev/null 2> /dev/null; then
```
brendanfalk, line 9: 1. You need to use `||` not `&&`
2. No need to check if applications folder exists else if fig app exists. Just check if fig app is installed

```
if ([[ -d /Applications/Fig.app ]] || [[ -d ~/Applications/Fig.app ]]) 
```

Taken from here: https://github.com/withfig/config/blob/main/fig.sh

scripts/dev.sh
brendanfalk: 1. yes agree with Matt `fig` should be lowercase 
2. Why is `--help` there? 

Use this code instead:
```
command -v fig 1> /dev/null 2> /dev/null
```

https://github.com/withfig/config/blob/main/fig.sh

brendanfalk (changes requested):
So a few things
This is a good first pass, however, has a few small coding errors + implementation isn't exactly what we want

### What do we want

I'm on a mac
* If you don't have fig app or fig cli - prompt user to download Fig using special invite link. I will DM this to you @Monkeyank 

I'm not on a mac
* You're fucked. Tell the user we are macOS only for now. Working on linux / windows build soon. You can build and contribute to completion specs but you will not be able to test them unless you are on a mac or a macOS VM!

Extra
* Any time the `fig` cli is used in **any** of the scripts across this repo, make sure we fail gracefully (ie send error (`2>`) to `/dev/null`)

QuiiBz:
This looks fine to me, but again I'm not very familiar with `bash`. Let's wait for Brendan and Matt's reviews.

mschrage (approved):

Nutlope:
@Monkeyank Can you resolve this merge conflict? @brendanfalk are we good to merge this after this merge conflict is resolved? Looks like Matt approved it 2 weeks ago.

QuiiBz:
Bumping this @Monkeyank @brendanfalk

scripts/dev.sh
```diff
@@ -1,6 +1,47 @@
...
+    # Check if Fig app is installed
+    if !([[ -d /Applications/Fig.app ]] || [[ -d ~/Applications/Fig.app ]]) 1> /dev/null 2> /dev/null; then
+        echo
+        echo
+        echo "$(tput setaf 7)******";
+        echo "******";
+        echo "$(tput setaf 3)$(tput bold) WARNING: Fig App is not installed";
+        echo
+        echo "$(tput setaf 6) Get access to download fig by joining the fig community at:"; 
+        echo "https://fig.io/community"
```
fedeci, line 17:
```suggestion
        echo " https://fig.io/community"
```

fedeci (commented):
Could you rebase please?

PR #283, opened 2021-06-11
Add auto generator of completion spec files

Alonely0:
Now by using the command `CMD=YOUR_COMMAND npm run auto-fig-command` you can create a completion spec file of that command based on its man page. The use of this script it's not intended to replace the manual created spec files, the objective of this script is to help in their creation and create rudimentary command spec files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "dev": "\"$npm_execpath\" run build && source scripts/dev.sh",
     "test": "tsc --noEmit && echo 'All specs passed validation. You are ready to push!'",
+    "auto-fig-command": "ts-node ./scripts/auto-fig-command.ts $CMD",
     "copy": "func() { mkdir -p ~/.fig/autocomplete/$(dirname ${1}.js); cp \"$(pwd)\"/specs/${1}.js ~/.fig/autocomplete/${1}.js; }; func",
     "copy:all": "rsync -a --include '*/' --include '*.js' --exclude '*' $(pwd)/specs/ ~/.fig/autocomplete/",
     "create-boilerplate": "./scripts/create-boilerplate.sh",
@@ -62,7 +63,7 @@
   "license": "ISC",
   "devDependencies": {
     "@types/inquirer": "^7.3.1",
-    "@types/node": "^14.14.33",
+    "@types/node": "^14.17.3",
     "@types/progress": "^2.0.3",
     "@typescript-eslint/eslint-plugin": "^4.18.0",
     "@typescript-eslint/parser": "^4.18.0",
@@ -80,5 +81,13 @@
     "progress": "^2.0.3",
     "ts-node": "^9.1.1",
     "typescript": "^4.2.3"
+  },
+  "dependencies": {
+    "exec-sh": "^0.4.0",
+    "fs": "^0.0.1-security",
+    "manolo": "^0.1.0",
+    "node-html-parser": "^3.3.5",
+    "pretty": "^2.0.0",
+    "remove-trailing-spaces": "^1.0.8"
   }
 }
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "manolo": "^0.1.0",
     "node-html-parser": "^3.3.5",
     "pretty": "^2.0.0",
-    "remove-trailing-spaces": "^1.0.8"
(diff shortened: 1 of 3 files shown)


QuiiBz:
Really cool! As mentioned on Discord, I think those sorts of tooling should be in a separate repo. @falky97 @mattschrage what's your thought on this?

fedeci:
`@fig/cli` sounds good to me😄

package.json
```diff
@@ -80,5 +81,14 @@
     "progress": "^2.0.3",
     "ts-node": "^9.1.1",
     "typescript": "^4.2.3"
+  },
+  "dependencies": {
+    "exec-sh": "^0.4.0",
+    "fs": "^0.0.1-security",
```
cstrnt, line 87: !!!! fs isn't a depedency and comes with node! Please be careful when installing packages mindlessly

scripts/auto-fig-command.ts
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+// Dependencies
+const fs = require("fs");
+const { sleep } = require("sleep");
+const execSh = require("exec-sh");
+const { parse } = require("node-html-parser");
+const pretty = require("pretty");
+const manolo = require("manolo");
+const removeTrailingSpaces = require("remove-trailing-spaces");
```
cstrnt, line 15: Any reason why you are using `require` over `import`?

scripts/auto-fig-command.ts
```diff
@@ -0,0 +1,159 @@
...
+    // prepare file
+    const file = `export const completionSpec: Fig.Spec = 
+${JSON.stringify(figCompletionSpec)}
+`;
+    // write file
+    fs.writeFileSync(
+      `./dev/${process.argv[2]}.ts`,
+      file,
+      "utf8",
+      (err: never) => {
```
cstrnt, line 152: I don't think errors never happen! I think the type should be infered here

scripts/auto-fig-command.ts
```diff
@@ -0,0 +1,159 @@
...
+        if (err) {
+          console.error("Failed to create file:\n", err);
+        }
+      }
+    );
+  }
+);
```
cstrnt, line 159: I would prefer creating an async function, maybe even called run to avoid .then callbacks

scripts/auto-fig-command.ts
```diff
@@ -0,0 +1,159 @@
...
+      })
+      // catch errors
+      .catch((_) => {
+        _ = "";
+      });
+    continue;
+  }
+  // init base completion spec
+  const figCompletionSpec = {
+    name: process.argv[2],
```
cstrnt, line 117: maybe create a top level variable called specName because this is used multiple times along this file

scripts/auto-fig-command.ts
```diff
@@ -0,0 +1,159 @@
...
+  // Check command was provided
+  if (cmd == undefined) {
+    console.log("Usage: `CMD=YOUR_COMMAND npm run auto-fig-command`");
+    process.exit(1);
+  }
+
+  // Wait until we get the parsed man page
+  const manPage = await new Promise((resolve) => {
+    // man -w MAN_PAGE gives us the path to it
+    execSh(
```
cstrnt, line 29: this package actually offers a promise-based option which would reduce the callback hell here as well 😆

cstrnt (changes requested):

scripts/auto-fig-command.ts
Alonely0: because some packages can't be imported and have to be required (the linter throws me an error)

package.json
Alonely0: Ok, thanks!

scripts/auto-fig-command.ts
Alonely0: Could happen a write error because of lack of permissions for example.

scripts/auto-fig-command.ts
Alonely0: Ok!

scripts/auto-fig-command.ts
Alonely0: No, I'm just more used to require, but I'm going to change them.

Alonely0:
https://github.com/withfig/man-page-scraper/pull/1

PR #288, opened 2021-06-12
feat: add `next` subcommand to `npx`

fedeci:
This PR adds the `next` also as a subcommand of `npx`. I don't think it is a common practice to have `next`, `babel`, `tailwind`, etc...installed globally. 
Instead many `npm` CLI tools are convenient to be run as `npx --no-install [package-name]` from the project dir.

I think the same applies to other `npm` commands.

### DIFF
--- a/dev/npx.ts
+++ b/dev/npx.ts
@@ -7,6 +7,11 @@ export const completionSpec: Fig.Spec = {
       icon: "https://reactnative.dev/img/pwa/manifest-icon-512.png",
       loadSpec: "react-native",
     },
+    {
+      name: "next",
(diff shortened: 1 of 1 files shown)


QuiiBz (approved):
Nice addition! I think we will need to have the same specs for `npx <cli>` and `yarn <cli>`, but we can do this in another PR.

fedeci:
Yes, it would also be great not having to specify description and icon twice for loaded specs.

QuiiBz:
> Yes, it would also be great not having to specify description and icon twice for loaded specs.

Totally agree, I just opened an issue for this. For now, let's merge this and we will update it when https://github.com/withfig/fig/issues/164 will be implemented!

Issue #220, opened 2021-05-20
[DOTSLASH] Prioritize executable files


Tyriar:
**Is your feature request related to a problem? Please describe.**

Normally in bash the .bat file would be ignored when tabbing this because it's not executable:

![](https://user-images.githubusercontent.com/2193314/118982937-c2a39000-b930-11eb-8e48-c45f916a5d1d.png)

**Describe the solution you'd like**

I'd like the file to still show up but as a deprioritized option:

![image](https://user-images.githubusercontent.com/2193314/118983227-0e563980-b931-11eb-81e1-c3134ed22449.png)

QuiiBz:
Merged in #290, thank you all!

PR #290, opened 2021-06-12
[DOTSLASH] Prioritize executable files

Monkeyank:
Prioritizes executable files
#220 No priority given to executable files.
Prioritizes executable files
**Additional info:**

### DIFF
--- a/dev/dotslash.ts
+++ b/dev/dotslash.ts
@@ -29,7 +29,7 @@ const dotslashGenerator: Fig.Generator = {
       }
 
       // Removes character at the end of the filepath
-      if (path.endsWith("*" || "@" || "%" || "|")) {
+      if (["@", "*", "%", "|"].includes(path.slice(-1))) {
         filename = filename.slice(0, -1);
       }
 
--- a/dev/dotslash.ts
+++ b/dev/dotslash.ts
@@ -1,10 +1,55 @@
-// use to complete on files when they are the first string in a branch
+const dotslashGenerator: Fig.Generator = {
+  script: function (context) {
+    let path = context[context.length - 1];
+
+    if (path.includes("/")) {
+      const components = path.split("/");
+      components.pop();
+      path = components.join("/");
+    } else {
+      path = ".";
+    }
+
+    // escape spaces in path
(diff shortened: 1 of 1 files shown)


QuiiBz:
Looks like a `@` should be removed at the end of some files, as you can see:
Screenshot 2021-06-12 at 13 24 20

Monkeyank:
> Looks like a `@` should be removed at the end of some files, as you can see:
> Screenshot 2021-06-12 at 13 24 20

I removed the other characters that are present from `ls -F`, like `"@", "%", "|"`

QuiiBz:
I'm not sure if that would not be better to directly make this feature in the Fig app, because it can be used by other specs. @cstrnt @falky97 any thought on this?

dev/dotslash.ts
```diff
@@ -1,10 +1,55 @@
...
+      const isFolder = path.endsWith("/");
+
+      if (isFolder) {
+        filename = components[components.length - 2] + "/";
+      }
+
+      // Removes character at the end of the filepath
+      if (path.endsWith("*" || "@" || "%" || "|")) {
+        filename = filename.slice(0, -1);
+      }
```
QuiiBz, line 34: Still having the `@` at the end - the others `OR` will never be called in this code. You can either duplicate `path.endsWith("<symbol>")` for each symbol, or grab the last letter and check if an array of all the symbols contains it. That would be something like:
```js
if (["*", "*", "%", "|"].includes(lastLetter)) {
...
```

Screenshot of the `@`:
Screenshot 2021-06-24 at 09 05 57

mattschrage:
> I'm not sure if that would not be better to directly make this feature in the Fig app, because it can be used by other specs. @cstrnt @falky97 any thought on this?

We could potentially include this as a new template called `executables` but I think for the time being, it is fine for this to be defined here.

mattschrage (approved):
I'm excited for this one to land in production! Great work @Monkeyank 🙌

QuiiBz (approved):
This is going to be really helpful, thanks a lot @Monkeyank!

Issue #274, opened 2021-06-08
[yarn] Yarn workspace support


davidfant:
### Completion Spec
yarn workspace

### Description
There is no autocomplete for `yarn workspace`. It's expected to run scripts within specific workspaces

`https://yarnpkg.com/cli/workspace`

### Steps to reproduce
package.json

```json
{
  "workspaces": ["subdirectory"]
}
```

subdirectory/package.json

```json
{
  "name": "subdirectory-project"
  "scripts": {
    "hello": "echo 'world'"
  }
}
```

### Expected result
Typing `yarn workspace subdirectory-project` should autocomplete `yarn workspace subdirectory-project hello`

### Actual result
No autocomplete

Nutlope:
Working on this! Will submit a PR soon

Nutlope:
#361 has been merged

PR #294, opened 2021-06-13
feat: enhance yarn workspace

Nutlope:
feature

no autocomplete for yarn workspace

add autocomplete for yarn workspace

**Additional info:**

* Recorded a [loom video](https://www.loom.com/share/0ca49e6ec5fc4e0f93bdf16b1c23ef1f) to show demo working
* Fulfills #274

### DIFF
--- a/dev/yarn.ts
+++ b/dev/yarn.ts
@@ -1019,6 +1019,31 @@ export const completionSpec: Fig.Spec = {
             },
           },
         },
+        {
+          name: "script",
+          generators: {
+            script: function (context) {
+              return `cat ${context[2]}/package.json`;
+            },
+            postProcess: function (out) {
+              if (out.trim() == "") {
+                return [];
+              }
+              try {
+                const packageContent = JSON.parse(out);
+                const scripts = packageContent["scripts"];
+                if (scripts) {
+                  const scriptKeys = [];
+                  for (let i = 0; i < Object.keys(scripts).length; i++) {
+                    scriptKeys.push({ name: Object.keys(scripts)[i] });
+                  }
+                  return scriptKeys;
+                }
+              } catch (e) {}
+              return [];
+            },
+          },
+        },
         // TODO arg 1 is script suggestion from the workspace specified in arg 0.
       ],
     },
--- a/dev/yarn.ts
+++ b/dev/yarn.ts
@@ -1023,7 +1023,7 @@ export const completionSpec: Fig.Spec = {
           name: "script",
           generators: {
             script: function (context) {
-              return `cat ${context[2]}/package.json`;
+              return `cat ${context[context.length - 2]}/package.json`;
             },
             postProcess: function (out) {
               if (out.trim() == "") {


QuiiBz:
Very cool! Just tried out and found an edge case:

If you set the workspaces to something like `packages/*` (which is very common), it will autocomplete `packages/*` instead of the actual packages inside `packages/`:

Screenshot 2021-06-13 at 09 27 58

This is caused by the previous generator here https://github.com/withfig/autocomplete/blob/d92f28fefe9f4b08d93499e373ced23eb2ec3364/dev/yarn.ts#L1005 and so break this new generator too.

It would be really helpful if you could fix this issue along with this neat feature.

Nutlope:
> Very cool! Just tried out and found an edge case:
> 
> If you set the workspaces to something like `packages/*` (which is very common), it will autocomplete `packages/*` instead of the actual packages inside `packages/`:
> 
> Screenshot 2021-06-13 at 09 27 58
> 
> This is caused by the previous generator here
> 
> https://github.com/withfig/autocomplete/blob/d92f28fefe9f4b08d93499e373ced23eb2ec3364/dev/yarn.ts#L1005
> 
> and so break this new generator too.
> It would be really helpful if you could fix this issue along with this neat feature.

Thanks! I have no real experience with yarn workspaces so it's interesting to learn about edge cases like this. If I'm understanding this correctly, we want to autocomplete `packages` when workspaces is set to something like `packages/*`? 

If so, would a simple if statement be appropriate in the first generator, to loop through all the workspaces and if any of them are set to `SOMETHING/*`, replace the suggestion with `SOMETHING`?

QuiiBz:
The `packages/*` notation means « all folders inside packages/ are part of the workspace ». In your example (the loom video), you could put workspace-1 and workspace-2 inside a `packages` folder, and set `workspaces` in the main `package.json` to `packages/*`, and it will do the same.

Did you understand my example and see a bit more how workspaces works?

Nutlope:
Ahh I see, yes that makes sense. So for the example you just mentioned, if the package.json workspaces were set to `packages/*`, we'd want to autocomplete `yarn workspace` with a list of folders inside the packages folder?
 
So `workspace-1` and `workspace-2` (assuming these are the two folders inside the `packages` folder)?

QuiiBz:
> Ahh I see, yes that makes sense. So for the example you just mentioned, if the package.json workspaces were set to `packages/*`, we'd want to autocomplete `yarn workspace` with a list of folders inside the packages folder?
> 
> So `workspace-1` and `workspace-2` (assuming these are the two folders inside the `packages` folder)?

Exactly! As you can see it's going to be a bit more challenging than the current implementation. Do not hesitate if you need help.

dev/yarn.ts
```diff
@@ -1019,6 +1019,31 @@ export const completionSpec: Fig.Spec = {
             },
           },
         },
+        {
+          name: "script",
+          generators: {
+            script: function (context) {
+              return `cat ${context[2]}/package.json`;
```
brendanfalk, line 1026: Rather than `context[2]` I would use `context[context.length-2]`. 

Why? If a user input `yarn -a workspace abc[space]` then `context[2]` would yield `workspace` not `abc`. 

In general, when using script as a function, always work your way backwards in the context array not forwards

brendanfalk (changes requested):

dev/yarn.ts
Nutlope: Makes perfect sense, thanks for the explanation! Added a commit for this

QuiiBz:
Hello @Nutlope, do you want to continue working on this PR?

Nutlope:
Hey @QuiiBz, completely forgot about this, thanks for the reminder! Will work on this later this week.

QuiiBz:
Alright! Feel free to ask if you need any help.

Nutlope:
Created a cleaner PR #361 , closing this one

Issue #295, opened 2021-06-14
[NODE & PYTHON] Prioritize files over folders


Monkeyank:
### Completion Spec
Python and Node

### Description
Give higher suggestion priority to files over folders.

PR #296, opened 2021-06-14
[Node & Python] Prioritize file suggestions over folder suggestions

Monkeyank:
Enhancement to suggestion priority
#295 
Priority is given to files (76) over folders.
**Additional info:**

### DIFF
--- a/dev/node.ts
+++ b/dev/node.ts
@@ -15,7 +15,8 @@ export const completionSpec: Fig.Spec = {
             return false;
           })
           .map((file) => {
-            const isJsFile = typeof file.name === "string" && file.name.endsWith(".js")
+            const isJsFile =
+              typeof file.name === "string" && file.name.endsWith(".js");
 
             return {
               ...file,
--- a/dev/node.ts
+++ b/dev/node.ts
@@ -15,10 +15,8 @@ export const completionSpec: Fig.Spec = {
             return false;
           })
           .map((file) => {
-            let isJsFile = false;
-            if (typeof file.name === "string" && file.name.endsWith(".js")) {
-              isJsFile = true;
-            }
+            const isJsFile = typeof file.name === "string" && file.name.endsWith(".js")
+
             return {
               ...file,
               priority: isJsFile && 76,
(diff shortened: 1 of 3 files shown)


dev/node.ts
```diff
@@ -7,12 +7,23 @@ export const completionSpec: Fig.Spec = {
...
+              return file.name.endsWith(".js") || file.name.endsWith("/");
+            }
+            return false;
+          })
+          .map((file) => {
+            let isJsFile = false;
+            if (typeof file.name === "string" && file.name.endsWith(".js")) {
+              isJsFile = true;
+            }
+            return {
```
QuiiBz, line 22: Can be simplified to:
```suggestion
            const isJsFile = typeof file.name === "string" && file.name.endsWith(".js")

            return {
```

dev/python.ts
```diff
@@ -7,12 +7,23 @@ export const completionSpec: Fig.Spec = {
...
+              return file.name.endsWith(".py") || file.name.endsWith("/");
+            }
+            return false;
+          })
+          .map((file) => {
+            let isPyFile = false;
+            if (typeof file.name === "string" && file.name.endsWith(".py")) {
+              isPyFile = true;
+            }
+            return {
```
QuiiBz, line 22: Same:
```suggestion
            const isPyFile = typeof file.name === "string" && file.name.endsWith(".js")

            return {
```

dev/python3.ts
```diff
@@ -7,12 +7,23 @@ export const completionSpec: Fig.Spec = {
...
+              return file.name.endsWith(".py") || file.name.endsWith("/");
+            }
+            return false;
+          })
+          .map((file) => {
+            let isPyFile = false;
+            if (typeof file.name === "string" && file.name.endsWith(".py")) {
+              isPyFile = true;
+            }
+            return {
```
QuiiBz, line 22: Same:
```suggestion
            const isPyFile = typeof file.name === "string" && file.name.endsWith(".js")

            return {
```

QuiiBz (changes requested):
Cool improvement! I left some review to simplify the code

QuiiBz (approved):
LGTM

PR #300, opened 2021-06-15
refactor: Use generateSpec to generate fig settings subcommand

QuiiBz:
Refactor

Hard-coded `fig settings [subcommand] [arg]`

Use `generateSpec` and the settings schema at `~/.fig/tools/all-settings.json` to dynamically generate the subcommand and args.

**Additional info:**


dev/fig.ts
```diff
@@ -10,337 +20,44 @@ export const completionSpec: Fig.Spec = {
...
+              default: defaultValue,
+            }) => {
+              const suggestions =
+                type === "boolean"
+                  ? ["true", "false"]
+                  : options?.map((option) => ({
+                      name: option["name"] || option,
+                      description: option["description"] || "",
+                    }));
+              const variadic = type === "multi_select";
```
mattschrage, line 45: When a setting is `multiselect` it must be written as a JSON array. See the existing completion spec, specifically `autocomplete.disableForCommands` for the current approach. 

We could also default to just inserting `'["{cursor}"]'`.

mattschrage (commented):
One small fix!

dev/fig.ts
QuiiBz: Added the `insertValue` and generators `devCompletionsFolder` & `disableForCommands`!

dev/fig.ts
```diff
@@ -10,337 +71,47 @@ export const completionSpec: Fig.Spec = {
...
+                  : options?.map((option) => ({
+                      name: option["name"] || option,
+                      description: option["description"] || "",
+                    }));
+              const insertValue =
+                type === "multiselect" ? `${name} '{cursor}'` : undefined;
+              const generators = SETTINGS_GENERATOR[name];
+
+              return {
+                name,
```
mattschrage, line 101: Can we add the same icon we had before? eg. `fig://icon?type=commandkey`.

mattschrage (approved):
One small change and then it looks good to me! (As long as you've tested it with the current `all-settings.json` file and everything works?)

dev/fig.ts
QuiiBz: Done!

dev/fig.ts
```diff
@@ -1,3 +1,64 @@
...
+
+interface Settings {
+  settingName: string;
+  description: string;
+  type: "boolean" | "text" | "single_select" | "multiselect";
+  options?: Array<string | { name: string; description: string }>;
+  default?: string;
+}
+
+const devCompletionsFolderGenerator: Fig.Generator = {
```
cstrnt, line 11: Is there any reason why we don't use the folder template here?

QuiiBz: The path should be absolute and not relative, so we use a custom generator for this:
Screenshot 2021-06-23 at 17 44 36

Issue #307, opened 2021-06-17
[Stripe] CLI Request


Monkeyank:
### CLI Tool
Stripe

### Useful Links
[Stripe Docs](https://stripe.com/docs/cli)

PR #308, opened 2021-06-17
Stripe

Monkeyank:
Add Stripe Completion Spec
#307 

**Additional info:**


dev/stripe.ts
```diff
@@ -0,0 +1,759 @@
...
+      name: "setting",
+      suggestions: ["on", "off", "auto"],
+    },
+  },
+  {
+    name: "--config",
+    description: "Sets your config file",
+    args: {
+      name: "config filepath",
+    },
```
QuiiBz, line 22: I think this can have a template of `filepaths`

dev/stripe.ts
```diff
@@ -0,0 +1,759 @@
...
+const sharedOptions: Fig.Option[] = [
+  {
+    name: ["-s", "--show-headers"],
+    description: "Shows response HTTP headers",
+  },
+  {
+    name: ["-c", "--confirm"],
+    description:
+      "Skips the warning prompt and automatically confirms the command being entered",
+  },
```
QuiiBz, line 59: We can add `isDangerous` here

dev/stripe.ts
```diff
@@ -0,0 +1,759 @@
...
+            description: "Key1:Value1, Key2:Value2",
+          },
+        },
+        {
+          name: ["-e", "--events"],
+          description: "A comma-seperated list of which events to listen for",
+          args: {
+            variadic: true,
+            name: "events types",
+          },
```
QuiiBz, line 449: Could this arg provide suggestions for the list of events? Looks like they are defined here: https://github.com/withfig/autocomplete/pull/308/files#diff-e38baeee0f96b87459e1901e70e24b9f815f6a3a4ddd9b7e27f811f7dadf4408R508

dev/stripe.ts
```diff
@@ -0,0 +1,759 @@
...
+            "Listen for all webhook events based on existing webhook endpoints",
+        },
+        {
+          name: ["-j", "--print-json"],
+          description: "Print JSON objects to stdout",
+        },
+        {
+          name: "--skip-verify",
+          description:
+            "Skip certificate verification when forwarding to HTTPS endpoints",
```
QuiiBz, line 496: We can also add `isDangerous` here

QuiiBz (changes requested):
Awesome CLI! I left some comments

QuiiBz:
Are those suggestions the same?

https://github.com/withfig/autocomplete/pull/308/files#diff-e38baeee0f96b87459e1901e70e24b9f815f6a3a4ddd9b7e27f811f7dadf4408R451-R1235

https://github.com/withfig/autocomplete/pull/308/files#diff-e38baeee0f96b87459e1901e70e24b9f815f6a3a4ddd9b7e27f811f7dadf4408R1296-R1342

Monkeyank:
> Are those suggestions the same?
> 
> https://github.com/withfig/autocomplete/pull/308/files#diff-e38baeee0f96b87459e1901e70e24b9f815f6a3a4ddd9b7e27f811f7dadf4408R451-R1235
> 
> https://github.com/withfig/autocomplete/pull/308/files#diff-e38baeee0f96b87459e1901e70e24b9f815f6a3a4ddd9b7e27f811f7dadf4408R1296-R1342

They're not the same, because one of them is just all/general events, while the other is specific to webhook events.

QuiiBz:
> They're not the same, because one of them is providing suggestions for all/general events, while the other is providing suggestions specific to webhook events.

Alright, then it would be better not to duplicate the webhooks events and put them in a top-level const Fig.Suggestion[]. Then, you can use this const for the webhooks events suggestions and spread it for the general events suggestions.

QuiiBz (approved):
LGTM!

PR #313, opened 2021-06-19
Remove preceding '+' for branches checked out in another worktree

sullivan-sean:
Bug fix for git cli. Current behavior when branch `other` is checked out in another worktree is to show `+ other` as a suggestion.

New behavior handles this case, suggesting `other` instead, while still handling branch names with the `+` character in them.

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -95,17 +95,21 @@ const gitGenerators: Record<string, Fig.Generator> = {
 
       return output.split("\n").map((elm) => {
         // current branch
-        if (elm.includes("*")) {
+        let name = elm.trim();
+        const parts = elm.match(/\S+/g);
+        if (parts.length > 1 && parts[0] == "*") {
           return {
             name: elm.replace("*", "").trim(),
             description: "current branch",
             icon: "⭐️",
             // priority: 100,
           };
+        } else if (parts.length > 1 && parts[0] == "+") {
+          name = elm.replace("+", "").trim();
         }
 
         return {
-          name: elm.trim(),
+          name,
           description: "branch",
           icon: "fig://icon?type=git",
         };
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -94,18 +94,18 @@ const gitGenerators: Record<string, Fig.Generator> = {
       }
 
       return output.split("\n").map((elm) => {
-        // current branch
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -95,17 +95,21 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+        if (parts.length > 1 && parts[0] == "*") {
           return {
             name: elm.replace("*", "").trim(),
             description: "current branch",
             icon: "⭐️",
             // priority: 100,
           };
+        } else if (parts.length > 1 && parts[0] == "+") {
+          name = elm.replace("+", "").trim();
         }
```
QuiiBz, line 109: The `parts.length > 1` can be moved if an upper `if` to avoid duplication. Also, shouldn't this be a `>=` instead of `>`?

sullivan-sean: I think `>` is preferable here. There will be at most two parts, with the first being a flag like `*` or `+` and the second being a branch name. We only want to remove the `*` or `+` if there are two parts. If we use `>=` and there is a branch named `+` that is not the current branch and is not in a worktree, then we would hit this case and trim the name of the branch to an empty string (weird case I know but it is possible to have a branch named `+`)

QuiiBz: I see, thanks for the clarification!

dev/git.ts
```diff
@@ -95,17 +95,23 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
-          };
+        let name = elm.trim();
+        const parts = elm.match(/\S+/g);
+        if (parts.length > 1) {
+          if (parts[0] == "*") {
+            return {
+              name: elm.replace("*", "").trim(),
+              description: "current branch",
+              icon: "⭐️",
+              // priority: 100,
```
QuiiBz, line 106: Any reason why the priority is commented?

sullivan-sean: Not sure, I just left this as it was. Maybe @mattschrage has more context?

mattschrage: I think @brendanfalk and I were going back and forth on whether the current branch should *always* be the top suggestions (which is the behavior that uncommenting would enable). 

Can be removed.

mattschrage (approved):
LGTM. ⛵️

Issue #292, opened 2021-06-12
[Rustc] Rustc autocomplete and more comprehensive cargo


dvtkrlbs:
### CLI Tool
Rustc

### Useful Links
https://doc.rust-lang.org/rustc/command-line-arguments.html

### Top Workflows
This mostly relates to cargo. I dont if it is possible but you can pass rustc arguments from cargo and some of the flags are hard to memorize and even sometimes hard to find documentation about. Fig could really help with that.

PR #314, opened 2021-06-19
feat: add rustc + cargo rustc spec

Monkeyank:
Feature
#292 No support for rustc or cargo rustc
Adds support for rustc and cargo rustc
**Additional info:**

### DIFF
--- a/dev/cargo.ts
+++ b/dev/cargo.ts
@@ -81,6 +81,7 @@ export const completionSpec: Fig.Spec = {
                 description: `Ex: verbose' or 'feature="serde"'`,
               },
             },
+
             {
               name: "-L",
               description: "Adds a directory to the library search path",
@@ -90,27 +91,32 @@ export const completionSpec: Fig.Spec = {
                   description: "The kind of search path",
                   suggestions: [
                     {
-                      name: "KIND=dependency",
+                      name: "dependency",
+                      insertValue: "KIND=dependency",
                       description:
                         "Only search for transitive dependencies in this directory",
                     },
                     {
-                      name: "KIND=crate",
+                      name: "crate",
+                      insertValue: "KIND=crate",
                       description:
                         "Only search for this crate's direct dependencies in this directory",
                     },
                     {
-                      name: "KIND=native",
+                      name: "native",
+                      insertValue: "KIND=native",
                       description:
                         "Only search for native libraries in this directory",
                     },
                     {
-                      name: "KIND=framework",
+                      name: "framework",
+                      insertValue: "KIND=framework",
                       description:
                         "Only search for macOS frameworks in this directory",
                     },
                     {
-                      name: "KIND=all",
+                      name: "all",
+                      insertValue: "KIND=all",
                       description:
                         "Search for all library kinds in this directory",
                     },
@@ -121,6 +127,7 @@ export const completionSpec: Fig.Spec = {
                   name: "path",
                   description:
                     "Path to search for external crates and libraries",
+                  template: "folders",
                 },
               ],
             },
@@ -133,15 +140,18 @@ export const completionSpec: Fig.Spec = {
                   description: "Kind of library",
                   suggestions: [
                     {
-                      name: "KIND=dylib",
+                      name: "dylib",
+                      insertValue: "KIND=dylib",
                       description: "A native dynamic library",
                     },
                     {
-                      name: "KIND=static",
+                      name: "static",
+                      insertValue: "KIND=static",
                       description: "A native static library",
                     },
                     {
-                      name: "KIND=framework",
+                      name: "framework",
+                      insertValue: "KIND=framework",
                       description: "A macOS framework",
                     },
                   ],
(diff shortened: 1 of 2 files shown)


dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+                    },
+                    {
+                      name: "KIND=framework",
+                      description:
+                        "Only search for macOS frameworks in this directory",
+                    },
+                    {
+                      name: "KIND=all",
+                      description:
+                        "Search for all library kinds in this directory",
```
QuiiBz, line 115: I think it's hard to understand the name of the suggestions with this `KIND=`. Let's simply write `dependency`, `crate`, `native`...

dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+                    },
+                  ],
+                  isOptional: true,
+                },
+                {
+                  name: "path",
+                  description:
+                    "Path to search for external crates and libraries",
+                },
```
QuiiBz, line 124: Looks like this can have a template of folders?

dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+                      description: "A native dynamic library",
+                    },
+                    {
+                      name: "KIND=static",
+                      description: "A native static library",
+                    },
+                    {
+                      name: "KIND=framework",
+                      description: "A macOS framework",
+                    },
```
QuiiBz, line 146: Same as above

dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+                suggestions: ["auto", "always", "never"],
+              },
+            },
+          ],
+        },
+      ],
+      args: {
+        template: "filepaths",
+        name: "file",
+        description: "Additional files to compile",
```
QuiiBz, line 433: Based on this description, it seems that this arg is variadic

dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+        },
+        {
+          name: "--example",
+          description: "Builds only the specified example",
+          args: {
+            name: "Name",
+          },
+        },
+        {
+          name: "--examples  ",
```
QuiiBz, line 479: 2 extra spaces

dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+          args: {
+            name: "Info",
+          },
+        },
+        {
+          name: "--target-dir",
+          description: "Directory for all generated artifacts",
+          args: {
+            name: "Directory",
+          },
```
QuiiBz, line 541: Can this arg have a template of `folders`?

dev/cargo.ts
```diff
@@ -60,6 +60,582 @@ export const completionSpec: Fig.Spec = {
...
+        },
+        {
+          name: "--manifest-path",
+          description: "Path to Cargo.toml",
+          args: {
+            name: "Path",
+          },
```
QuiiBz, line 548: Based on the description, this arg can have a template of `filepaths`

dev/rustc.ts
```diff
@@ -0,0 +1,339 @@
+export const completion: Fig.Spec = {
+  name: "rustc",
+  description: "CLI for rustc",
+  subcommands: [],
```
QuiiBz, line 4: No need to set an empty array if this command doesn't have any subcommands

dev/rustc.ts
```diff
@@ -0,0 +1,339 @@
...
+              description: "Only search for native libraries in this directory",
+            },
+            {
+              name: "KIND=framework",
+              description: "Only search for macOS frameworks in this directory",
+            },
+            {
+              name: "KIND=all",
+              description: "Search for all library kinds in this directory",
+            },
```
QuiiBz, line 52: Same as above for this `KIND=` thing

dev/rustc.ts
```diff
@@ -0,0 +1,339 @@
...
+      name: ["-v", "--verbose"],
+      description: "Use verbose output",
+    },
+    {
+      name: "--extern",
+      description: "Specify where an external library is located",
+      args: {
+        name: "Path",
+        description: "Path where crate can be found",
+      },
```
QuiiBz, line 312: Looks like this arg can have a template of `folders`

QuiiBz (changes requested):
Rustaceans will be very happy to have those specs!

dev/cargo.ts
Monkeyank: The required argument value provided has to be `KIND="type"`, so instead should I change the name to `dependency`, `crate`, `native` ... . Then, add the insert value prop, and add the full value, ex: `KIND=crate`?

QuiiBz: Agree, that would be better for the users to have a clear and concise name. Let's do this!

QuiiBz (approved):
LGTM!

PR #316, opened 2021-06-20
Laravel Valet autocomplete

floq-design:
NEW laravel valet autocomplete

N/A

New autocomplete

**Additional info:**
Autocomplete spec for laravel (v8) valet (v2.15.x)

### DIFF
--- a/dev/valet.ts
+++ b/dev/valet.ts
@@ -1,0 +1,99 @@
+export const completion: Fig.Spec = {
+  name: "valet",
+  description:
+    "Valet is a Laravel development environment for macOS minimalists.",
+  subcommands: [
+    {
+      name: "forget",
+      description:
+        'Run this command from a "parked" directory to remove it from the parked directory list.',
+    },
+    {
+      name: "install",
+      description:
+        "This will configure and install Valet and DnsMasq. In addition, the daemons Valet depends on will be configured to launch when your system starts.",
+    },
+    {
+      name: "log",
+      description: "View a list of logs which are written by Valet's services.",
+    },
+    {
+      name: "link",
+      description: "Link the current working directory to Valet.",
+    },
+    {
+      name: "on-latest-version",
+      description: "Determine if this is the latest version of Valet.",
+    },
+    {
+      name: "paths",
+      description: 'View all of your "parked" paths.',
+    },
+    {
+      name: "park",
+      description:
+        "Register the current working (or specified) directory with Valet.",
+    },
+    {
+      name: "parked",
+      description: "Display all the current sites within parked paths.",
+    },
+    {
+      name: "restart",
+      description: "Restart the Valet daemons.",
+    },
+    {
+      name: "secure",
+      description: "Secure the given domain with a trusted TLS certificate",
+    },
+    {
+      name: "start",
+      description: "Start the Valet daemons.",
+    },
+    {
+      name: "stop",
+      description: "Stop the Valet daemons.",
+    },
+    {
+      name: "tld",
+      description: "Link the current working directory to Valet.",
+      args: {
+        name: "tld",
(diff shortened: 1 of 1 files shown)


QuiiBz:
Hello, could you change the base branch to `master`? Looks like the current one is your Wordpress branch

floq-design:
> Hello, could you change the base branch to `master`? Looks like the current one is your Wordpress branch

TBH I should've branched master before I worked on the wp-cli... So I branched master for this valet work and of course I get all the master commits.. I deleted the `wp.ts` file from this branch so just the one file.. 
If I merge back into master then there will be both `wp.ts` & `valet.ts` which happy to do but didn't want to confuse more than already have!

Lemme know..

dev/valet.ts
```diff
@@ -0,0 +1,718 @@
...
+      description: "Generate a publicly accessible URL for your project",
+      options: [
+        {
+          name: "--auth",
+          description: "enforce basic auth on tunnel endpoint, 'user:password'",
+        },
+        {
+          name: "--authtoken",
+          description: "ngrok.com authtoken identifying a user",
+        },
```
mattschrage, line 449: Does this take the auth token as an argument?

dev/valet.ts
```diff
@@ -0,0 +1,718 @@
...
+            suggestions: [
+              { name: "true", insertValue: "'true'" },
+              { name: "false", insertValue: "'false'" },
+              { name: "both", insertValue: "'both'" },
+            ],
+          },
+        },
+        {
+          name: "--config",
+          description: "path to config files; they are merged if multiple",
```
mattschrage, line 464: Seems like this takes a `path` as an argument.

```
args: {
name: "path",
template: "filepath"
}
```

dev/valet.ts
```diff
@@ -0,0 +1,718 @@
...
+        },
+        {
+          name: "--host-header",
+          description:
+            "set Host header; if 'rewrite' use local address hostname",
+        },
```
mattschrage, line 470: Seems like `--host-head` takes an argument?

mattschrage (changes requested):
Thanks so much for contributing @floq-design! 🙌

I could be wrong as I'm not familiar with the `valet` CLI, but it seems like several options are missing arguments.

(You don't need to provide generators or suggestions for them, but if the option doesn't have a corresponding `arg` object then the completions we generate at runtime will be incorrect!)

Otherwise, great work! Excited to get this merged in :)

QuiiBz:
Hello @floq-design, I rebased your branch to remove the previous commits from `wp`. Do you want to resolve Matt's review?

dev/valet.ts
floq-design: Good spot.. suggestion added

dev/valet.ts
floq-design: Yes it does.. args with suggestions added

dev/valet.ts
floq-design: Yes it does.. added arg

floq-design:
> I could be wrong as I'm not familiar with the `valet` CLI...

I'm not familiar with the `share` command for Valet.
These options I believe are based on `ngrok` which I'm also unfamiliar with.
https://ngrok.com/docs#tunnel-definitions

I've reworked a lot of the options based upon the ngrok docs and what the `valet share -h` command shows.
Also added your suggestions - thanks.

floq-design:
Sorry made a right hash of pulling latest build into branch and merging.

QuiiBz:
Are the commits from wp intentional?

QuiiBz:
I again removed the +200 commits. Please pay attention when you pull/merge branches.

QuiiBz (approved):
Laravel users are going to love this one!

PR #317, opened 2021-06-20
[FASTLY] Update CLI spec to 0.31.0

kailan:
This release introduced a rework of the `--version` option, a new `--autoclone` option, and two new subcommands.


QuiiBz (approved):
Nice update!

Issue #312, opened 2021-06-19
[git] missing some options for `git branch`


brendanfalk:
For instance, the `-r` flag is missing.

Need to do a full check to make sure we have everything

QuiiBz:
Merged in #319!

PR #319, opened 2021-06-22
Add missing options for git branch spec

Monkeyank:
Completion Spec Update

#312 
`git branch` is currently missing certain options such as `-r`

Adds missing options for `git branch`

**Additional info:**

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -3580,21 +3580,25 @@ export const completionSpec: Fig.Spec = {
         },
         {
           name: ["-t", "--track"],
+          exclusive: ["--no-track"],
           description:
             "When creating a new branch, set up 'upstream' configuration.",
           args: [
             {
+              name: "branch",
               generators: gitGenerators.branches,
             },
             {
+              name: "start point",
               isOptional: true,
-              generators: gitGenerators.branches,
+              generators: gitGenerators.commits,
             },
           ],
         },
 
         {
           name: ["--no-track"],
+          exclusive: ["--track", "-t"],
           description:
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -3538,6 +3545,121 @@ export const completionSpec: Fig.Spec = {
...
+        {
+          name: ["-t", "--track"],
+          description:
+            "When creating a new branch, set up 'upstream' configuration.",
+          args: [
+            {
+              generators: gitGenerators.branches,
+            },
+            {
+              isOptional: true,
```
QuiiBz, line 3590: Any reason why there is an `isOptional` here?

dev/git.ts
```diff
@@ -3538,6 +3545,121 @@ export const completionSpec: Fig.Spec = {
...
+              generators: gitGenerators.branches,
+            },
+          ],
+        },
+
+        {
+          name: ["--no-track"],
+          description:
+            "Do not set up 'upstream' configuration, even if the branch.autoSetupMerge configuration variable is true.",
+          args: [
```
QuiiBz, line 3600: Looks like we can add an `exclusive: ["-t", "--track"]` here. And the invert for `--track` option

QuiiBz (changes requested):
The git spec is getting really big! Left you some comments

dev/git.ts
Monkeyank: The 2nd generator is supposed to be commits, which defines the start-point for tracking, which is an optional arg.

mattschrage (approved):
Looks good to me! 👍

PR #322, opened 2021-06-22
chore(eslint): add `no-missing-spec-export` rule

fedeci:
In this PR we add an ESLint rule that ensure that each spec has a `completionSpec` export so that Typescript no longer has to transform the code assuming that the first export is the correct one.
In particular the ESLint rule groups all `NamedExports` in one `exports` array:
```js
export const foo = 100
export { bar, baz }
```
and then checks the number of exports of the file:
- if `0`: it raises an error asking the user to add a `completionSpec` export;
- if `1`: it checks if the export is correctly named fixing the export name otherwise;
- if `> 1`: checks if any export is named `completionSpec` and it raises an error if none was found.

### DIFF
--- a/dev/flutter.ts
+++ b/dev/flutter.ts
@@ -661,7 +661,7 @@ const run = [
 
 // spec
 
-const completionSpec = {
+export const completionSpec: Fig.Spec = {
   name: "flutter",
   description: "Run flutter command.",
   subcommands: [
--- a/dev/adb.ts
+++ b/dev/adb.ts
@@ -148,7 +148,7 @@ const reverseConnectionSuggestions: Fig.Suggestion[] = [
   },
 ];
 
-export const completion: Fig.Spec = {
+export const completionSpec: Fig.Spec = {
   name: "adb",
   description: "Android Debug Bridge",
   subcommands: [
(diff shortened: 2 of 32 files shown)


dev/flutter.ts
```diff
@@ -661,7 +661,7 @@ const run = [
 
 // spec
 
-const completionSpec = {
+export const completionSpec: Fig.Spec = {
   name: "flutter",
```
fedeci, line 665: Are we sure that this spec is correct?

QuiiBz: This spec is working fine but you're right - it shouldn't. Our compiler transforms the first exported variable to a `var` named `completionSpec`, without the `export` keyword. Since the `flutter` spec already has the right name, and that `const` is transformed to `var`, it produces the right output (`const completionSpec` -> `var completionSpec`).

We should still update this with `export` and the type to follow the conventions.

QuiiBz:
As stated in https://github.com/withfig/autocomplete/pull/330, we will update this behavior later, so this rule is not needed for now. Thanks anyway, we might re-open this PR later!

dev/flutter.ts
fedeci: I was not referring to the `export...` part, but to the TS typecheck failing

PR #325, opened 2021-06-23
chore(eslint): add no-duplicate-options-subcommands rule

### LINKED ISSUE #323: Add ESLint check to determine if the same flag (or subcommand) appears more than once.
Screen Shot 2021-06-22 at 6 46 50 PM

fedeci:
Closes #323 
Failing tests are related as I have not yet updated all the specs. I'll wait for a review to do that.

### DIFF
--- a/dev/expo-cli.ts
+++ b/dev/expo-cli.ts
@@ -1323,14 +1323,8 @@ export const completionSpec: Fig.Spec = {
       ],
     },
     {
-      name: "url",
-      description: "",
+      name: ["url", "u"],
       options: [
-        {
-          name: ["-w", "--web"],
-          description: "Return the URL of the web app\n",
-          args: {},
-        },
         {
           name: ["-a", "--android"],
           description:
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -1513,7 +1513,7 @@ export const completionSpec: Fig.Spec = {
         },
 
         {
-          name: "--dry-run",
+          name: ["-n", "--dry-run"],
           description: "Do everything except actually send the updates.",
         },
 
@@ -2007,7 +2007,7 @@ export const completionSpec: Fig.Spec = {
             "By default, Git will report, to the server, commits reachable from all local refs to find common commits in an attempt to reduce the size of the to-be-received packfile",
         },
         {
-          name: ["--dry-run"],
+          name: "--dry-run",
           description: "Show what would be done, without making any changes.",
         },
         {
@@ -3395,7 +3395,7 @@ export const completionSpec: Fig.Spec = {
             "Skip move or rename actions which would lead to an error condition",
         },
         {
-          name: "--dry-run",
+          name: ["-n", "--dry-run"],
           description: "Do nothing; only show what would happen",
         },
         {
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -1513,7 +1513,7 @@ export const completionSpec: Fig.Spec = {
         },
 
         {
-          name: ["-n", "--dry-run"],
+          name: "--dry-run",
           description: "Do everything except actually send the updates.",
         },
 
@@ -2029,7 +2029,7 @@ export const completionSpec: Fig.Spec = {
             "Before fetching, remove any local tags that no longer exist on the remote if --prune is enabled",
         },
         {
-          name: ["-n", "--no-tags"],
+          name: "--no-tags",
           description:
             "By default, tags that point at objects that are downloaded from the remote repository are fetched and stored locally. This option disables this automatic tag following",
         },
@@ -2280,10 +2280,6 @@ export const completionSpec: Fig.Spec = {
           args: {
             name: "pattern",
           },
-        },
-        {
-          name: "--source",
-          description: "show source",
         },
       ],
       args: [
@@ -3075,7 +3071,7 @@ export const completionSpec: Fig.Spec = {
             "Automatically setup .git/objects/info/alternates to share the objects with the source repository",
         },
         {
-          name: ["-n", "--dry-run"],
+          name: "--dry-run",
           description: "Do nothing; only show what would happen",
         },
         {
@@ -3399,7 +3395,7 @@ export const completionSpec: Fig.Spec = {
             "Skip move or rename actions which would lead to an error condition",
         },
         {
-          name: ["-n", "--dry-run"],
+          name: "--dry-run",
           description: "Do nothing; only show what would happen",
         },
         {
--- a/dev/heroku.ts
+++ b/dev/heroku.ts
@@ -3149,31 +3149,7 @@ export const spec: Fig.Spec = {
       description: "display node-foreman version",
     },
     {
-      name: "oauth:create",
-      description: "create a new OAuth authorization",
-      options: [
-        {
-          name: ["-d", "--description"],
-          description: "set a custom authorization",
-          args: {},
-        },
-        {
-          name: ["-S", "--short"],
-          description: "only output token",
-        },
-        {
-          name: ["-j", "--json"],
-          description: "output in json format",
-        },
-        {
-          name: ["-s", "--scope"],
-          description: "set custom OAuth scopes",
-          args: {},
-        },
-      ],
-    },
-    {
-      name: "oauth",
+      name: "authorizations",
       description: "list OAuth authorizations",
       options: [
         {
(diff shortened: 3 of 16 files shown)


scripts/eslint-plugin-fig-linter/rules/no-duplicate-options-subcommands.js
```diff
@@ -0,0 +1,56 @@
...
+                names = [nameProp.value];
+                break;
+              case "ArrayExpression":
+                // e.g. name: ["-v", "--version"]
+                names = nameProp.value.elements;
+                break;
+              default:
+                return;
+            }
+            names.forEach((v) => {
```
QuiiBz, line 40: I would prefer a more verbose name for `v` - looks like it's a `node`

QuiiBz (changes requested):

mattschrage (approved):
This is awesome and by the looks of things will help catch a lot of common errors. Thanks so much for contributing @fedeci 🎉

Everything looks good to me. Once you've addressed @QuiiBz's comment, we can get this merged in :)

QuiiBz:
That's going to be very helpful! Also, before merging this could you run `fig lint:fix` so this new rule can remove the duplicated options/subcommands that are failing right now?

fedeci:
@QuiiBz nope, I have to do it manually. An autofixer would not make sense in this case because we can't predict which option is correct or wrong.

QuiiBz:
> @QuiiBz nope, I have to do it manually. An autofixer would not make sense in this case because we can't predict which option is correct or wrong.

You're right, I was thinking about something else. Let's do this so we can get this merged!

dev/expo-cli.ts
```diff
@@ -1323,14 +1323,8 @@ export const completionSpec: Fig.Spec = {
...
     {
-      name: "url",
-      description: "",
+      name: ["url", "u"],
       options: [
-        {
-          name: ["-w", "--web"],
-          description: "Return the URL of the web app\n",
-          args: {},
-        },
```
fedeci, line 1333: It was mentioned twice in the `expo` docs. I guess that this is a secondary result.

dev/heroku.ts
```diff
@@ -3149,31 +3149,7 @@ export const spec: Fig.Spec = {
       description: "display node-foreman version",
     },
     {
-      name: "oauth:create",
```
fedeci, line 3152: This file was such a mess, full of these non-existent commands. I may have missed something. But reviewing 5000 lines of file is mindblowing. We should provide some tool to pretty output current subcommands structure.

QuiiBz:
There is an issue in the current implementation:

For example, in the `gh` spec, the rule detected `-h` which is an aliase of `--help` - it shouldn't flag aliases.
Same for the `git` spec, the rule detected `-n` from `--dry-run` and `--no-tags`, while those are not in the same subcommand/option.

fedeci:
@QuiiBz `-h` is not a valid alias for subcommands. 
e.g. `gh repo create`
```bash
FLAGS
  -y, --confirm               Skip the confirmation prompt
  -d, --description string    Description of the repository
      --enable-issues         Enable issues in the new repository (default true)
      --enable-wiki           Enable wiki in the new repository (default true)
  -h, --homepage URL          Repository home page URL
      --internal              Make the new repository internal
      --private               Make the new repository private
      --public                Make the new repository public
  -t, --team name             The name of the organization team to be granted access
  -p, --template repository   Make the new repository based on a template repository

INHERITED FLAGS
  --help   Show help for command
```
Note that just `--help` is inherited.

dev/git.ts
```diff
@@ -3075,7 +3071,7 @@ export const completionSpec: Fig.Spec = {
             "Automatically setup .git/objects/info/alternates to share the objects with the source repository",
         },
         {
-          name: ["-n", "--dry-run"],
+          name: "--dry-run",
           description: "Do nothing; only show what would happen",
```
fedeci, line 3075: `-n` used by `--no-checkout` (git clone)

dev/git.ts
```diff
@@ -2029,7 +2029,7 @@ export const completionSpec: Fig.Spec = {
             "Before fetching, remove any local tags that no longer exist on the remote if --prune is enabled",
         },
         {
-          name: ["-n", "--no-tags"],
+          name: "--no-tags",
```
fedeci, line 2032: `-n` used by `--no-stat`(git pull)

QuiiBz:
I tough you only removed duplicated options/subcommands caught by this new rule, sorry for this.

dev/heroku.ts
mattschrage: Can you expand on this? There is definitely room to improve the tooling around the completion specs. 

We could definitely add something like this to the repo. Is this what you're imagining?

`npm run list-subcommands <cli>`

mattschrage:
So this looks good to me. What do you think @cstrnt?

dev/heroku.ts
cstrnt: This was autogenerated from the source code of the heroku cli.

dev/psql.ts
```diff
@@ -215,7 +215,7 @@ export const completionSpec: Fig.Spec = {
         "Set the record separator for unaligned output to a zero byte. This is useful for interfacing, for example, with xargs -0. This is equivalent to pset recordsep_zero.",
     },
     {
-      name: ["-l", "--single-transaction"],
+      name: ["-1", "--single-transaction"],
```
cstrnt, line 218: Uhm was this done by you or the rule?

dev/serverless.ts
```diff
@@ -355,11 +355,11 @@ export const completion: Fig.Spec = {
...
     {
       name: "package",
-      description: "Packages a Serverless service",
+      description: "Package a Serverless service",
     },
     {
       name: "plugin",
-      description: "Install/uninstall and add/remove a plugin to your service",
+      description: "Handle plugins",
     },
```
cstrnt, line 363: This was also done manually, right?

cstrnt (commented):
So the rule itself is pretty awesome! Good job! Also the selector for which looks like css is mindblowing to me! I would prefer splitting up those manual changes to another PR because they are kinda unrelated to the eslint rule

dev/psql.ts
fedeci: The flag was duplicate and checking on the `psql` docs I discovered it was wrong.

dev/serverless.ts
fedeci: Yes it was, cause I had to remove a duplicate `plugin` subcommand and from what I can understand from the docs it does not only `install/uninstall` plugins.

dev/heroku.ts
fedeci: @mattschrage yes it is

QuiiBz:
@fedeci could you add back the `heroku` OAuth subcommands that have been removed? Thanks

fedeci:
Is there any place in the docs that still mentions those? It seems to me that they have been replaced by `client:create...`.

QuiiBz:
> Is there any place in the docs that still mentions those? It seems to me that they have been replaced by `client:create...`.

Just checked the documentation and you're right, apologize for this.

dev/heroku.ts
```diff
@@ -6236,133 +6032,6 @@ export const spec: Fig.Spec = {
         },
       ],
     },
-    {
-      name: "run:detached",
-      description: "run a detached dyno, where output is sent to your logs",
-      options: [
```
QuiiBz, line 6242: This subcommand still exists: https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-run-detached

dev/heroku.ts
```diff
@@ -6236,133 +6032,6 @@ export const spec: Fig.Spec = {
...
-        {
-          name: "ru",
-          variadic: true,
-        },
-      ],
-    },
-    {
-      name: "run",
-      description: "run a one-off process inside a heroku dyno",
-      options: [
```
QuiiBz, line 6316: This subcommand still exists: https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-run

dev/heroku.ts
```diff
@@ -3482,35 +3307,6 @@ export const spec: Fig.Spec = {
...
-        {
-          name: "--json",
-          description: "output in json format",
-        },
-      ],
-    },
-    {
-      name: "clients:info",
-      description: "show details of an oauth client",
-      options: [
```
QuiiBz, line 3498: Those 2 subcommands still exist: https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-clients & https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-clients-info-id

dev/heroku.ts
```diff
@@ -3379,16 +3214,6 @@ export const spec: Fig.Spec = {
         },
       ],
     },
-    {
-      name: "authorizations",
-      description: "list OAuth authorizations",
-      options: [
```
QuiiBz, line 3385: This subcommand still exists: https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-authorizations

QuiiBz (changes requested):

dev/heroku.ts
fedeci: They was all duplicates.

QuiiBz: I see, thanks for the clarification.

QuiiBz (approved):
LGTM!

For more information, the removed subcommands in the `heroku` spec were either duplicated, either renamed in the new versions of the CLI, so everything is good.

dev/heroku.ts
fedeci: You're welcome!

PR #329, opened 2021-06-25
Fix type issues

judofyr:

Bug fixes related to type errors not caught by TypeScript. See the commit messages for more details.

The definitions don't really match up with the types. It just turns out that TypeScript is so lenient that these don't actually cause a type error to occur.


QuiiBz:
Really good catches! That's strange that typescript doesn't complain about bad types.

Could you run `yarn lint:fix` to fix the linting issues?

judofyr:
> Could you run `yarn lint:fix` to fix the linting issues?

Done! Rebased and forced-pushed.

QuiiBz (approved):
Thanks for those fixes!

PR #330, opened 2021-06-25
Use consistent naming of exported spec

judofyr:

Refactoring.

Different files uses different names for the exported spec.

Change all files to use `export const completionSpec` (which is the documented version).

### DIFF
--- a/dev/adb.ts
+++ b/dev/adb.ts
@@ -148,7 +148,7 @@ const reverseConnectionSuggestions: Fig.Suggestion[] = [
   },
 ];
 
-export const completion: Fig.Spec = {
+export const completionSpec: Fig.Spec = {
   name: "adb",
   description: "Android Debug Bridge",
   subcommands: [
--- a/dev/command.ts
+++ b/dev/command.ts
@@ -1,4 +1,4 @@
-export const completion: Fig.Spec = {
+export const completionSpec: Fig.Spec = {
   name: "command",
   description: "run a command!",
   args: {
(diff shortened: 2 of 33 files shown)


QuiiBz:
Hello! Completion specs don't **need** to be named `completionSpec`, they can be anything. The only criteria is that they should be exported.

Could you share the link in the documentation where it says that they must be named `completionSped`? We will update this.

judofyr:
Huh, I could have sworn that I read somewhere that they should be exported with `completionSpec`, but seems like it's not explicitly mentioned anywhere.

I found it super confusing as a JavaScript/TypeScript developer that the the exported name can be _anything_. I'm used to that the name is very significant so I was a bit stumped. At first I thought that the ones who had `export const completion` (not `completionSpec`) was somehow treated differently.

`export default` is kinda the convention if you don't want to name it. Is there a reason you're not using that?

cstrnt:
Hey @judofyr,

I agree with you that `export default` would be the most idiomatic way of doing it in JS. But saying that we do have a custom way of parsing the completion specs which requires a custom way of doing that. But it is on our roadmap to change it , but not right now :)

QuiiBz:
As Tim stated, we agree with you. This behavior will be updated to be more clear and intuitive, it's in our roadmap. 

Thanks anyway for this PR!

fedeci:
It should be quite quick to implement `export default`, also from the swift engine. Waiting just increases the number of custom specs that will fail because using the old behavior.

QuiiBz:
Fig's engine is currently being rewritten so this is not the top priority right now. I really think that updating the specs later won't be a time-consuming task.

fedeci:
Oh, that's definitely the priority then!

Issue #103, opened 2021-04-07
[ripgrep] and [grep] request for specs


classicrob:
### CLI Tool
What CLI tool is missing autocompletion in Fig?

ripgrep and grep

### Useful Links
Please add some links to the docs of that tool or the `--help`text

https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md

https://www.gnu.org/software/grep/manual/grep.html

### Top Workflows
What are the most used commands or workflows in that CLI tool (for you)?

Searching text within directories and providing the appropriate amount of context before and after. Piping that into other outputs.

M-ZubairAhmed:
Hey how can i contribute to this?

cstrnt:
Hey there @classicrob,

`grep` is already implemented in a very basic manner, but `ripgrep` isn't yet.
Do you know how those tools work? We would really appreciate if you could help us :)

classicrob:
@cstrnt , how can I help?

cstrnt:
@classicrob Either by creating a new completion spec for `ripgrep` yourself or having a look at the current `grep` spec and checking if that one is correct 😊

QuiiBz:
`ripgrep` has been merged into #332!

classicrob:
Thank you!

PR #332, opened 2021-06-28
[Ripgrep] Add Completion Spec

Monkeyank:
Feature
#103 
Adds basic Ripgrep completion spec
**Additional info:**

### DIFF
--- a/dev/rg.ts
+++ b/dev/rg.ts
@@ -1,0 +1,649 @@
+import { isOptionalChain } from "typescript";
+
+export const completion: Fig.Spec = {
+  name: "rg",
+  description:
+    "recursively searches current directory for lines matching a pattern",
+  subcommands: [],
+  options: [
+    {
+      name: ["-A", "--after-context"],
+      description: "Shows NUM lines after each match.",
+      args: {
+        name: "Num",
+      },
+    },
+    {
+      name: "--auto-hybrid-regex",
+      description:
+        "When this flag is used, ripgrep will dynamically choose between supported regex engines depending on the features used in a pattern",
+    },
+    {
+      name: ["-B", "--before-context"],
+      description: "Shows NUM lines before each match",
+      args: {
+        name: "Num",
+      },
+    },
+    {
+      name: "--binary",
+      description:
+        "Enabling this flag will cause ripgrep to search binary files",
+    },
+    {
+      name: "--block-buffered",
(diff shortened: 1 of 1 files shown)


dev/rg.ts
```diff
@@ -0,0 +1,649 @@
...
+      name: "--hidden",
+      description: "Searches hidden files and directories",
+    },
+    {
+      name: "--iglob",
+      description:
+        "Includes or excludes files and directories for searching that match given the glob",
+      args: {
+        name: "GLOB",
+        variadic: true,
```
QuiiBz, line 209: Based on the description, I think this arg can have a `template: ["filepaths", "folders"]`

dev/rg.ts
```diff
@@ -0,0 +1,649 @@
...
+        name: "Replacement Text",
+      },
+    },
+    {
+      name: ["-z", "--search-zip"],
+      description: "Search in compressed files",
+    },
+    {
+      name: ["-S, --smart-case"],
+      description: "Seaches case insensitively if the pattern is all lowercase",
```
QuiiBz, line 486: Type at `Seaches` -> `Searches`

dev/rg.ts
```diff
@@ -0,0 +1,649 @@
...
+            description:
+              "Sort by the last accessed time on a file. Always single-threaded",
+          },
+          {
+            name: "created",
+            description:
+              "Sort by the creation time on a file. Always single-threaded",
+          },
+        ],
+      },
```
QuiiBz, line 551: I would love to see the args moved to a top-level const to avoid duplication!

QuiiBz (changes requested):
Excited about this spec! I left you some comments

dev/rg.ts
```diff
@@ -0,0 +1,649 @@
...
+    },
+    {
+      name: "--colors",
+      description: "Specifies color settings for use in the output",
+      args: {
+        name: "Color Spec",
+        variadic: true,
+        description: "{type}:{attribute}:{value}, ex: `match:fg:magenta`",
```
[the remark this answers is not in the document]
Monkeyank: The` {type}:{attribute}:{value}` isn't 3 separate arguments, rather it's one argument, and you can supply multiple of these.
Ex:  `rg --colors 'line:bg:yellow' 'match:fg:magenta'`

QuiiBz (approved):
LGTM!

PR #334, opened 2021-06-28
[Open] Update Spec

Monkeyank:
Feature
Currently open doesn't have any options, but just suggests filepaths/directories.
Adds options!
**Additional info:**

### DIFF
--- a/dev/open.ts
+++ b/dev/open.ts
@@ -1,3 +1,14 @@
+const appGenerator = (path: string): Fig.Generator => ({
+  script: `ls -1 ${path}`,
+  postProcess: (out) => {
+    return out.split("\n").map((line) => ({
+      name: line,
+      icon: `fig://${path}/${line}`,
+      priority: line.endsWith(".app") && 76,
+    }));
+  },
+});
+
 export const completionSpec: Fig.Spec = {
   name: "open",
   description: "open files using default application",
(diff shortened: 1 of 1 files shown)


dev/open.ts
```diff
@@ -1,7 +1,82 @@
...
+    {
+      name: "-h",
+      description:
+        "Searches header locations for a header whose name matches the given string and then opens it",
+    },
+    {
+      name: "--args",
+      description:
+        "All remaining arguments are passed to the opened application in the argv parameter to main().  These arguments are not opened or interpreted by the open tool",
+    },
```
mattschrage, line 76: I think this should have a variadic `arg` object.

mattschrage (changes requested):
I'm excited for this one! (Especially if you can get the icon API working for the applications - will be very useful and cool)

dev/open.ts
```diff
@@ -1,7 +1,99 @@
...
+          },
+          {
+            script: "ls -1p ~/Applications",
+            postProcess: function (out) {
+              return out.split("\n").map((line) => ({
+                name: line,
+                icon: `fig://~/Applications/${line}`,
+                priority: line.endsWith(".app/") && 76,
+              }));
+            },
```
QuiiBz, line 29: Those generators are amazing! But there is a bit of duplication, let's create a top-level function to generate a generator:

```ts
const appsGenerator = (path: string): Fig.Generator => ({
  script: `ls -1p ${path}`,
  postProcess: (out) => {
    return out.split("\n").map((line) => ({
      name: line,
      icon: `fig://${path}/${line}`,
      priority: line.endsWith(".app/") && 76,
    }));
  },
});
```

And we can use it like this:
```suggestion
          appsGenerator("/Applications"),
          appsGenerator("~/Applications"),
```

Monkeyank: Haha, just was about to push this up 😅

QuiiBz (approved):
Amazing job! Love the generators with the icons

Issue #102, opened 2021-04-07
[sed] request for spec


classicrob:
### CLI Tool
sed

### Useful Links
https://www.gnu.org/software/sed/manual/sed.html

### Top Workflows
What are the most used commands or workflows in that CLI tool (for you)?

Generally just text management and formatting! I use Obsidian for my notes, so all of my work is in text files and I would love an easier way to write commands for sed!

Nutlope:
Would love to work on this as my first issue. Thanks for adding the tag @cstrnt. Do you recommend I just look up the top most common workflows with sed and include them with the PR?

QuiiBz:
Merged in #335, thanks everyone!

classicrob:
Thank you!

PR #335, opened 2021-06-29
[Sed] Add Completion Spec

Monkeyank:
Completion Spec
#102 

**Additional info:**

### DIFF
--- a/dev/sed.ts
+++ b/dev/sed.ts
@@ -1,0 +1,259 @@
+const labelArg: Fig.Arg = {
+  name: "label",
+  isOptional: true,
+};
+
+const textArg: Fig.Arg = {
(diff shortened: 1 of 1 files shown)


dev/sed.ts
```diff
@@ -0,0 +1,259 @@
...
+  ],
+  args: [
+    {
+      name: "command",
+    },
+    {
+      name: "file",
+      template: "filepaths",
+      variadic: true,
+    },
```
QuiiBz, line 257: I think this arg is optional

QuiiBz (commented):
I'm not familiar with the Sed CLI but this looks fine to me!

QuiiBz (approved):
LGTM!

Issue #270, opened 2021-06-07
Make it possible to increase the default generator timeout


nagypeterjob:
### Completion Spec
What completion spec does this enhancement refer to? Or is it Fig itself?
Each long-running generators

### Description
Describe the issue that you're seeing.

### Steps to reproduce
Create any long running generator & see it fail after 5 seconds

### Expected result
What should happen?

There are generators that create long-running request. It would be nice to have a way to set this timeout for a given generator.

### Actual result
What happened.

SeparateRecords:
This is fixed now! Generators can individually specify a timeout in ms. IMO the default timeout is sane and should be kept as is, long-running generators should have their timeout increased individually.

Issue #96, opened 2021-04-06
Add ability to specify custom timeout when running a script


mschrage:
Currently, this defaults to 5 seconds. Should we add a new property to the `Generator` that allows this to be customized?

See #91 for an example of when this is necessary. (tl;dr long running commands like `adb` and `xcodebuild` frequently timeout)

alexnaiman:
I think that an additional property, as mentioned, would be really helpful.
Some commands like those from `adb` take more time to execute just the first time. The subsequent runs after should take less time to execute.

falky97:
Will actually be a little hard to do. We set time outs for the entire autocomplete function, not just the generators... To do this, we would have to remove timeouts for the whole function... This is doable, will just need to think it through so we can support this but make sure Fig still handles non-generator errors.

Also, we need to add a loading icon when a generator is running...

mattschrage:
Just added #104

mschrage:
@fedeci this is worth implementing, if you have some spare time - I don’t think it’ll be too tricky anymore.

PR #337, opened 2021-06-30
[AWS] [Lambda] Spec

nagypeterjob:

**Additional info:**

### DIFF
--- a/dev/aws/lambda.ts
+++ b/dev/aws/lambda.ts
@@ -357,7 +357,7 @@ const listCustomSIDGenerator = async (
 const MultiSuggestionsGenerator = async (
   context: string[],
   executeShellCommand: Fig.ExecuteShellCommandFunction,
-  enabled: any[]
+  enabled: Record<string, string>[]
 ) => {
   try {
     const list: Fig.Suggestion[][] = [];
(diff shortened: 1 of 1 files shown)


dev/aws/lambda.ts
```diff
@@ -1,3 +1,1001 @@
...
+    cache: {
+      ttl: ttl,
+    },
+  },
+
+  listDestinationConfigArns: {
+    custom: async function (context, executeShellCommand) {
+      // Getting sqs queues is implemented, altough it has a huge performance toll.
+      // It seems Fig rejects long-running promises after a time.
+      // I am currently investigating if this is the case.
```
QuiiBz, line 747: Fig as a default timeout of 5 seconds for generators. Does `aws sqs list-queues` takes longer than that?

nagypeterjob: Yes, we would need more then 5 sec here

nagypeterjob: but this is not a blocker

QuiiBz: Let's follow #270, #96 to get this implemented later

QuiiBz (approved):

Issue #144, opened 2021-04-26
[pm2] request for spec


mpekala23:
### CLI Tool
pm2

### Useful Links
https://pm2.keymetrics.io/docs/usage/quick-start/

### Top Workflows
`pm2 start`
`pm2 restart all`
`pm2 stop`

cstrnt:
Hey there @mpekala23 
that's a good one! Are you proficient with PM2 and would be willing to create it yourself? This would be highly appreciated :)

QuiiBz:
Merged in #338, thanks everyone!

PR #338, opened 2021-06-30
[PM2] Add Spec

Monkeyank:
Completion Spec: PM2 Spec
#144 
**Additional info:**

### DIFF
--- a/dev/pm2.ts
+++ b/dev/pm2.ts
@@ -106,6 +106,7 @@ const sharedOptions: Fig.Option[] = [
     description: "Specify pid file",
     args: {
       name: "pid",
+      template: "filepaths",
     },
   },
   {
@@ -293,7 +294,6 @@ const sharedOptions: Fig.Option[] = [
     args: {
       name: "Node Args",
       description: "–node-args=`–debug=7001 –trace-deprecation`",
-      variadic: true,
     },
   },
   { name: "–-no-color", description: "skip colors" },
@@ -343,7 +343,7 @@ const sharedOptions: Fig.Option[] = [
 
 export const completion: Fig.Spec = {
   name: "pm2",
-  description: "",
+  description: "Daemon process manager",
   subcommands: [
     {
       name: "start",
@@ -374,6 +374,7 @@ export const completion: Fig.Spec = {
       description: "Deploy your json",
       args: {
         name: "File or environment",
+        template: "filepaths",
       },
     },
     {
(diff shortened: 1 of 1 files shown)


dev/pm2.ts
```diff
@@ -0,0 +1,799 @@
...
+    name: ["-e", "--error"],
+    description: "Specifies error log file",
+    args: pathArg,
+  },
+  {
+    name: ["-p", "--pid"],
+    description: "Specify pid file",
+    args: {
+      name: "pid",
+    },
```
QuiiBz, line 109: Looks like we can add a template of filepaths here

dev/pm2.ts
```diff
@@ -0,0 +1,799 @@
...
+      variadic: true,
+    },
+  },
+  {
+    name: "–-node-args",
+    description: "Space delimited arguments to pass to node in cluster mode",
+    args: {
+      name: "Node Args",
+      description: "–node-args=`–debug=7001 –trace-deprecation`",
+      variadic: true,
```
QuiiBz, line 296: Technically, this arg isn't variadic because each of these arguments are inside a string, which is counted as a single arg

dev/pm2.ts
```diff
@@ -0,0 +1,799 @@
...
+    name: "–-deep-monitoring",
+    description:
+      "Enables all monitoring tools (equivalent to –v8 –event-loop-inspector –trace)",
+  },
+  { name: ["-h, –-help"], description: "Outputs usage information" },
+];
+
+export const completion: Fig.Spec = {
+  name: "pm2",
+  description: "",
```
QuiiBz, line 346: Could you add a description?

dev/pm2.ts
```diff
@@ -0,0 +1,799 @@
...
+          isOptional: true,
+        },
+      ],
+    },
+    {
+      name: "deploy",
+      description: "Deploy your json",
+      args: {
+        name: "File or environment",
+      },
```
QuiiBz, line 377: Can this arg have a template of filepaths?

dev/pm2.ts
```diff
@@ -0,0 +1,799 @@
+const generators: Record<string, Fig.Generator> = {
+  jsonFileGenerator: {
+    template: "filepaths",
+    filterTemplateSuggestions: function (paths) {
+      const suffix = ".json";
+      return paths.filter((file) => {
+        return file.name.indexOf(suffix, this.length - suffix.length) >= 0;
+      });
```
QuiiBz, line 8: I think we can simplify this a lot:
```suggestion
    filterTemplateSuggestions: function (paths) {
      const suffix = ".json";
      return paths.filter((file) => file.endsWidth(suffix));
```

QuiiBz (changes requested):
A really useful specs! Left you some comments

QuiiBz (approved):
LGTM!

Issue #281, opened 2021-06-10
[hasura] Hasura CLI


deathemperor:
### CLI Tool
Hasura

### Useful Links
https://hasura.io/docs/latest/graphql/core/hasura-cli/index.html

### Top Workflows
* hasura migration
* hasura console

QuiiBz:
Merged in #345, thanks everyone!

PR #345, opened 2021-07-03
[Hasura] Add Spec

Monkeyank:
Feature
#281

### DIFF
--- a/dev/hasura.ts
+++ b/dev/hasura.ts
@@ -1,0 +1,732 @@
+const stringArg: Fig.Arg = {
+  name: "string",
+};
+
+const sharedOptions: Fig.Option[] = [
+  {
+    name: "--envfile",
+    description: ".env filename to load ENV vars from",
+    args: {
+      name: "string",
+      template: "filepaths",
+    },
+  },
+  {
+    name: ["-h", "--help"],
+    description: "Help for hasura",
+  },
+  {
+    name: "--log-level",
+    description: "Log level",
+    args: {
+      name: "string",
+      suggestions: ["DEBUG", "INFO", "WARN", "ERROR", "FATAL"],
+      default: "INFO",
+    },
+  },
+  {
+    name: "--no-color",
+    description: "Do not colorize output",
+  },
+  {
+    name: "--project",
+    description: "Directory where commands are executed",
+    args: {
+      name: "string",
+      template: "folders",
+    },
+  },
+  {
+    name: "--skip-update-check",
+    description: "Skip automatic update check on command execution",
+  },
+];
+
+const adminSecretOption: Fig.Option = {
+  name: "--admin-secret",
+  description: "Admin secret for Hasura GraphQL engine",
+  args: stringArg,
+};
+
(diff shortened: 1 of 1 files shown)


dev/hasura.ts
```diff
@@ -0,0 +1,732 @@
...
+const adminSecretOption: Fig.Option = {
+  name: "--admin-secret",
+  description: "Admin secret for Hasura GraphQL engine",
+  args: stringArg,
+};
+
+const certificateAuthorityOption: Fig.Option = {
+  name: "--certificate-authority",
+  description: "Path to a cert file for the certificate authority",
+  args: stringArg,
```
QuiiBz, line 54: This can have a template of `filepaths`

dev/hasura.ts
```diff
@@ -0,0 +1,732 @@
...
+            name: "string",
+            template: "filepaths",
+          },
+        },
+        ...sharedOptions,
+      ],
+      args: {
+        name: "shell",
+        isOptional: true,
+        suggestions: ["bash", "zsh"],
```
QuiiBz, line 184: Would love to see `fish` here - more and more peoples are using it!

dev/hasura.ts
```diff
@@ -0,0 +1,732 @@
...
+          name: "--no-browser",
+          description: "Do not automatically open console in browser",
+        },
+        {
+          name: "--static-dir",
+          description:
+            "Directory where static assets mentioned in the console html template can be served from",
+          args: {
+            name: "string",
+            template: "filepaths",
```
QuiiBz, line 237: Based on the name and description, shouldn't the template be `folders` instead of `filepaths`?

dev/hasura.ts
```diff
@@ -0,0 +1,732 @@
...
+          name: "delete",
+          description: "Clear migrations from local project and server",
+          options: [
+            {
+              name: "--all",
+              description: "Clears all migrations for selected database",
+            },
+            {
+              name: "--force",
+              description: "Executes operation without any confirmation",
```
QuiiBz, line 503: Adding `isDangerous` could allow users to identify that this option can be dangerous

dev/hasura.ts
```diff
@@ -0,0 +1,732 @@
...
+          options: [
+            {
+              name: "--database-name",
+              description:
+                "Database name for which the current migrations / seeds belong to",
+              args: stringArg,
+            },
+            {
+              name: "--force",
+              description: "Do not ask for confirmation",
```
QuiiBz, line 646: Same as above

QuiiBz (changes requested):

dev/hasura.ts
Monkeyank: The only supported completions by asura are `bash` and `zsh`. [Docs Section](https://hasura.io/docs/latest/graphql/core/hasura-cli/hasura_completion.html)

QuiiBz: Alright, thanks for the clarification!

QuiiBz (approved):
LGTM!

Issue #339, opened 2021-07-01
Fig wraps `insertValue` in single quotes when text is already wrapped in double quotes 


mattschrage:
If `insertValue` contains string wrapped in double quotes, Fig should not automatically wrap it again in single quotes
<img width="728" alt="Screen Shot 2021-06-30 at 11 32 12 PM" src="https://user-images.githubusercontent.com/3155109/124076928-bd3f6800-d9fb-11eb-85b6-c2513862f4da.png">

**Reproduce:**

1. have a git repo where a file with a space in the name has changed.
```
$ git status --short
M  fig/Bridging/WebBridge.swift
M  "fig/Companion Window/WebViewController.swift"
M  fig/Defaults.swift
M  "fig/Window Management/WindowManager.swift"
```
2. Type `git add` and then attempt to insert one of the files wrapped in quotes. 

3. Fig will insert the selected suggestion with an additional set of single quotes. eg. `'"fig/Window Management/WindowManager.swift"'``

mattschrage:
Alternatively, we could view this as a bug in the `git` spec and strip the double quotes.

PR #346, opened 2021-07-03
fix: Double quotes in git staged files generator

QuiiBz:
Bug fix

#339

Removes double-quotes using a regex at the start and/or at the end of each line of `git status --short` output

Screenshot 2021-07-03 at 11 54 31

**Additional info:**

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -178,7 +178,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
       });
 
       return items.map((item) => {
-        const file = item.file;
+        const file = item.file.replace(/^"|"$/g, "");
         let ext = "";
 
         try {


mattschrage:
it looks like double quotes inside the quotes that git adds are also escaped. See your second test case.

That would insert 'with\”quote' which would not match the file.

QuiiBz:
Isn't this the expected behavior? Fig added singles-quotes and so the file path is valid and can be added:

Screenshot 2021-07-03 at 18 56 31

mattschrage:
Ah okay! I wasn’t sure how the quotation goes would work. Is the original file `test”.ts`?

If you’ve tested it, the code looks good to me. Just wasn’t sure if this was the intended behavior.

QuiiBz:
>  Is the original file `test”.ts`?
Yes, it is! Just tested with more edge-cases to make sure everything will work properly.

mattschrage (approved):

PR #347, opened 2021-07-05
Add prisma

xlanex6:
Add prisma autocomplete based on CLI docs reference 
https://www.prisma.io/docs/reference/api-reference/command-reference

### DIFF
--- a/dev/prisma.ts
+++ b/dev/prisma.ts
@@ -1,21 +1,38 @@
+const commonOptions: Fig.Option[] = [
+  {
+    name: ["-h", "--help"],
+    description: "Display this help message",
+  },
+];
+const schemaOptions: Fig.Option[] = [
+  {
+    name: "--schema",
+    description: "Custom path to your Prisma schema",
+    args: {
+      name: "Schema path",
+      template: "filepaths",
+      isOptional: true,
+    },
+  },
+];
+
 export const completionSpec: Fig.Spec = {
   name: "prisma",
   description:
     "Prisma is a modern DB toolkit to query, migrate and model your database",
+  icon:
+    "https://raw.githubusercontent.com/prisma/docs/main/src/images/favicon-16x16.png",
   subcommands: [
     {
       name: "init",
       description: "Setup Prisma for your app",
       subcommands: [
-        {
-          name: ["-h", "--help"],
-          description: "Display this help message",
-        },
+        ...commonOptions,
         {
           name: "--datasource-provider",
           description: "Define the datasource provider to use",
-
           args: {
+            description: "Choose provider",
             suggestions: ["PostgreSQL", "MySQL", "SQLServer", "SQLite"],
             default: "postgres",
           },
@@ -24,7 +41,9 @@ export const completionSpec: Fig.Spec = {
           name: "--url",
           description: "Define a custom datasource url",
           args: {
-            name: "Full url",
+            name: "full url",
+            template: "filepaths",
+            isOptional: false,
           },
         },
       ],
@@ -33,10 +52,7 @@ export const completionSpec: Fig.Spec = {
       name: "introspect",
       description: "Get the datamodel of your database",
       options: [
-        {
-          name: ["-h", "--help"],
-          description: "Display this help message",
-        },
+        ...commonOptions,
         {
           name: "--force",
           description: "Ignore current Prisma schema file",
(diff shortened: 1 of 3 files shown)


xlanex6:
👍🏻

dev/prisma.ts
```diff
@@ -0,0 +1,271 @@
...
+            ...commonOptions,
+            ...schemaOptions,
+            {
+              name: "--skip-seed",
+              description: "Skip triggering seed",
+            },
+            {
+              name: "--skip-generate",
+              description: "Skip triggering generators (e.g. Prisma Client)",
+            },
```
QuiiBz, line 161: I would prefer to avoid duplication by using a top-level const of `Fig.Options[]`. The same options are duplicated here: https://github.com/withfig/autocomplete/pull/347/files#diff-5d922d2aefb78ba4c511d31448c65e489ffa0975ca389832ebcb486b6e5d5748R127-R134

dev/prisma.ts
```diff
@@ -0,0 +1,271 @@
...
+            {
+              name: ["-f", "--force"],
+              description: "Skip the confirmation prompt",
```
QuiiBz, line 164: Same for the `isDangerous` flag

dev/prisma.ts
```diff
@@ -0,0 +1,271 @@
...
+          options: [
+            ...commonOptions,
+            ...schemaOptions,
+            {
+              name: "--applied",
+              description: " Record a specific migration as applied",
+              args: {
+                name: "migration file path",
+                template: "filepaths",
+                isOptional: false,
```
QuiiBz, line 188: You don't have to specify `isOptional` if the value is false. Same at https://github.com/withfig/autocomplete/pull/347/files#diff-5d922d2aefb78ba4c511d31448c65e489ffa0975ca389832ebcb486b6e5d5748R197 and https://github.com/withfig/autocomplete/pull/347/files#diff-5d922d2aefb78ba4c511d31448c65e489ffa0975ca389832ebcb486b6e5d5748R46

dev/prisma.ts
```diff
@@ -0,0 +1,271 @@
...
+            {
+              name: "--accept-data-loss",
+              description:
+                "Ignore data loss warnings. This option is required if as a result of making the schema changes, data may be lost.",
+            },
+          ],
+        },
+        {
+          name: "seed",
+          isDangerous: true,
```
QuiiBz, line 253: I don't think we need `isDangerous` here

dev/prisma.ts
```diff
@@ -0,0 +1,271 @@
...
+  name: "prisma",
+  description:
+    "Prisma is a modern DB toolkit to query, migrate and model your database",
+  icon:
+    "https://raw.githubusercontent.com/prisma/docs/main/src/images/favicon-16x16.png",
+  subcommands: [
+    {
+      name: "init",
+      description: "Setup Prisma for your app",
+      subcommands: [
```
QuiiBz, line 29: These aren't subcommands but rather options - same for https://github.com/withfig/autocomplete/pull/347/files#diff-5d922d2aefb78ba4c511d31448c65e489ffa0975ca389832ebcb486b6e5d5748R263

QuiiBz (changes requested):
Thanks a lot for this contribution! I left you some comments to resolve before merging this.

Also, could you add `prisma` to this array? https://github.com/withfig/autocomplete/blob/c6c5e9ad75b1d7971b538c28837cf0a154f57fe8/dev/yarn.ts#L130
This is similar to `npx prisma ...`, but for yarn. It allows autocomplete while typing `yarn prisma ...` if the current project has prisma installed.

dev/prisma.ts
```diff
@@ -0,0 +1,271 @@
...
+  },
+];
+const schemaOptions: Fig.Option[] = [
+  {
+    name: "--schema",
+    description: "Custom path to your Prisma schema",
+    args: {
+      name: "Schema path",
+      template: "filepaths",
+      isOptional: true,
```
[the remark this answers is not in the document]
xlanex6: By default, is `./schema.prisma` || `./prisma/schema.prisma`, if decide to change default, you have to specify path. 

https://www.prisma.io/docs/reference/api-reference/command-reference#arguments-4

Here's the reason why I suggest `isOptional: true`

dev/prisma.ts
xlanex6: `isDangerous` was here to notice this command is `--preview-feature` on Prisma CLI and is not recommended for use in production environments.

https://www.prisma.io/docs/reference/api-reference/command-reference#db-seed-preview

xlanex6:
@QuiiBz 
I made all your suggestions. 
Free to talk about `isOptional` for schema option.  https://github.com/withfig/autocomplete/pull/347#discussion_r664052551

dev/prisma.ts
QuiiBz: Thanks for the clarification!

dev/prisma.ts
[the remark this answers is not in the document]
QuiiBz: Is it possible to only have the `--schema` option without the path argument? If so, `isOptional` makes sense, but reading the docs makes me think that whenever you specify `--schema`, we need a path argument.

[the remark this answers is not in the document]
xlanex6: True, my bad.
You're right.

xlanex6:
@QuiiBz I remove `isOptional`, you're right. If you provide  `--schema` option, you need to pass file path schema. 
My bad

QuiiBz:
> @QuiiBz I remove `isOptional`, you're right. If you provide `--schema` option, you need to pass file path schema.
> My bad

That's fine, no worries!

QuiiBz (approved):
LGTM!

Issue #73, opened 2021-03-24
[git] git add to offer a wildcard extension


MartinPacker:
When I update the docs for one of my GitHub projects ([md2pptx](https://github.com/MartinPacker/md2pptx) as it happens) I generate a bunch of `user_guide.*` files. It's my practice to add them with the following command:

```
git add user_guide.*

```

Today autocomplete offers each file individually. In my case I have several I want to add all at once. I like the individual file prompting but I'd like autocomplete to also offer the generic `user_guide.*`.

mattschrage:
I like this idea a lot! To summarize: we should check to see if any of the staged files share a common prefix and then add the corresponding wildcard pattern as a suggestion?

brendanfalk:
**Another user comment**

I was trying to git add a folder of changes eg:
* `app/web/files/change.js`
* `app/web/files/otherthing.js`

I wanted the autocomplete to stop at git add app/web/files, but there's no way to do this outside of breaking the autocomplete interface by typing a nonsensical character, or escaping the interface.

QuiiBz:
Merged in #349, thanks to everyone!

brendanfalk:
Going to reopen this. This PR uses two generators which is going to be really slow. We should do this with 1 generator.

QuiiBz:
There are many other specs that use 2 generators, I don't think that really slow. We should benchmark this to see if it's really slow or not.

QuiiBz:
Merged again in #358, using one generator (which is now really faster!)

PR #349, opened 2021-07-07
[git] Add wildcard generator

Monkeyank:
Add Wildcard Generator for staging files in git
#73 
Checks for matching paths of committed files and adds a suggestion for a wildcard.

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -200,6 +200,53 @@ const gitGenerators: Record<string, Fig.Generator> = {
       });
     },
   },
+
+  // Wildcard
+  wildcard_files_for_staging: {
+    script: "git status --short",
+    postProcess: (out) => {
+      const output = filterMessages(out);
+      if (output.startsWith("fatal:")) {
+        return [];
+      }
+
+      const items = output.split("\n").map((file) => {
+        file = file.slice(0, file.lastIndexOf("/") + 1);
+        file = file.trim();
+        const arr = file.split(" ");
+        return arr.slice(1).join(" ").trim();
+      });
+
+      if (items.length < 2) {
+        return [];
+      }
+
+      var dirArr = [];
+      var currentDir = items[0];
+      var count = 1;
+      for (var i = 1; i < items.length; i++) {
+        if (items[i].includes(currentDir) && i + 1 !== items.length) {
+          currentDir = items[i];
+          count++;
+        } else if (count >= 2) {
+          dirArr.push(currentDir);
+          currentDir = items[i];
+          count = 1;
+        } else {
+          currentDir = items[i];
+          count = 1;
+        }
+      }
+
+      return dirArr.map((name) => {
+        return {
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -200,6 +200,53 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+        return arr.slice(1).join(" ").trim();
+      });
+
+      if (items.length < 2) {
+        return [];
+      }
+
+      var dirArr = [];
+      var currentDir = items[0];
+      var count = 1;
```
QuiiBz, line 226: Let's use `const` and `let` instead of `var`

dev/git.ts
```diff
@@ -200,6 +200,53 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+          currentDir = items[i];
+          count++;
+        } else if (count >= 2) {
+          dirArr.push(currentDir);
+          currentDir = items[i];
+          count = 1;
+        } else {
+          currentDir = items[i];
+          count = 1;
+        }
```
QuiiBz, line 238: We can simplify this a bit because some parts are duplicated, something like this (needs to be tested):
```suggestion
        currentDir = items[i];
        
        if (items[i].includes(currentDir) && i + 1 !== items.length) {
          count++;
        } else {
          if (count >= 2) {
            dirArr.push(currentDir);
          }
          
          count = 1;
        }
```

dev/git.ts
```diff
@@ -200,6 +200,53 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+      if (output.startsWith("fatal:")) {
+        return [];
+      }
+
+      const items = output.split("\n").map((file) => {
+        file = file.slice(0, file.lastIndexOf("/") + 1);
+        file = file.trim();
+        const arr = file.split(" ");
+        return arr.slice(1).join(" ").trim();
+      });
```
QuiiBz, line 218: I would prefer not to re-assign variables. In this case, we can do something like this:
```suggestion
      const items = output.split("\n").map((file) => {
        const arr = file.slice(0, file.lastIndexOf("/" + 1))
          .trim()
          .split(" ");

        return arr.slice(1)
          .join(" ")
          .trim()
      });
```

QuiiBz (changes requested):

dev/git.ts
```diff
@@ -200,6 +200,52 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+
+      if (items.length < 2) {
+        return [];
+      }
+
+      const dirArr = [];
+      let currentDir = items[0];
+      let count = 1;
+      for (var i = 1; i < items.length; i++) {
```
QuiiBz, line 228: A last `var` to change for `let` and we will be ready to merge!

QuiiBz (changes requested):

QuiiBz (approved):
LGTM!

Issue #351, opened 2021-07-08
[git] Add support for `git switch`


jakobgilbert99:
The `git` completion spec is missing support for `switch`. Users have emailed us to suggest we add support for it, as they find it interrupts their workflow to find it's missing.

Monkeyank:
https://git-scm.com/docs/git-switch

QuiiBz:
Merged in #352, thanks to everyone!

PR #352, opened 2021-07-09
[git] switch subcommand

Monkeyank:
`switch` subcommand for `git`

#351

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -4600,17 +4600,31 @@ export const completionSpec: Fig.Spec = {
           name: ["-c", "--create"],
           description:
             "Create a new branch named <new-branch> starting at <start-point> before switching to the branch",
-          args: {
-            name: "new branch",
-          },
+          args: [
+            {
+              name: "new branch",
+            },
+            {
+              name: "start point",
+              isOptional: true,
+              generators: gitGenerators.commits,
+            },
+          ],
         },
         {
           name: ["-C", "--force-create"],
           description:
             "Similar to --create except that if <new-branch> already exists, it will be reset to <start-point>",
-          args: {
-            name: "new branch",
-          },
+          args: [
+            {
+              name: "new branch",
+            },
+            {
+              name: "start point",
+              isOptional: true,
+              generators: gitGenerators.commits,
+            },
+          ],
         },
         {
           name: ["-d", "--detach"],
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -4592,6 +4592,149 @@ export const completionSpec: Fig.Spec = {
...
+      name: "switch",
+      description: "Switch branches",
+      options: [
+        {
+          name: ["-c", "--create"],
+          description:
+            "Create a new branch named <new-branch> starting at <start-point> before switching to the branch",
+          args: {
+            name: "new branch",
+          },
```
QuiiBz, line 4605: Based on the description, it seems like there is another argument named `start-point`

dev/git.ts
```diff
@@ -4592,6 +4592,149 @@ export const completionSpec: Fig.Spec = {
...
+        },
+        {
+          name: ["-C", "--force-create"],
+          description:
+            "Similar to --create except that if <new-branch> already exists, it will be reset to <start-point>",
+          args: {
+            name: "new branch",
+          },
```
QuiiBz, line 4613: Same as above

dev/git.ts
```diff
@@ -4592,6 +4592,149 @@ export const completionSpec: Fig.Spec = {
...
+        },
+        {
+          name: ["-d", "--detach"],
+          description:
+            "Switch to a commit for inspection and discardable experiments",
+        },
+        {
+          name: ["--guess", "--no-guess"],
+          description:
+            "If <branch> is not found but there does exist a tracking branch in exactly one remote (call it <remote>) with a matching name",
```
QuiiBz, line 4623: Same as about for `branch`

dev/git.ts
```diff
@@ -4592,6 +4592,149 @@ export const completionSpec: Fig.Spec = {
...
+        },
+        {
+          name: ["-f", "--force"],
+          description: "An alias for --discard-changes",
+        },
+        {
+          name: "--discard-changes",
+          description:
+            "Proceed even if the index or the working tree differs from HEAD. Both the index and working tree are restored to match the switching target",
```
QuiiBz, line 4632: These options can have an `isDangerous` option

QuiiBz (changes requested):

dev/git.ts
Monkeyank: I don't think this actually has `args`, but guesses the what the `branch-name` is supposed to be

QuiiBz:
Could you resolve the conflict so we can get this merged?

QuiiBz (approved):
LGTM!

PR #354, opened 2021-07-10
Add Django Spec

### LINKED ISSUE #298: [Django] CLI Request
### CLI Tool
Django (`django-admin` and `python3 manage.py`)

### Useful Links
[https://docs.djangoproject.com/en/3.2/ref/django-admin/](https://docs.djangoproject.com/en/3.2/ref/django-admin/)

wilhelmklopp:
This PR adds the Django spec, meaning auto complete for `django-admin <command>` and `python manage.py <command>`

Closes #298 

**Additional info:**

I originally went down the path of dynamically auto generating this whole spec at runtime based on the help output provided by `django-admin` script. This has the benefit that we'd show suggestions for exactly those commands that are available, including custom-defined management commands. 

But I think there are some problems with that dynamic approach at the moment. Notably, from a user's perspective, it's slower and requires regular manual set up (ie running `fig set:path`). So i've decided to just stick to a fully static spec for now and circle back to the dynamic approach in the future. If you're curious, [this PR](https://github.com/wilhelmklopp/autocomplete/pull/1) contains some of my explorations in the more dynamic world.

I opened a couple of issues that would pave the path a little bit more towards making dynamic at-runtime generated specs work super well:
- https://github.com/withfig/fig/issues/225
- https://github.com/withfig/fig/issues/250

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,6 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
-  }
+  },
+  "editor.formatOnSave": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,6 +2,5 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
-  },
-  "editor.formatOnSave": true
+  }
 }
--- a/dev/django-admin.ts
+++ b/dev/django-admin.ts
@@ -552,14 +552,20 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
         description:
           "[{full,file,never}] Controls '#: filename:line' lines. If the option is 'full' (the default if not given), the lines include both file name and line number. If it's 'file', the line number is omitted. If it's 'never', the lines are suppressed (same as --no-location). --add-location requires gettext 0.19 or newer.",
         args: {
-           name: "location",
-           isOptional: true,
-           default: 'full'
-           suggestions: [
-               { name: "full", description: "the lines include both file name and line number" },
-               { name: "file", description: "the line number is omitted" },
-               { name: "never", description: "the lines are suppressed (same as --no-location)"},
-           ],
+          name: "location",
+          isOptional: true,
+          default: "full",
+          suggestions: [
+            {
+              name: "full",
+              description: "the lines include both file name and line number",
+            },
+            { name: "file", description: "the line number is omitted" },
+            {
+              name: "never",
+              description: "the lines are suppressed (same as --no-location)",
+            },
+          ],
         },
       },
       {
--- a/dev/django-admin.ts
+++ b/dev/django-admin.ts
@@ -93,7 +93,8 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
       },
       {
         name: ["--noinput", "--no-input"],
-        description: "Tells Django to NOT prompt the user for input of any",
+        description:
+          "Tells Django to NOT prompt the user for input of any kind.",
       },
       {
         name: "--database",
@@ -122,18 +123,21 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
     options: [
       {
         name: ["--noinput", "--no-input"],
-        description: "Tells Django to NOT prompt the user for input of any",
+        description:
+          "Tells Django to NOT prompt the user for input of any kind.",
       },
       {
         name: "--database",
-        description: "Nominates the database to use. Defaults to the",
+        description:
+          'Nominates the database to use. Defaults to the "default" database.',
         args: {
           name: "DATABASE",
         },
       },
       {
         name: "--include-stale-apps",
-        description: "Deletes stale content types including ones from",
+        description:
+          "Deletes stale content types including ones from previously installed apps that have been removed from INSTALLED_APPS.",
       },
       {
         name: "--skip-checks",
@@ -162,7 +166,8 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
       },
       {
         name: "--fail-level",
-        description: "Message level that will cause the command to exit with",
+        description:
+          "Message level that will cause the command to exit with a non-zero status. Default is ERROR.",
         args: {
           name: "{CRITICAL,ERROR,WARNING,INFO,DEBUG}",
           suggestions: ["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"],
@@ -191,14 +196,16 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
     options: [
       {
         name: ["--locale", "-l"],
-        description: "Locale(s) to process (e.g. de_AT). Default is to",
+        description:
+          "Locale(s) to process (e.g. de_AT). Default is to process all. Can be used multiple times.",
         args: {
           name: "LOCALE",
         },
       },
       {
         name: ["--exclude", "-x"],
-        description: "Locales to exclude. Default is none. Can be used",
+        description:
+          "Locales to exclude. Default is none. Can be used multiple times.",
         args: {
           name: "EXCLUDE",
         },
@@ -209,7 +216,8 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
       },
       {
         name: ["--ignore", "-i"],
-        description: "Ignore directories matching this glob-style pattern.",
+        description:
+          "Ignore directories matching this glob-style pattern. Use multiple times to ignore more.",
         args: {
           name: "PATTERN",
         },
@@ -224,14 +232,16 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
     options: [
       {
         name: "--database",
-        description: "Nominates a database onto which the cache tables will",
+        description:
+          'Nominates a database onto which the cache tables will be installed. Defaults to the "default" database.',
         args: {
           name: "DATABASE",
         },
       },
       {
         name: "--dry-run",
-        description: "Does not create the table, just prints the SQL that",
+        description:
+          "Does not create the table, just prints the SQL that would be run.",
       },
     ],
     description: "Creates the tables needed to use the SQL cache backend.",
--- a/dev/django-admin.ts
+++ b/dev/django-admin.ts
@@ -1248,10 +1248,7 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
 ].map((subcommand) => {
   return {
     ...subcommand,
-    options: {
-      ...subcommand.options,
-      ...ALWAYS_AVAILABLE_OPTIONS,
-    },
+    options: [...subcommand.options, ...ALWAYS_AVAILABLE_OPTIONS],
   };
 });
 
--- a/dev/django-admin.ts
+++ b/dev/django-admin.ts
@@ -704,6 +704,9 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
       {
         name: "email",
         description: "One or more email addresses to send a test email to.",
+        // This argument is not optional when no other arguments are supplied.
+        // But it is optional when either `--managers` or `--admins` are supplied as arguments,
+        // so keeping it as `isOptional` for now to keep things simple.
         isOptional: true,
         variadic: true,
       },
--- a/dev/django-admin.ts
+++ b/dev/django-admin.ts
@@ -1,6 +1,6 @@
 const DJANGO_ICON_URL = "https://github.com/django.png";
 
-const ALWAYS_AVAILABLE_OPTIONS = [
+const ALWAYS_AVAILABLE_OPTIONS: Fig.Option[] = [
   {
     name: ["-h", "--help"],
     description: "show this help message and exit",
@@ -17,6 +17,7 @@ const ALWAYS_AVAILABLE_OPTIONS = [
       "Verbosity level; 0=minimal output, 1=normal output, 2=verbose output, 3=very verbose output",
     args: {
       name: "{0,1,2,3}",
+      suggestions: ["0", "1", "2", "3"],
     },
     priority: 49,
   },
@@ -35,6 +36,7 @@ const ALWAYS_AVAILABLE_OPTIONS = [
       'A directory to add to the Python path, e.g. "/home/djangoprojects/myproject".',
     args: {
       name: "PYTHONPATH",
+      template: "folders",
     },
     priority: 49,
   },
@@ -163,6 +165,7 @@ const DJANGO_NATIVE_COMMANDS: Fig.Subcommand[] = [
         description: "Message level that will cause the command to exit with",
         args: {
           name: "{CRITICAL,ERROR,WARNING,INFO,DEBUG}",
+          suggestions: ["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"],
         },
       },
       {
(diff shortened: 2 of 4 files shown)


dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+      name: "SETTINGS",
+    },
+    priority: 49,
+  },
+  {
+    name: "--pythonpath",
+    description:
+      'A directory to add to the Python path, e.g. "/home/djangoprojects/myproject".',
+    args: {
+      name: "PYTHONPATH",
```
QuiiBz, line 37: When an argument is a path, you can use the [template](https://fig.io/docs/reference/arg#template) property to provide suggestions for a real path. In this case, the `folders` template can be used!

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+      },
+      {
+        name: "--deploy",
+        description: "Check deployment settings.",
+      },
+      {
+        name: "--fail-level",
+        description: "Message level that will cause the command to exit with",
+        args: {
+          name: "{CRITICAL,ERROR,WARNING,INFO,DEBUG}",
```
QuiiBz, line 165: Based on the arg's name, I think we can also add `suggestions` here

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+    description:
+      "Runs the command-line client for specified database, or the default database",
+  },
+  {
+    name: "diffsettings",
+    icon: DJANGO_ICON_URL,
+    options: [
+      {
+        name: "--all",
+        description: "Display all settings, regardless of their value. In",
```
QuiiBz, line 266: Incomplete description?

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+      {
+        name: "--default",
+        description: "The settings module to compare the current settings",
+        args: {
+          name: "MODULE",
+        },
+      },
+      {
+        name: "--output",
+        description: "{hash,unified}",
```
QuiiBz, line 277: Looks like this option takes an argument. Also, we can provide `suggestions` for `hash` and `unified`

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+        name: "--no-startup",
+        description:
+          "When using plain Python, ignore the PYTHONSTARTUP environment variable and ~/.pythonrc.py script.",
+      },
+      {
+        name: ["-i", "--interface"],
+        description:
+          '{ipython,bpython,python}, --interface {ipython,bpython,python} Specify an interactive interpreter interface. Available options: "ipython", "bpython", and "python"',
+        args: {
+          name: "{ipython,bpython,python}",
```
QuiiBz, line 715: Can this arg have `suggestions`?

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+        name: "--skip-checks",
+        description: "Skip system checks.",
+      },
+    ],
+    description: "Installs the named fixture(s) in the database.",
+    args: [
+      {
+        name: "fixture",
+        description: "Fixture labels.",
+        isOptional: false,
```
QuiiBz, line 469: There is no need to specify `isOptional` when the value is `false` - it's assumed by default. Same for underneath

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+        args: {
+          name: "TESTRUNNER",
+        },
+      },
+      {
+        name: ["-t", "--top-level-directory"],
+        description: "Top level of project for unittest discovery.",
+        args: {
+          name: "TOP_LEVEL",
+        },
```
QuiiBz, line 1016: Does this arg need a path to a folder? If so, we can use a `template` of `folders`

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+    {
+      name: "help",
+      description: "Usage and help information for django-admin",
+      icon: DJANGO_ICON_URL,
+      subcommands: DJANGO_NATIVE_COMMANDS.filter(
+        (command) => command.name != "help"
+      ).map((command) => ({
+        name: command.name,
+        icon: DJANGO_ICON_URL,
+      })),
```
QuiiBz, line 1269: Each command will lose other properties such as `args`, `description`, etc. It should be better to spread `command` and keep the `icon` property.

QuiiBz (changes requested):
Thanks a lot for this big PR! I left some comments to resolve before merging this.

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
+const DJANGO_ICON_URL = "https://github.com/django.png";
+
+const ALWAYS_AVAILABLE_OPTIONS = [
+  {
+    name: ["-h", "--help"],
+    description: "show this help message and exit",
+    priority: 49,
```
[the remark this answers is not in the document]
wilhelmklopp: Yeah, my intention with setting `49` as a priority is so that these options (which are available for every single subcommand) show up below the options that are specific to each command.

IMHO the command specific options tend to be more relevant than the always-present ones, so should be shown first.

As an example, when I type `python manage.py test ` I care more about options like `--failfast` ( Tells Django to stop running the test suite after first failed test.) or `--keepdb` (Preserves the test DB between runs.) than I care about `--version` or `--verbosity`

dev/django-admin.ts
wilhelmklopp: Ah good point! I think I actually don't want subcommands here I want "suggestions", but can still keep around name, description, and icon 💡

dev/django-admin.ts
```diff
@@ -0,0 +1,1273 @@
...
+    name: "--version",
+    description: "show program's version number and exit",
+    priority: 49,
+  },
+  {
+    name: ["-v", "--verbosity"],
+    description:
+      "Verbosity level; 0=minimal output, 1=normal output, 2=verbose output, 3=very verbose output",
+    args: {
+      name: "{0,1,2,3}",
```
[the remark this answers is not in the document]
wilhelmklopp: Ah sweet, great idea!! 💡

dev/django-admin.ts
wilhelmklopp: ahhh, very cool ⭐ 😎

.vscode/settings.json
```diff
@@ -2,5 +2,6 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
-  }
+  },
+  "editor.formatOnSave": true
```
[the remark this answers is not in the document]
wilhelmklopp: So for this one, I am actually intentionally suggesting this as an addition to the `.vscode/settings.json` workspace settings. I think having this set enables a better default, works out-of-the-box editing experience for Fig specs, irrespective of what Fig spec authors already have enabled in their global vscode user settings.

But definitely happy to remove this if you all feel differently. What do you think?

wilhelmklopp:
Thanks so much for the review @QuiiBz 🙌 

Have made a bunch of improvements and responded to your comments, so I think this is ready for another review 👀

dev/django-admin.ts
wilhelmklopp: good catch! Have fixed this up

.vscode/settings.json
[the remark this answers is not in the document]
QuiiBz: Agree with you, but I think it should be done in a separate PR. Could you remove this from this PR and open another with this addition?

dev/django-admin.ts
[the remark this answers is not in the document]
QuiiBz: Thanks for the clarification on this. I'm not 100% sure if it's best to have this, since all options with a priority equal or lower than 49 will always be at the bottom, even if we selected one of those option before. It's very useful to have recently selected options at the top, but with a priority of 49 it will break this feature.

dev/django-admin.ts
```diff
@@ -0,0 +1,1282 @@
...
+    name: "remove_stale_contenttypes",
+    icon: DJANGO_ICON_URL,
+    options: [
+      {
+        name: ["--noinput", "--no-input"],
+        description: "Tells Django to NOT prompt the user for input of any",
+      },
+      {
+        name: "--database",
+        description: "Nominates the database to use. Defaults to the",
```
QuiiBz, line 129: Unfinished sentence?

dev/django-admin.ts
```diff
@@ -0,0 +1,1282 @@
...
+  {
+    name: "compilemessages",
+    icon: DJANGO_ICON_URL,
+    options: [
+      {
+        name: ["--locale", "-l"],
+        description: "Locale(s) to process (e.g. de_AT). Default is to",
+        args: {
+          name: "LOCALE",
+        },
```
QuiiBz, line 197: Is there a pre-defined list of locales for this option, or can we set anything? It would be cool to have suggestions for it

QuiiBz (changes requested):

dev/django-admin.ts
QuiiBz: Apologise for this, I didn't saw that it was for the `help` subcommand - it's totally fine like this!

dev/django-admin.ts
wilhelmklopp: I'm pretty unfamiliar with the Internationalisation piece in Django to be honest. I had a quick look but couldn't find a "definitive" list. This is probably the best I was able to find: https://stackoverflow.com/a/3191729/2162900

.vscode/settings.json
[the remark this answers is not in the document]
wilhelmklopp: Done ✅ 

https://github.com/withfig/autocomplete/pull/356

dev/django-admin.ts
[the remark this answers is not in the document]
wilhelmklopp: ahh interesting, yeah that doesn't seem good that it's overriding the "recently selected options" feature.

Is there a way for me to specify that these generic options should have lower priority than the command-specific options, without breaking the recently selected options feature?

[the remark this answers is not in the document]
QuiiBz: Sadly, it's not currently possible. This behavior might change in the future since we are currently porting the core app over to TypeScript, so we might add this feature later. For now, I think we should prioritize what users expected to have: recently selected options at the top (so without a priority <= 49)

dev/django-admin.ts
QuiiBz: Alright, then I don't think we should add this gigantic list of suggestions!

dev/django-admin.ts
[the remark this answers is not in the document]
wilhelmklopp: @QuiiBz Just thinking about this one some more:

image


Could I get the desired behaviour by:
- removing the current setting of priority 49 for the "always available" options (meaning Fig gives them a default of 50)
- setting the priority of all "command specific" options to **51**

Setting them to 51 means they should show up ahead of the "always available" options, unless those have been used previously, right?

I _think_ this could work 🤞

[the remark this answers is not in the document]
wilhelmklopp: Just implemented this in 52eeecaa6a32bbd2a527ff8fd982975db2fce5dc and it seems to work for me locally 🎉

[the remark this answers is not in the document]
mattschrage: Honestly the number should probably be an implementation detail. We could define constants, like `HIGH`, `MEDIUM`, `LOW`, with `LOW` being the default.

dev/django-admin.ts
```diff
@@ -0,0 +1,1290 @@
...
+    ],
+    description:
+      "Sends a test email to the email addresses specified as arguments.",
+    args: [
+      {
+        name: "email",
+        description: "One or more email addresses to send a test email to.",
+        isOptional: true,
+        variadic: true,
+      },
```
mattschrage, line 709: Since this is one or more emails, I don't think it is optional.

```suggestion
      {
        name: "email",
        description: "One or more email addresses to send a test email to.",
        variadic: true,
      },
```

dev/django-admin.ts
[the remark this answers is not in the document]
QuiiBz: That would be great to implement this feature instead of numbers, which are not very easy to understand.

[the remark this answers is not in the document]
wilhelmklopp: Nice, I like LOW, MEDIUM, HIGH 👍 

For the purposes of this PR, how should we proceed? I'm happy with how it works now, I think it checks all the boxes of what we were looking for – but happy to leave off the priority number entirely for now if you all prefer

dev/django-admin.ts
wilhelmklopp: Just checked this out 👀 

As I understand it, you're right that it's not optional UNLESS you pass the --managers or --admin options:
```bash
manage.py sendtestemail: error: You must specify some email recipients, or pass the --managers or --admin options.
```

So does that mean it should stay as `isOptional: true`? Or how can this sort of dynamic be best expressed in a Fig spec?

dev/django-admin.ts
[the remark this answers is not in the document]
QuiiBz: I think your current implementation is fine - it's a bit hacky but it's working really well. We can then update it later if we implement the above constants.

dev/django-admin.ts
QuiiBz: This could be done using `generateSpec` (see https://fig.io/docs/reference/subcommand#generatespec). This function gives a variable named `context`, which is an array of the previous subcommands/options/args. We could then check if this array includes `managers` or `admin`, and return the corresponding subcommand.

Not sure if my explanation was clear enough - tell me if it wasn't the case!

mattschrage: Ah I see @wilhelmklopp! Thanks for clarifying. I would add a comment here that makes a note of this and keep your original code.

@QuiiBz is right that you could use the under-documented `generateSpec` function on `sendtestemail ` to dynamically configure the value of `isOptional`.  My understanding is that `generateSpec` is somewhat expensive, so I'm not sure if it's worth running it for such a small change. cc @brendanfalk

wilhelmklopp: > I would add a comment here that makes a note of this and keep your original code.

Was just about to write this comment, but then GitHub Copilot started trolling me with its suggestion hahaha 😆 

image

wilhelmklopp: Comment added in ce51309 ✅

dev/django-admin.ts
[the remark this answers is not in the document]
mattschrage: I agree! Just opened a new issue for this. #364

dev/django-admin.ts
```diff
@@ -0,0 +1,1293 @@
...
+          suggestions: ["ipython", "bpython", "python"],
+        },
+      },
+      {
+        name: ["-c", "--command"],
+        description:
+          "Instead of opening an interactive shell, run a command as Django and exit.",
+        args: {
+          name: "COMMAND",
+        },
```
mattschrage, line 739: Is this a shell command? If so, add `isCommand: true` and then Fig will continue to provide autocompletion on this argument.
```suggestion
        args: {
          name: "COMMAND",
          isCommand: true,
        },
```

dev/django-admin.ts
```diff
@@ -0,0 +1,1293 @@
...
+      },
+      {
+        name: "--no-location",
+        description: "Don't write '#: filename:line' lines.",
+      },
+      {
+        name: "--add-location",
+        description:
+          "[{full,file,never}] Controls '#: filename:line' lines. If the option is 'full' (the default if not given), the lines include both file name and line number. If it's 'file', the line number is omitted. If it's 'never', the lines are suppressed (same as --no-location). --add-location requires gettext 0.19 or newer.",
+      },
```
mattschrage, line 554: It seems like `--add-location` takes an argument! (I've added it here as a suggestion, but the linter might complain!)

```suggestion
        name: "--add-location",
        description:
          "[{full,file,never}] Controls '#: filename:line' lines. If the option is 'full' (the default if not given), the lines include both file name and line number. If it's 'file', the line number is omitted. If it's 'never', the lines are suppressed (same as --no-location). --add-location requires gettext 0.19 or newer.",
        args: {
           name: "location",
           isOptional: true,
           default: 'full'
           suggestions: [
               { name: "full", description: "the lines include both file name and line number" },
               { name: "file", description: "the line number is omitted" },
               { name: "never", description: "the lines are suppressed (same as --no-location)"},
           ],
        },
      },
```

mattschrage (changes requested):
Looks good to me! Found 2 small things but after that, lets get this merged in! 🎉

dev/django-admin.ts
wilhelmklopp: Just tested this out and it looks like "command" is a bit of a misleading name for this (though it is the one Django uses).

It looks like this takes a python statement and evaluates it, so you can do something like `python manage.py shell -c 'print(2 + 2)'` and it will return `4`

but if you try and do something like `python manage.py shell -c 'echo "hello world"'` it throws a SyntaxError

So that means it should probably stay as it is?

dev/django-admin.ts
wilhelmklopp: ooo, great catch! ✨

dev/django-admin.ts
QuiiBz: In this case I think it would be better to not use `isCommand` since it's not a shell command but a python command.

wilhelmklopp:
This should be ready for a final review @QuiiBz @mschrage 🥳

QuiiBz (approved):
Thanks @wilhelmklopp, looks good to me!

wilhelmklopp:
This PR should be ready to merge 💃 – @mschrage want to take a final look?

mschrage (approved):

Issue #255, opened 2021-06-03
[bazel] Bazel support


rolandtshen:
### CLI Tool
Bazel

### Useful Links
https://docs.bazel.build/versions/4.1.0/guide.html
https://github.com/bazelbuild/bazel

### Top Workflows
bazel run (target)
bazel test (target)
bazel build (target)

Having Fig autosuggest the targets based on BUILD.bazel files (similar to makefiles and how Fig is suggesting make targets rn) would be a massive plus

QuiiBz:
Merged in #355, thanks to everyone!

PR #355, opened 2021-07-10
[bazel] Bazel support

dereklowlind:
New feature request for Bazel
https://github.com/withfig/autocomplete/issues/255

**Additional info:**
this copy works but has an issue where it will only pickup Bazel target suggestions from one Bazel BUILD file. Not sure if using the custom flag in the generator over postProcess would make a difference or not. I think the script command is returning before the for loop opens and prints the contents from all of the Bazel build files to the command line.

### DIFF
--- a/dev/bazel.ts
+++ b/dev/bazel.ts
@@ -1,0 +1,66 @@
+const bazelBuildFiles: Fig.Generator = {
+  script: `FILES=( $(find ./ -name BUILD) ); for f in $FILES; do echo "----$f"; cat "$f"; done`,
+  // returns filepaths and contents in the form below, note the "----" to indicate the filepath
+  // ----.//lib/BUILD
+  // load("@rules_cc//cc:defs.bzl", "cc_library")
+
+  // cc_library(
+  //     name = "hello-time",
+  //     srcs = ["hello-time.cc"],
+  //     hdrs = ["hello-time.h"],
+  //     visibility = ["//main:__pkg__"],
+  // )
+
+  postProcess: function (out) {
+    const lines = out.split("\n");
+    // return lines
+    const targets = [];
+    let currPath = "";
+    for (let i = 0; i < lines.length; i++) {
+      const isFilepath = lines[i].match("----.(.*)/BUILD");
+      const isBazelTarget = lines[i].match('name = "(.*)"');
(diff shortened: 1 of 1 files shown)


dev/bazel.ts
```diff
@@ -0,0 +1,66 @@
...
+      }
+    }
+    return targets;
+  },
+};
+
+export const completionSpec: Fig.Spec = {
+  name: "bazel",
+  description: "Bazel the build system!",
+  options: [
```
QuiiBz, line 40: This should be `subcommands`, not `options`

QuiiBz (changes requested):
Thanks a lot for this PR! I'm not familiar at all with bazel, but does those commands are the most important ones? I see a lot more commands and options in the [CLI reference](https://docs.bazel.build/versions/main/command-line-reference.html), so it would be amazing to have them all! (or at least the most important ones)

dev/bazel.ts
dereklowlind: Would it be better to split the bazel support into two PRs? In issue https://github.com/withfig/autocomplete/issues/255 rolandtshen asks for run, build, and test with a generator for recommending targets. I was thinking we get this "mvp" to rolandtshen to test out and then add in the important/rest of the commands in another PR. I'm not a bazel power user so I might need to find one to help out (maybe from their github) on the what commands are important and testing the tool.

Monkeyank: I've built out all the other options/commands, so I can add these in if you'd like!

QuiiBz: @Monkeyank that would definitely help a lot! Could you push a commit directly here, so we won't need to have two PRs for bazel?

QuiiBz:
@Monkeyank could you commit the other subcommands/options/args that you made, so we can get this merged in?

Monkeyank:
Don't seem to have perms to commit directly to the PR. Also I have only been able to add shared options so far

QuiiBz:
> Don't seem to have perms to commit directly to the PR. Also I have only been able to add shared options so far

Alright, then you can open another PR later when done. We can merge this one to already get the skeleton, and then merge your PR to complete `bazel`!

QuiiBz (approved):
LGTM!

PR #356, opened 2021-07-11
Add editor.formatOnSave

wilhelmklopp:
Pulling this out from https://github.com/withfig/autocomplete/pull/354/files#r667475392

IMO this setting improves the default Fig spec editing dev experience without requiring everyone to manually set this in their global user VSCode settings 🙌 

When saving a Fig spec in VScode, prettier does not currently automatically run.

When saving a Fig spec in VSCode, prettier will automatically run.


QuiiBz (approved):
LGTM!

PR #357, opened 2021-07-11
[FEAT] Add support for pnpm cli

kevinfaveri:
**What kind of change does this PR introduce? (Bugfix, feature, docs update, ...)**

Feature: this PR introduces support for pnpm cli: https://pnpm.io

Fig doesn't support autocomplete for the pnpm cli: https://pnpm.io

Autocomplete file `pnpm.ts` added to the repo, introducing support for autocomplete with the pnpm cli: https://pnpm.io/cli/add

**Additional info:**
I tried to cover all places of their docs for the `pnpm cli`, including workspaces support and recursivity. This is done by reusing `constants`. If there is a way we can improve this PR, please suggest it. It is pretty usable right now, but maybe I missed something that can make this one better.

Some suggestions:
- Have some API for reusing subcommands inside other subcommands, so we do not need to manually create recursivity as I did here...
- Make it possible to organize things in a folder, so I could create a structure like:
```
--- pnpm folder
----- index.ts file, exporting the spec
----- subcommand folder
---------- *subcommand*.ts
----- options folder
---------- *option*.ts
----- generators folder
---------- *generator*.ts
```
- Maybe reorganize the subcommand as multiple specs as done with Docker for avoiding using algo looping through subcommand and better organizing code (though I still get the feeling that we would be repeating a lot of code for importing the specs it would work, so...)

EDIT, some improvements after re-reading my PR: 
- Add suggestions for some args in some subcommand (I was just using the name)... I will try to work on that tomorrow.
- Add support for https://pnpm.io/cli/publish#publish-branch as well as develop a generator that gets git branches


dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+          ({
+            name: item.package.name,
+            description: item.package.description,
+          } as Fig.Suggestion)
+      ) as Fig.Suggestion[];
+    } catch (e) {
+      return [];
+    }
+
+    return temp;
```
QuiiBz, line 21: Let's remove the `var` and return directly the output of `map`:
```suggestion
      return JSON.parse(out).results.map(
        (item) =>
          ({
            name: item.package.name,
            description: item.package.description,
          } as Fig.Suggestion)
      ) as Fig.Suggestion[];
    } catch (e) {
      return [];
    }
```

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+      description:
+        "If true, pnpm doesn't generate a lockfile and fails to install if the lockfile is out of sync with the manifest / an update is needed or no lockfile is present.",
+    },
+    {
+      name: "--reporter",
+      description: `Allows you to choose the reporter that will log debug info to the terminal about the installation progress.
+      \n silent - no output is logged to the console, except fatal errors
+      \n default - the default reporter when the stdout is TTY
+      \n append-only (Added in v1.29.1) - the output is always appended to the end. No cursor manipulations are performed
+      \n ndjson - the most verbose reporter. Prints all logs in ndjson format`,
```
QuiiBz, line 151: Looks like this option takes an argument, and we can also provide suggestions for it! Here is the documentation for [args](https://fig.io/docs/reference/option#args) and [suggestions](https://fig.io/docs/reference/arg#suggestions)

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+        description:
+          "Using --save-peer will add one or more packages to peerDependencies and install them as dev dependencies",
+      },
+      {
+        name: ["--ignore-workspace-root-check", "-W#"],
+        description: `Adding a new dependency to the root workspace package fails, unless the --ignore-workspace-root-check or -W flag is used. 
+        For instance, pnpm add debug -W.`,
+      },
+      {
+        name: ["--global"],
```
QuiiBz, line 212: Not sure about this for `pnpm`, but doesn't this option has an `-g` alias?

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+  INSTALL_SUBCOMMAND,
+  {
+    name: ["update", "up"],
+    description: `pnpm update updates packages to their latest version based on the specified range.
+      \nWhen used without arguments, updates all dependencies. You can use patterns to update specific dependencies.`,
+    args: {
+      name: "Package",
+      isOptional: true,
+      generators: searchDependenciesGenerator,
+      debounce: true,
```
QuiiBz, line 231: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    ],
+  },
+  {
+    name: ["remove", "rm", "uninstall", "un"],
+    description: `Removes packages from node_modules and from the project's package.json.`,
+    args: {
+      name: "Package",
+      isOptional: true,
+      generators: searchDependenciesGenerator,
+      debounce: true,
```
QuiiBz, line 281: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+  },
+  {
+    name: ["link", "ln"],
+    description: `Makes the current local package accessible system-wide, or in another location.`,
+    args: [
+      {
+        name: "Package",
+        isOptional: true,
+        generators: searchDependenciesGenerator,
+        debounce: true,
```
QuiiBz, line 317: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    name: "unlink",
+    description: `Unlinks a system-wide package (inverse of pnpm link).
+      \nIf called without arguments, all linked dependencies will be unlinked.
+      \nThis is similar to yarn unlink, except pnpm re-installs the dependency after removing the external link.`,
+    args: [
+      {
+        name: "Package",
+        isOptional: true,
+        generators: searchDependenciesGenerator,
+        debounce: true,
```
QuiiBz, line 344: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+  },
+  {
+    name: ["rebuild", "rb"],
+    description: `Rebuild a package.`,
+    args: [
+      {
+        name: "Package",
+        isOptional: true,
+        generators: searchDependenciesGenerator,
+        debounce: true,
```
QuiiBz, line 370: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+  },
+  {
+    name: "prune",
+    description: `Removes unnecessary packages.`,
+    args: [
+      {
+        name: "Package",
+        isOptional: true,
+        generators: searchDependenciesGenerator,
+        debounce: true,
```
QuiiBz, line 391: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+];
+
+const SUBCOMMANDS_RUN_SCRIPTS: Fig.Subcommand[] = [
+  {
+    name: ["run", "run-script"],
+    description: "Runs a script defined in the package's manifest file.",
+    args: {
+      name: "Scripts",
+      generators: searchScriptsGenerator,
+      debounce: true,
```
QuiiBz, line 436: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    ],
+  },
+  {
+    name: "exec",
+    description: `Execute a shell command in scope of a project.
+      \nnode_modules/.bin is added to the PATH, so pnpm exec allows executing commands of dependencies.`,
+    args: {
+      name: "Scripts",
+      generators: searchDependenciesGenerator,
+      debounce: true,
```
QuiiBz, line 469: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+      },
+    ],
+  },
+  {
+    name: "why",
+    description: `Shows all packages that depend on the specified package.`,
+    args: {
+      name: "Scripts",
+      generators: searchDependenciesGenerator,
+      debounce: true,
```
QuiiBz, line 629: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+SUBCOMMANDS_MISC[1].subcommands = recursiveSubcommands;
+
+// SPEC
+export const completionSpec: Fig.Spec = {
+  name: "pnpm",
+  description: "Fast, disk space efficient package manager",
+  args: {
+    name: "Scripts",
+    generators: searchScriptsGenerator,
+    debounce: true,
```
QuiiBz, line 868: `debounce` is not needed here.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+};
+
+const SUBCOMMANDS_MANAGE_DEPENDENCIES: Fig.Subcommand[] = [
+  {
+    name: "add",
+    description: `Installs a package and any packages that it depends on. 
+      \n By default, any new package is installed as a production dependency.`,
+    args: {
+      name: "Package",
+      isOptional: true,
```
QuiiBz, line 174: Doesn't `pnpm add` always need at least one argument?

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    options: [
+      {
+        name: "--audit-level",
+        description: `Only print advisories with severity greater than or equal to <severity>.`,
+        args: [
+          { name: "low" },
+          { name: "moderate" },
+          { name: "high" },
+          { name: "critical" },
+        ],
```
QuiiBz, line 515: I think this should be `suggestions` inside `args`, not multiple args.

The expected behavior is: `pnpm audit --audit-level <low OR moderate OR high OR critical>`
But the current implementation behaves like this: `pnpm audit --audit-level <low> <moderate> <high> <critical>`

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    options: [
+      {
+        name: "--tag",
+        description: `Publishes the package with the given tag. By default, pnpm publish updates the latest tag.`,
+        args: [{ name: "<tag>" }],
+      },
+      {
+        name: "--access",
+        description: `Tells the registry whether the published package should be public or restricted.`,
+        args: [{ name: "<public|restricted>" }],
```
QuiiBz, line 682: We could also add suggestions here! Also, you can specify directly an `arg` object when there is a single value in the array. `args` allow both an arg object or an array of arg objects.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    subcommands: [
+      {
+        name: "start",
+        description:
+          "Starts a server that performs all interactions with the store. Other commands will delegate any store-related tasks to this server.",
+        options: [
+          {
+            name: "--background",
+            description: `Runs the server in the background, similar to daemonizing on UNIX systems.`,
+            args: [{ name: "bool" }],
```
QuiiBz, line 746: Does this option really take an argument? Usually, booleans options don't take arguments - if the option is specified, CLIs assume that it's `true`, and `false` else.

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+          },
+          {
+            name: "--network-concurrency",
+            description: `The maximum number of network requests to process simultaneously.`,
+            args: [{ name: "number" }],
+          },
+          {
+            name: "--protocol",
+            description: `The communication protocol used by the server. When this is set to auto, IPC is used on all systems except for Windows, which uses TCP.`,
+            args: [{ name: "<auto | tcp | ipc>" }],
```
QuiiBz, line 756: Would love to see suggestions here too!

QuiiBz (changes requested):
Thanks a lot for this PR, excited to support `pnpm`! I'm not very familiar with it, even if it looks quite similar to `npm` or `yarn`. I left some comments to fix before merging.

Feel free to ask if you need more information on specific options, args, etc

dev/pnpm.ts
mattschrage: Something like this would work @kevinfaveri

```
args: {
  name: "access",
  suggestions: [ "public", "restricted" ]
}
```

dev/pnpm.ts
```diff
@@ -0,0 +1,872 @@
...
+    } catch (e) {
+      return [];
+    }
+
+    return temp;
+  },
+};
+
+const searchDependenciesGenerator: Fig.Generator = {
+  script: () => `cat package.json`,
```
Monkeyank, line 26: I think it'd be nice to check if there's a package.json in the enclosing directory.

For reference: https://github.com/withfig/autocomplete/blob/05f52678515324618251aa41139d203692aedb19/dev/yarn.ts#L28-L31

dev/pnpm.ts
kevinfaveri: It does accept it, but it is not documented https://pnpm.io/cli/add#--global

kevinfaveri:
Another suggestion while I'm reviewing the PR comments: have suggestions which is just a type, like:
```typescript
--port

args: [
name: 'Port',
suggestion: [
'auto',
ANY_NUMBER
]
]
```

I can think of ANY_NUMBER, ANY_ALPHANUMERIC, ANY_STRING, ANY_BOOLEAN. If the user typed anything other than what is specified in the suggestion, it would trigger a warning like 'suggestion not supported'

dev/pnpm.ts
kevinfaveri: Yeah. Since it is only bool, we can hide the args here.

kevinfaveri:
It would be interesting to also have the ability to create new templates so we can reuse generators

kevinfaveri:
Okay, guys, I did some reorganizing of the code and tackled the review points. 
Everything is working as expected now.

Support for --publish branch arg has been added.

Besides what I suggested above for improving DX, I would:
- Have a 'shared' or 'utils' folder in the dev folder for reusing generators, utility functions, options, subcommands, anything

QuiiBz:
> Besides what I suggested above for improving DX, I would:
> 
> * Have a 'shared' or 'utils' folder in the dev folder for reusing generators, utility functions, options, subcommands, anything

We are going to implement a very similar feature soon, you will be able to import and export anything across any spec!

dev/pnpm.ts
```diff
@@ -0,0 +1,993 @@
...
+    return [];
+  },
+};
+
+// SUBCOMMANDS
+const SUBCOMMANDS_MANAGE_DEPENDENCIES: Fig.Subcommand[] = [
+  {
+    name: "add",
+    description: `Installs a package and any packages that it depends on. 
+      \n By default, any new package is installed as a production dependency.`,
```
QuiiBz, line 143: You can remove every `\n` from the descriptions, since we render them inline

kevinfaveri:
That's cool to hear, @QuiiBz! Btw I've updated the branch

QuiiBz (approved):

Issue #348, opened 2021-07-07
[npx] Add support for npx completion spec


brendanfalk:
Create a completion spec for npx, specifically for local packages

QuiiBz:
@brendanfalk could you clarify what "specifically for local packages" means? The current `npx` spec have completion for cli such as prisma, react-native, etc. Those suggestions are always shown since you don't need to have those packages installed in your cwd to run them.

Would you mean that packages for cli installed in the current cwd should have a higher priority?

QuiiBz:
Merged in #359!

PR #359, opened 2021-07-13
[npx] Local Package Support

Monkeyank:
Feature
#348


dev/npx.ts
```diff
@@ -1,6 +1,34 @@
 export const completionSpec: Fig.Spec = {
   name: "npx",
   description: "Execute binaries from npm packages",
+  generateSpec: async (_context, executeShellCommand) => {
+    const commands = (
+      await executeShellCommand(
+        `until [[ -d node_modules/ ]] || [[ $PWD = '/' ]]; do cd ..; done; ls -1 node_modules/.bin/`
```
kevinfaveri, line 7: Alternative solution for getting the dependencies...

https://github.com/withfig/autocomplete/pull/357/files#diff-bf5ffaedcd1f66b9d30632707ab23d8b88f80e98c0cd6e797a3b9929c5f20f6fR25

Monkeyank: Good catch, thanks!

Monkeyank: In this case I'm not getting dependencies, but the executables from the modules.

Monkeyank:
Do we want to get rid of subcommand suggestions that we can't load a spec for? 
Ex: `acorn` (which isn't a cli).

QuiiBz:
Hi, not sure if we really need this. Instead of `yarn <cli>`, `npx <cli>` can be used without any `package.json`, so it will broke this feature.

Monkeyank:
> Hi, not sure if we really need this. Instead of `yarn <cli>`, `npx <cli>` can be used without any `package.json`, so it will broke this feature.

I thought that the issue referenced in #348, wanted suggestions based on packages installed locally for `npx`. Obviously this doesn't cover everything, because `npx` can also run packages not installed locally.

QuiiBz:
`npx` already has default subcommands (see https://github.com/withfig/autocomplete/blob/master/dev/npx.ts#L4-L31), so I believe that this feature doesn't change the suggestions.

dev/npx.ts
```diff
@@ -1,34 +1,58 @@
...
-    },
-  ],
+
+  args: {
+    name: "command",
+    isCommand: true,
+    generators: {
+      script: `until [[ -d node_modules/ ]] || [[ $PWD = '/' ]]; do cd ..; done; ls -1 node_modules/.bin/`,
+      postProcess: function (out) {
+        out.split("\n");
```
QuiiBz, line 34: Useless split here

dev/npx.ts
```diff
@@ -1,34 +1,58 @@
...
+
+        const cli = [
+          "react-native",
+          "tailwindcss",
+          "next",
+          "gltfjsx",
+          "prisma",
+        ];
```
QuiiBz, line 42: We can avoid duplicating those specs with something like this:
```suggestion
        const cli = [...suggestions].reduce((acc, { name }) => [...acc, name], []);
```

QuiiBz (changes requested):

QuiiBz:
Could you fix the conflict so we can get this PR merged in? Thanks!

.github/ISSUE_TEMPLATE/main_issue_template.yml
```diff
@@ -3,36 +3,36 @@ title: "[CLI_NAME]"
 description: All issues relating to specific CLI tools (e.g. `git`, `kubectl`, `chmod`)
```
QuiiBz, line 3: Could you remove those modifications in this file? It will be done in #388. After that, we will be ready to merge!

QuiiBz (approved):
LGTM!

dev/npx.ts
```diff
@@ -1,50 +1,68 @@
...
+  args: {
+    name: "command",
+    isCommand: true,
+    generators: {
+      script: `until [[ -d node_modules/ ]] || [[ $PWD = '/' ]]; do cd ..; done; ls -1 node_modules/.bin/`,
+      postProcess: function (out) {
+        const cli = [...suggestions].reduce(
+          (acc, { name }) => [...acc, name],
+          []
+        );
```
fedeci, line 51: Why not an hash Map? `.has` in constant time complexity while `.includes` is linear.

QuiiBz: I don't really understand why you want to use a `Set` here. We just need to grab the `name` property of each `suggestions` in an array.

fedeci: Consider that the complexity of `Array.prototype.includes` is `O(n)`. We are running it `n` times in `filter` so we can estimate about a quadratic time complexity.
Furthermore `[...suggestions]` is an unused copy of all the items in a new array. We should only run a map on top of the existing suggestions. Both get fixed in #394

QuiiBz: I agree that `cli` can and should be moved to a top-level const instead of being calculated in each `postProcess` call. However, the gain in terms of performances is ridiculous, and I think we should really not focus on over-optimizing stuff, but rather covering as many CLIs as possible (while of course not forgetting to optimize generators)

PR #362, opened 2021-07-13
feat: add eslint completion spec

fwesss:
Feature

No autocomplete for eslint

Autocompletion added for eslint

**Additional info:**


dev/eslint.ts
```diff
@@ -0,0 +1,245 @@
...
+    },
+    {
+      name: "--stdin",
+      description: "Lint code provided on <STDIN>",
+    },
+    {
+      name: "--stdin-filename",
+      description: "Specify filename to process STDIN as",
+      args: {
+        template: ["folders"],
```
Monkeyank, line 130: Should this be `filepaths`?

dev/eslint.ts
```diff
@@ -0,0 +1,245 @@
...
+        "Use this configuration, overriding .eslintrc.* config options if present",
+      args: {
+        template: ["filepaths"],
+      },
+    },
+    {
+      name: "--env",
+      description: "Specify environments",
+      args: {
+        default: ".js",
```
QuiiBz, line 23: Looks like the environments are defined here: https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments

dev/eslint.ts
```diff
@@ -0,0 +1,245 @@
...
+      },
+    },
+    {
+      name: "--ext",
+      description: "Specify JavaScript file extensions",
+      args: {
+        default: ".js",
```
QuiiBz, line 30: Could you add a `name` property (something like `"extension"`)

dev/eslint.ts
```diff
@@ -0,0 +1,245 @@
...
+      },
+    },
+    {
+      name: "--color",
+      description: "Force enabling of color",
+    },
+    {
+      name: "--no-color",
+      description: "Force disabling of color",
+    },
```
QuiiBz, line 179: We can use the [exclusive](https://fig.io/docs/reference/option#exclusive) property for those options

QuiiBz (changes requested):
Thanks a lot for this PR, `eslint` is really useful!

Could you also add `eslint` here: https://github.com/withfig/autocomplete/blob/05f52678515324618251aa41139d203692aedb19/dev/yarn.ts#L130 and here: https://github.com/withfig/autocomplete/blob/05f52678515324618251aa41139d203692aedb19/dev/npx.ts#L4

This will allow autocompletion with `yarn eslint` and `npx eslint`!

dev/eslint.ts
fwesss: I've added the envs as static suggestions. The user can also add multiple envs separated by a comma as indicated in the eslint docs.

dev/eslint.ts
fwesss: Name property added

dev/eslint.ts
fwesss: That's correct! I've pushed the change.

dev/eslint.ts
```diff
@@ -0,0 +1,298 @@
...
+                "greasemonkey",
+              ];
+
+              const currentToken = [ctx[ctx.length - 1]][0];
+
+              // Return suggestions minus those that have already been entered
+              return suggestions
+                .filter((sugg) => !currentToken.split(",").includes(sugg))
+                .map((sugg) => ({ name: sugg }));
+            },
```
QuiiBz, line 71: That's smart, but we can avoid running a dummy script using `custom`. This property is a function that also takes the context, and should also return a list of suggestions. Here is an example implementation:
```suggestion
            /*
             * Multiple envs can be set with a ',' seperator.
             * We want to filter over this static list of suggestions
             * so we don't suggest envs that have already been entered.
             * */
            filterTerm: ",",
            trigger: ",",
            custom: async (ctx) => {
              const suggestions = [
                "browser",
                "node",
                "commonjs",
                "shared-node-browser",
                "es6",
                "ecmaVersion",
                "es2017",
                "es2020",
                "es2021",
                "worker",
                "amd",
                "mocha",
                "jasmine",
                "jest",
                "phantomjs",
                "protractor",
                "qunit",
                "jquery",
                "prototypejs",
                "shelljs",
                "meteor",
                "mongo",
                "applescript",
                "nashorn",
                "servicerworker",
                "atomtest",
                "embertest",
                "webextensions",
                "greasemonkey",
              ];

              const currentToken = [ctx[ctx.length - 1]][0];

              // Return suggestions minus those that have already been entered
              return suggestions
                .filter((sugg) => !currentToken.split(",").includes(sugg))
                .map((sugg) => ({ name: sugg }));
            },
```

QuiiBz (changes requested):

dev/eslint.ts
fwesss: Thanks for the suggestion! That's perfect since I was just looking for a way to get the context but didn't care about any actual output. I saw custom in the docs but didn't think to apply it here so that's a helpful example.

dev/eslint.ts
```diff
@@ -0,0 +1,297 @@
...
+      args: {},
+    },
+    {
+      name: "--parser-options",
+      description: "Specify parser options",
+      args: {},
+    },
+    {
+      name: "--resolve-plugins-relative-to",
+      description: "A folder where plugis should be resolved from",
```
Monkeyank, line 102: Typo: `plugins`

dev/eslint.ts
```diff
@@ -0,0 +1,297 @@
...
+      },
+    },
+    {
+      name: "--no-ignore",
+      description: "Disable use of ignore files and patterns",
+    },
+    {
+      name: "--ignore-pattern",
+      description:
+        "Pattern of files to ignore (in addition to those in .eslintignore",
```
Monkeyank, line 169:
```suggestion
        "Pattern of files to ignore (in addition to those in .eslintignore)",
```

dev/eslint.ts
Monkeyank, line 102:
```suggestion
      description: "A folder where plugins should be resolved from",
```

Monkeyank (changes requested):

dev/eslint.ts
```diff
@@ -0,0 +1,297 @@
...
+        template: ["filepaths"],
+      },
+    },
+    {
+      name: "--env",
+      description: "Specify environments",
+      args: {
+        generators: [
+          {
+            // dummy script to get postProcess to execute
```
QuiiBz, line 23: You can safely remove this comment since we now use `custom`. After this, I think we will be ready to merge!

fwesss: Just removed the comment. Thanks for being so patient and making sure this was a clean PR. Next one will be much better!

QuiiBz (approved):
LGTM! Thanks a lot for this amazing PR @fwesss, this is going to be really useful!

PR #363, opened 2021-07-13
[Prettier] Add Spec

Monkeyank:
Feature
Adds the `Prettier` Spec

### DIFF
--- a/dev/npx.ts
+++ b/dev/npx.ts
@@ -28,6 +28,11 @@ export const completionSpec: Fig.Spec = {
         "https://raw.githubusercontent.com/prisma/docs/main/src/images/favicon-16x16.png",
       loadSpec: "prisma",
     },
+    {
+      name: "prettier",
+      icon: "https://prettier.io/icon.png",
+      loadSpec: "prettier",
+    },
   ],
   options: [
     {
--- a/dev/prettier.ts
+++ b/dev/prettier.ts
@@ -1,0 +1,333 @@
+export const completion: Fig.Spec = {
+  name: "prettier",
+  description: "Run Prettier from the command line",
+  icon: "https://prettier.io/icon.png",
+  options: [
+    {
+      name: ["-c", "--check"],
+      description: "Check if your files are formatted",
+      args: {
+        name: "file, dir, or glob",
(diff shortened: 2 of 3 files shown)


dev/npx.ts
```diff
@@ -28,6 +28,11 @@ export const completionSpec: Fig.Spec = {
         "https://raw.githubusercontent.com/prisma/docs/main/src/images/favicon-16x16.png",
       loadSpec: "prisma",
     },
+    {
+      name: "prettier",
+      icon: "https://prettier.io/icon.png",
+      loadSpec: "prettier",
+    },
```
QuiiBz, line 35: Could you also add `pretter` to `yarn` here? https://github.com/withfig/autocomplete/blob/bf83070f73b690d3f4ede32feda59a6bb3b9c60d/dev/yarn.ts#L130

dev/prettier.ts
```diff
@@ -0,0 +1,333 @@
...
+        template: "filepaths",
+        default: ".prettierignore",
+      },
+    },
+    {
+      name: "--plugin",
+      description: "Add a plugin",
+      args: {
+        name: "path",
+        template: "filepaths",
```
QuiiBz, line 216: Based on prettier's example:
```bash
prettier --write main.foo --plugin-search-dir=./dir-with-plugins --plugin=./foo-plugin
```
I'm not sure if the path is a file or a folder - looks like it's a folder but that's strange.

dev/prettier.ts
```diff
@@ -0,0 +1,333 @@
...
+    {
+      name: "--no-error-on-unmatched-pattern",
+      description: "Prevent errors when pattern is unmatched",
+    },
+  ],
+  args: {
+    name: "file, dir or glob",
+    isOptional: true,
+    variadic: true,
+  },
```
QuiiBz, line 332: We should add `template: ["filepaths", "folders"]` here

QuiiBz (changes requested):
Only a few things to fix, I don't know how we didn't yet have a spec for `prettier`!

QuiiBz (approved):
LGTM!

PR #366, opened 2021-07-13
Add composer autocompletion spec

zeo:
This PR introduces a new autocompletion spec, Composer. Composer is a dependency manager for PHP, you can find more information about it [here](https://getcomposer.org/).

There currently is no behavior, as the spec did not exist yet.

It adds autocompletion for composer subcommands.

**Additional info:**


dev/php/composer.ts
```diff
@@ -0,0 +1,87 @@
...
+              default: argDefault,
+              variadic: arg.is_array
+            };
+          }),
+
+          options: Object.keys(command.definition.options).map(optionKey => {
+            const option = command.definition.options[optionKey];
+
+            return {
+              name: [option.name, option.shortcut],
```
QuiiBz, line 69: Is there always a `shortcut` property defined on each option? This might render `undefined` into Fig if it's not the case.

dev/php/composer.ts
```diff
@@ -0,0 +1,87 @@
...
+              description: option.description,
+              required: option.is_value_required,
+              args: option.accept_value ? {} : undefined,
+            };
+          }),
+        });
+      }
+    } catch (err) {
+      //
```
QuiiBz, line 78: Could you add a `console.error(err)` here?

QuiiBz (changes requested):
Whoa, that's really handy to have all the subcommands, options and args generated this way! Thanks a lot for this PR. I left some comments on things to update.

QuiiBz:
Could you also run the lint script to make sure everything is formatted the good way?

dev/php/composer.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    return {
+      name: 'composer',
+      debounce: true,
```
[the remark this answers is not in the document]
zeo: I added it since the existing laravel spec, figured Fig would perform better this way. I'll remove this as it's not the case.

dev/php/composer.ts
zeo: After some further looking, it seems that when the shortcut is missing, composer just providers an empty string. I'll modify the code to remove the shortcut if it's empty.

dev/php/composer.ts
zeo: Yep sure!

QuiiBz (approved):
LGTM! Thank you a lot, this spec is going to be really useful for PHP developers!

PR #367, opened 2021-07-14
[uniq] Add Spec

Monkeyank:
Feature
Adds `uniq` spec

**Additional info:**

### DIFF
--- a/dev/uniq.ts
+++ b/dev/uniq.ts
@@ -67,6 +67,7 @@ export const completion: Fig.Spec = {
     {
       name: "input",
       isOptional: true,
+      template: ["filepaths", "folders"],
     },
     {
       name: "output",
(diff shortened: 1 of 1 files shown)


dev/uniq.ts
```diff
@@ -0,0 +1,76 @@
...
+  ],
+  args: [
+    {
+      name: "input",
+      isOptional: true,
+    },
+    {
+      name: "output",
+      isOptional: true,
+    },
```
QuiiBz, line 74: Doesn't those args should have a template of `filepaths`?

Monkeyank: For `input` it should have `filepaths` and `folders`. For `output` there shouldn't be any templates.

QuiiBz (approved):
LGTM!

Issue #321, opened 2021-06-22
[aws] Suggest when a user does not have valid credentials


brendanfalk:
If a user downloads the AWS CLI but has not set up their credentials, it would be great if Fig could suggest "hit enter to set up your AWS credentials"

This would rely on a `generateSpec` function which would check if credentials were set up if, and if not, append a suggestion to the top of the suggestion list with `priority: 100` suggesting how to set up your credentials!

PR #368, opened 2021-07-14
[aws] [configure] prioritize when credentials aren't set up 

Monkeyank:
Enhancement

#321 
Prioritizes the `configure` subcommand, when `~/.aws/credentials` or `~/.aws/config` don't exist (user doesn't have credentials set up).


Monkeyank:
Any suggestions to make this more robust?

Do you think that `aws sts get-caller-identity` might be a better approach in case a user's config/credential files aren't in the default location?

When we use this command, we can check if the user has an account and increase priority based on that.

https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

QuiiBz:
I think `aws sts get-caller-identity` is going to be significantly slower than a `ls`. Is there another way to find the location of the config file?

dev/aws.ts
```diff
@@ -1,5 +1,23 @@
...
+          loadSpec: "aws/configure",
+          priority: prioritize && 100,
```
QuiiBz, line 16: I think it can be cool to have an `alert` icon to visually see that we didn't configured aws yet.

Here is a preview of it:
Screenshot 2021-07-16 at 16 32 05

QuiiBz (changes requested):

Monkeyank:
> I think `aws sts get-caller-identity` is going to be significantly slower than a `ls`. Is there another way to find the location of the config file?

I tried `aws configure list-profiles`. Although this is faster than `aws sts get-caller-identity` it still takes a significant amount of time to load. If the config/credentials file is not in the default location, then you have to use an `aws` command to find the location.

QuiiBz:
Alright. Let's keep `ls` for this. Could you add an alert icon as shown in my review above?

QuiiBz (approved):
LGTM!

PR #370, opened 2021-07-14
[HUGO] feat: Add spec for Hugo v0.85.0

danpoynor:
feature

No autocomplete spec for Hugo CLI commands

Adds a comprehensive autocompletion spec for Hugo CLI commands

**Additional info:**
References:
- [Hugo CLI documentation on gohugo.io](https://gohugo.io/commands/)
- [Hugo hugo/tree/master/commands/ source code on github](https://github.com/gohugoio/hugo/tree/master/commands)
- `hugo help` output


dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+    name: ["-h", "--help"],
+    description: `help for ${name}`,
+    insertValue: "--help",
+  };
+  return help;
+};
+
+// global options
+
+const config = {
```
QuiiBz, line 14: Would love to see a type of `Fig.Option` here! Same for option objects each underneath

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+  name: "--config",
+  description: "config file (default is path/config.yaml|json|toml)",
+  insertValue: "--config {cursor}",
```
QuiiBz, line 17: No need to specify `insertValue` here, Fig will handle this behaviour automatically. Same for option objects each underneath

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+
+const themesDir = {
+  name: "--themesDir",
+  description: "filesystem path to themes directory",
+  insertValue: '--themesDir="{cursor}"',
+  priority: 50,
+  args: {
+    name: "file",
+    description: "default is path/config.yaml|json|toml",
+    template: "filepaths",
```
QuiiBz, line 145: I think the template here should be `folders` , not `filepaths`

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+  args: {
+    name: "boolean",
+    suggestions: [
+      { name: "false", icon: "❌", description: "default" },
+      { name: "true", icon: "✅" },
+    ],
+  },
+};
+
+const globalOptions = [
```
QuiiBz, line 176: Could you add a type of `Fig.Option[]` here?

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+  verbose,
+  verboseLog,
+];
+
+// options common to 'hugo', 'hugo mod', 'hugo new', and 'hugo server' commands
+
+const baseURL = {
+  name: ["-b", "--baseURL"],
+  description: "hostname (and path) to the root, e.g. http://spf13.com/",
+  insertValue: '--baseURL="{cursor}"',
```
QuiiBz, line 197: Same as above, no need for `insertValue`, and would love to see a `Fig.Option` type! Same for other objects underneath

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+  args: {
+    name: "hostname and path",
```
QuiiBz, line 199: Does this option takes two arguments, hostname and path? Or is it a single argument?

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+  },
+};
+
+const disableKinds = {
+  name: "--disableKinds",
+  description: "disable different kind of pages (home, RSS etc.)",
+  insertValue: '--disableKinds="{cursor}"',
+  args: {
+    name: "kind,kind",
+    suggestions: [
```
QuiiBz, line 293: I think this arg can be `variadic` (see https://fig.io/docs/reference/arg#variadic)

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+const debug = {
+  name: "--debug",
+  description: "debug output (default false)",
+  insertValue: "--debug={cursor}",
+  priority: 50,
+  args: {
+    name: "boolean",
+    suggestions: [
+      { name: "false", icon: "❌", description: "default" },
+      { name: "true", icon: "✅" },
```
QuiiBz, line 47: This is kind of strange. Usually, those options doesn't takes an argument at all. If the option is present, the cli assume that its value is true, and false else. I'm not very familiar with Hugo, so is this the expected behaviour?

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+        {
+          name: ["--invalidateCDN"],
+          description:
+            "invalidate the CDN cache listed in the deployment target (default true)",
+        },
+        {
+          name: ["--maxDeletes"],
+          description:
+            "maximum # of files to delete, or -1 to disable (default 256)",
+          insertValue: "--maxDeletes {cursor}",
```
QuiiBz, line 654: No need for `insertValue` here, same for other options underneath

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+            },
+            {
+              name: ["-t", "--type"],
+              description:
+                "autocompletion type (bash, zsh, fish, or powershell) (default 'bash')",
+              insertValue: "--type {cursor}",
+              args: {
+                name: "type",
+                suggestions: [
+                  { name: "bash", description: "default" },
```
QuiiBz, line 712: We can add a `default` property of `bash` in this arg!

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+            },
+            {
+              name: ["--style"],
+              description:
+                "highlighter style (see https://help.farbox.com/pygments.html) (default 'friendly')",
+              insertValue: "--style {cursor}",
+              args: {
+                name: "style name",
+                suggestions: [
+                  { name: "friendly", description: "default" },
```
QuiiBz, line 753: Same as above

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+          description:
+            "Import from Jekyll requires two paths, e.g. ` + `hugo import jekyll jekyll_root_path target_path`.",
+          insertValue: "jekyll {cursor}",
+          args: [
+            {
+              name: "jekyll_root_path",
+            },
+            {
+              name: "target_path",
+            },
```
QuiiBz, line 882: Can those args have a template of `folders` or `filepaths`?

dev/hugo.ts
```diff
@@ -0,0 +1,1277 @@
...
+              name: "--force",
+              description: "init inside non-empty directory",
+              isDangerous: true,
+            },
+            {
+              name: ["-f", "--format"],
+              description: 'config & frontmatter format (default "toml")',
+              insertValue: "--format '{cursor}'",
+              args: {
+                name: "format",
```
QuiiBz, line 1069: Could we have suggestions for the format option?

QuiiBz (changes requested):
Thanks a lot for this huge spec! I'm not familiar with Hugo so some points in my review might be wrong.

You can also remove every `insertValue` properties since they are using the default behaviour of Fig.

dev/hugo.ts
danpoynor: @QuiiBz When I add the `default` property I don't see anything indicating that bash is the default value. I'm not sure what I'm suppose to see when using that property though.

Using the description property to literally state what the default is might be more user friendly.

dev/hugo.ts
danpoynor: @QuiiBz If I remove the `insertValue` property, users are required to type a `space` before the `template` property is invoked - which doesn't seem as automated as could be.

If I leave the `insertValue` property, users are not required to enter a `space` and the `template` is automatically invoked - which seems more like a more automated behavior.

I'm not sure if removing `insertValue` (so the ux is less automated) is a Fig standard or a preference.

dev/hugo.ts
```diff
@@ -0,0 +1,1207 @@
+import { stdout } from "node:process";
```
mattschrage, line 1: Seems like this was imported accidentally?

dev/hugo.ts
QuiiBz: All other specs don't use `insertValue` for this behaviour. I don't think that's an issue to press space - and not adding a space also make it easy to come back and edit the inserted value in case of a mistake. cc @mattschrage

dev/hugo.ts
danpoynor: @QuiiBz `default: "bash"` added. I'm not sure if it's doing what it's suppose to do though (not sure how to test this).

danpoynor:
@QuiiBz Thanks for the thorough review and feedback!

I think all the issues are resolved if you could take another look.

dev/hugo.ts
mschrage: If the argument that the `--config` flag takes is not optional, Fig will automatically insert a space for the user.

dev/hugo.ts
```diff
@@ -0,0 +1,1204 @@
...
+  {
+    name: "--debug",
+    description: "debug output (default false)",
+    insertValue: "--debug=",
+    priority: 50,
+    args: {
+      name: "boolean",
+      suggestions: [
+        { name: "false", icon: "❌", description: "default" },
+        { name: "true", icon: "✅" },
```
QuiiBz, line 38: I'm still wondering about this. Usually, those options don't take an argument at all. If the option is present, the cli assumes that its value is true, and false else. I'm not very familiar with Hugo, so is this the expected behavior?

dev/hugo.ts
```diff
@@ -0,0 +1,1204 @@
+const help = (name: string): Fig.Option => ({
+  name: ["-h", "--help"],
+  description: `help for ${name}`,
+});
+
+const globalOptions: Fig.Option[] = [
+  {
+    name: "--config",
+    description: "config file (default is path/config.yaml|json|toml)",
+    insertValue: "--config=",
```
QuiiBz, line 10: You can safely remove every `insertValue` in this spec, as mentioned in the previous review.

QuiiBz (changes requested):

QuiiBz (approved):
LGTM!

PR #373, opened 2021-07-14
[jq] Add Minimal Spec

Monkeyank:
Feature
Adds `jq`, but doesn't actually provide suggestions for the filters that users can use. Only includes options and args that it takes.
**Additional info:**
https://code.tools/man/1/jq/


QuiiBz (approved):
LGTM!

Issue #376, opened 2021-07-14
[git] Prioritize unstaged files for `git add`


mschrage:
<img width="807" alt="Screen Shot 2021-07-14 at 4 21 02 PM" src="https://user-images.githubusercontent.com/3155109/125705264-6b20de70-70cf-4594-8261-7e33129b2f8d.png">

For example, `tools/drip/prompt_to_update.sh` should be the top suggestion. 

Another solution: files that have already been staged could be removed from the list of suggestions?

dereklowlind:
@mschrage  please check out the PR I did for this!

QuiiBz:
Merged in #380, thanks everyone!

PR #380, opened 2021-07-15
Git prioritize unstaged files

dereklowlind:
This adds the ability to prioritize unstaged files in the suggestions for git add. This feature was suggested in https://github.com/withfig/autocomplete/issues/376

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -171,15 +171,12 @@ const gitGenerators: Record<string, Fig.Generator> = {
       }
 
       const files = output.split("\n").map((file) => {
-        let alreadyAdded = false;
         // From "git status --short"
         // M  dev/github.ts // test file that was added
         //  M dev/kubectl.ts // test file that was not added
         // A  test2.txt // new added and tracked file
         // ?? test.txt // new untracked file
-        if (file.charAt(0) === "M" || file.charAt(0) === "A") {
-          alreadyAdded = true;
-        }
+        const alreadyAdded = file.charAt(0) === "M" || file.charAt(0) === "A";
 
         file = file.trim();
         const arr = file.split(" ");
@@ -236,12 +233,11 @@ const gitGenerators: Record<string, Fig.Generator> = {
             ext = "folder";
           }
 
-          let priority = 100;
           // If the current file already is already added
           // we want to lower the priority
-          if (item.alreadyAdded || context.some((ctx) => ctx.includes(file))) {
-            priority = 50;
-          }
+          const priority = item.alreadyAdded || context.some(ctx => ctx.includes(file))
+            ? 50
+            : 100;
           return {
             name: file,
             icon: `fig://icon?type=${ext}&color=ff0000&badge=${item.working}`,
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -235,9 +235,10 @@ const gitGenerators: Record<string, Fig.Generator> = {
 
           // If the current file is already added
           // we want to lower the priority
-          const priority = item.alreadyAdded || context.some(ctx => ctx.includes(file))
-            ? 50
-            : 100;
+          const priority =
+            item.alreadyAdded || context.some((ctx) => ctx.includes(file))
+              ? 50
+              : 100;
           return {
             name: file,
             icon: `fig://icon?type=${ext}&color=ff0000&badge=${item.working}`,
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -176,7 +176,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
         //  M dev/kubectl.ts // test file that was not added
         // A  test2.txt // new added and tracked file
         // ?? test.txt // new untracked file
-        const alreadyAdded = file.charAt(0) === "M" || file.charAt(0) === "A";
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -171,10 +171,24 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
       const files = output.split("\n").map((file) => {
+        let alreadyAdded = false;
+        // From "git status --short"
+        // M  dev/github.ts // test file that was added
+        //  M dev/kubectl.ts // test file that was not added
+        // A  test2.txt // new added and tracked file
+        // ?? test.txt // new untracked file
+        if (file.charAt(0) === "M" || file.charAt(0) === "A") {
+          alreadyAdded = true;
+        }
```
QuiiBz, line 182: We can simplify since by assigning directly `alreadyAdded`:
```suggestion
        // From "git status --short"
        // M  dev/github.ts // test file that was added
        //  M dev/kubectl.ts // test file that was not added
        // A  test2.txt // new added and tracked file
        // ?? test.txt // new untracked file
        const alreadyAdded = file.charAt(0) === "M" || file.charAt(0) === "A";
```

dev/git.ts
```diff
@@ -222,13 +236,17 @@ const gitGenerators: Record<string, Fig.Generator> = {
             ext = "folder";
           }
 
+          let priority = 100;
+          // If the current file already is already added
+          // we want to lower the priority
+          if (item.alreadyAdded || context.some((ctx) => ctx.includes(file))) {
+            priority = 50;
+          }
```
QuiiBz, line 244: Same as above:
```suggestion
          // If the current file already is already added
          // we want to lower the priority
          const priority = item.alreadyAdded || context.some(ctx => ctx.includes(file))
            ? 50
            : 100;
```

QuiiBz (changes requested):
Cool change! Left some simplifications changes

dev/git.ts
```diff
@@ -222,13 +236,17 @@ const gitGenerators: Record<string, Fig.Generator> = {
             ext = "folder";
           }
 
+          let priority = 100;
+          // If the current file already is already added
```
fedeci, line 240: typo:
```suggestion
          // If the current file is already added
```

dev/git.ts
```diff
@@ -171,10 +171,24 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+        }
+
         file = file.trim();
         const arr = file.split(" ");
 
-        return { working: arr[0], file: arr.slice(1).join(" ").trim() };
+        return {
+          working: arr[0],
+          file: arr.slice(1).join(" ").trim(),
+          alreadyAdded: alreadyAdded,
```
fedeci, line 190: nitpick:
```suggestion
          alreadyAdded,
```

dev/git.ts
```diff
@@ -222,13 +236,17 @@ const gitGenerators: Record<string, Fig.Generator> = {
...
+            priority = 50;
+          }
           return {
             name: file,
             icon: `fig://icon?type=${ext}&color=ff0000&badge=${item.working}`,
             description: "Changed file",
-            // If the current file already is already added
-            // we want to lower the priority
-            priority: context.some((ctx) => ctx.includes(file)) ? 50 : 100,
+            priority: priority,
```
fedeci, line 249: nit:
```suggestion
            priority,
```

fedeci (approved):

Monkeyank:
While we're doing this can we also change `git status --short` to `git status --porcelain`?

Screen Shot 2021-07-16 at 11 08 42 AM

dereklowlind:
@Monkeyank yes the --porcelain change can totally be made! The output is very similar but the file paths displayed are different
image

Is there a way to change the fig suggestion window size? This is only a problem for repos like this bazel tutorial one with a lot of sub directories but it leads to not being able to read the file name.
-- with git status --short
image
-- with git status --porcelain
image

Monkeyank:
@dereklowlind 
On second thought I think the way the paths are displayed are better in `--short`. I think we can separate this into a different PR if we'd like to to do this in the future.

Also you can change your own Fig window size through the settings!

dereklowlind:
@QuiiBz I merged in your suggested changes. Do you have to rereview them for the pull request to get merged?

dev/git.ts
```diff
@@ -171,10 +171,21 @@ const gitGenerators: Record<string, Fig.Generator> = {
       }
 
       const files = output.split("\n").map((file) => {
+        // From "git status --short"
+        // M  dev/github.ts // test file that was added
+        //  M dev/kubectl.ts // test file that was not added
+        // A  test2.txt // new added and tracked file
+        // ?? test.txt // new untracked file
+        const alreadyAdded = file.charAt(0) === "M" || file.charAt(0) === "A";
```
QuiiBz, line 179: This might be a bit overkill but we can simplify it with:
```suggestion
        const alreadyAdded = ["M", "A"].includes(file.charAt(0));
```

QuiiBz (changes requested):

dev/git.ts
dereklowlind: You're making Linus Torvalds proud! 😄

QuiiBz (approved):
LGTM! Thanks a lot for this fix!

PR #387, opened 2021-07-17
Feat/pnpx support

kevinfaveri:
This PR adds support for pnpx CLI: https://pnpm.io/pnpx-cli

Fig doesn't support autocomplete for the pnpx CLI: https://pnpm.io

Autocomplete file pnpx.ts added to the repo, introducing support for autocomplete with the pnpx cli: https://pnpm.io/pnpx-cli

**Additional info:**
It uses the same API as npx according to their docs:  https://pnpm.io/pnpx-cli


QuiiBz:
> It uses the same API as npx according to their docs: https://pnpm.io/pnpx-cli

That is good news, we can use `loadSpec` to load the `npx` spec instead of duplicating the code. However, looks like there is an issue with `loadSpec` at the spec level. Opened an issue for this here: https://github.com/withfig/fig/issues/281

QuiiBz:
Update: I'm gonna merge this PR but we will refactor this to use the import/export syntax when #388 will be merged!

QuiiBz (approved):

PR #388, opened 2021-07-19
[INTERNAL]: use esbuild and esm syntax (import/export)

cstrnt:
Feature

Right now we use a custom compiler in order to work with fig.

We use the idiomatic way of doing import / export in JavaScript which allow to share functionality between files. We also now use esbuild which makes the developer experience much better and faster!

`npm run build` now only takes 800ms (before ~17secs)
which means 2100% faster!!

**Additional info:**


package.json
```diff
@@ -80,5 +81,6 @@
     "progress": "^2.0.3",
     "ts-node": "^9.1.1",
     "typescript": "^4.2.3"
-  }
+  },
+  "dependencies": {}
```
QuiiBz, line 85: We can remove this

package.json
```diff
@@ -9,7 +9,7 @@
     "copy:all": "rsync -a --include '*/' --include '*.js' --exclude '*' $(pwd)/specs/ ~/.fig/autocomplete/",
     "create-boilerplate": "./scripts/create-boilerplate.sh",
     "create-example": "./scripts/create-example.sh",
-    "build": "ts-node-script scripts/compiler.ts",
+    "build": "esbuild dev/*.ts dev/**/*.ts --outdir=specs --minify --format=esm --bundle",
```
QuiiBz, line 12: Not sure about this but I think this will only work if your shell expands glob paths. It may be interesting to run `compiler.ts` instead, and get the TS files to build from there.

The compiler will also need an option `watch` to watch or not the files. This would also avoid duplicating build options (outdir, minify, format, etc)

scripts/compiler.ts
```diff
@@ -40,96 +26,26 @@ function invalidateCache() {
...
-if (process.argv[2] == "INVALIDATE_CACHE") {
+async function processFile(fileName: string) {
+  SpecLogger.log(`Building ${fileName}`);
+  await esbuild
+    .build({
+      entryPoints: [fileName],
+      outdir: DESTINATION_FOLDER_NAME,
+      bundle: true,
+      format: "esm",
+      minify: true,
```
QuiiBz, line 37: That would be cool to add [`sourcemaps`](https://esbuild.github.io/api/#sourcemap) too! What do you think?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "moduleResolution": "node",
     "outDir": "./specs",
-    "target": "es5",
+    "target": "ES2020",
```
QuiiBz, line 5: Any reason why not using `ESNext`?

QuiiBz (commented):
The DX is going to be amazing! I left some comments that we can discuss

scripts/compiler.ts
fedeci: Not sure about how useful those could be here🤔

package.json
fedeci: We may create a new `watch` npm script and also keep the static `build` one.
```suggestion
    "build": "esbuild dev/*.ts dev/**/*.ts --outdir=specs --minify --format=esm --bundle",
    "watch": "ts-node-script scripts/compiler.ts",
```

fedeci (approved):

package.json
QuiiBz: > We may create a new `watch` npm script and also keep the static `build` one.

There is already the `dev` script that does this

scripts/compiler.ts
QuiiBz: This can simplify a lot debugging your specs, because you will get the exact line/column number

package.json
fedeci: Oops, sorry😆

scripts/compiler.ts
cstrnt: Hey Tom, I don't know this will work with fig, so let's maybe do this in the next iteration, okay?

fedeci: @QuiiBz I am not even sure we can easily debug the specs with that since they are not run in a node process but from a hidden webkit view.

QuiiBz: > @QuiiBz I am not even sure we can easily debug the specs with that since they are not run in a node process but from a hidden webkit view.

You can right-click on the webview to open the inspector, and here you can see the logs!

fedeci: Oh that's awesome🤩

fedeci:
```suggestion
      minify: true,
      sourcemap: "external",
```

scripts/compiler.ts
```diff
@@ -40,96 +27,38 @@ function invalidateCache() {
...
+async function processFiles(files: string | string[]) {
+  const fileName = typeof files === "string" ? files : "all specs";
+  await esbuild
+    .build({
+      entryPoints: [].concat(files),
+      outdir: DESTINATION_FOLDER_NAME,
+      bundle: true,
+      format: "esm",
+      minify: true,
+    })
```
fedeci, line 39: This works to me
```suggestion
      minify: true,
      sourcemap: "external",
    })
```

fedeci (approved):

scripts/compiler.ts
cstrnt: I added them in Dev mode because they would bloat the file size in prod mode imho.

QuiiBz (approved):
Amazing work on this! There is only a merge conflict with expo

cstrnt:
Gonna resolve the merge conflict and will merge together with the rest of the team :)

PR #389, opened 2021-07-19
[AWS] [Cloudwatch] add generators

nagypeterjob:

**Additional info:**

- [x] delete-alarms
- [x] delete-anomaly-detector
- [x] delete-dashboards
- [x] delete-insight-rules
- [x] delete-metric-stream
- [x] describe-alarm-history
- [x] describe-alarms
- [x] describe-alarms-for-metric
- [x] describe-anomaly-detectors
- [x] describe-insight-rules
- [x] disable-alarm-actions
- [x] disable-insight-rules
- [x] enable-alarm-actions
- [x] enable-insight-rules
- [x] get-dashboard
- [x] get-insight-rule-report
- [x] get-metric-data
- [x] get-metric-statistics
- [x] get-metric-stream
- [x] get-metric-widget-image
- [x] list-dashboards
- [x] list-metric-streams
- [x] list-metrics
- [x] list-tags-for-resource
- [x] put-anomaly-detector
- [x] put-composite-alarm
- [x] put-dashboard
- [x] put-insight-rule
- [x] put-metric-alarm
- [x] put-metric-data
- [x] put-metric-stream
- [x] set-alarm-state
- [x] start-metric-streams
- [x] stop-metric-streams
- [x] tag-resource
- [x] untag-resource
- [x] wait

### DIFF
--- a/dev/aws/cloudwatch.ts
+++ b/dev/aws/cloudwatch.ts
@@ -168,10 +168,7 @@ const postPrecessGenerator = (
           icon: "fig://icon?type=aws",
         };
       })
-      .filter(
-        (value, index, self) =>
-          self.map((x) => x.name).indexOf(value.name) === index
-      );
+      .filter(uniqueNames);
   } catch (e) {
     console.log(e);
   }
@@ -220,10 +217,7 @@ const listCustomGenerator = async (
           icon: "fig://icon?type=aws",
         };
       })
-      .filter(
-        (value, index, self) =>
-          self.map((x) => x.name).indexOf(value.name) === index
-      );
+      .filter(uniqueNames);
   } catch (e) {
     console.log(e);
   }
@@ -262,10 +256,7 @@ const listDimensionTypes = async (
         });
       })
       .flat()
-      .filter(
-        (value, index, self) =>
-          self.map((x) => x.name).indexOf(value.name) === index
-      );
+      .filter(uniqueNames);
   } catch (e) {
     console.log(e);
   }
@@ -382,6 +373,9 @@ const postProcessFiles = (out: string, prefix: string): Fig.Suggestion[] => {
 
   return finalArr;
 };
+
+const uniqueNames = (value, index, self) =>
+  self.map((x) => x.name).indexOf(value.name) === index;
 
 const triggerPrefix = (
   newToken: string,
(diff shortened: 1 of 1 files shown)


fedeci:
We should find a way to test generators.

dev/aws/cloudwatch.ts
```diff
@@ -1,3 +1,579 @@
...
+        const name = (childKey ? elm[childKey] : elm) as string;
+        return {
+          name,
+          icon: "fig://icon?type=aws",
+        };
+      })
+      .filter(
+        (value, index, self) =>
+          self.map((x) => x.name).indexOf(value.name) === index
+      );
```
QuiiBz, line 174: Looks like this filter is duplicated three times (https://github.com/withfig/autocomplete/pull/389/files#diff-79c9f7bd3db0aa84c298fe0e7005311e4e63c82f54f904b7c8f343e12203db37R223 & https://github.com/withfig/autocomplete/pull/389/files#diff-79c9f7bd3db0aa84c298fe0e7005311e4e63c82f54f904b7c8f343e12203db37R265). Could you extract it to a top-level const to avoid duplication?

fedeci: Q: If we run `filter` before `map` can we avoid the nested `map`?

nagypeterjob: @QuiiBz I have extracted the filter into a top-level const, nice catch.

@fedeci I can't think of an other way of doing this in the `map/filter` way without adding parameters to `tsconfig.json`. Although I am happy with any suggestion you have!

nagypeterjob: I am aware of the fact, that a nested map may have a performance toll in case of big arrays, but what is a big array in the context of AWS alerts really? Furthermore a nested map is a nested loop: you cannot do some things without them.

I could use `Map` or `Set` to solve this in a more compact way, but I did not want to edit `tsconfig` because of this single spec.

fedeci: Map or Set do not require you to edit the tsconfig. Sure without nested loops we cannot achieve many things, but we should avoid them when possible :)

QuiiBz: Let's not focus on those very small optimizations. I agree that optimizing specs is important, but I don't think that those types of optimizations are going to be noticeable to the end users. Reducing by a few ms a very specific aws subcommand is not the most important task.

QuiiBz (approved):
Thanks for this PR! It's always very useful to have generators, especially for complicated CLIs like `aws`.

Issue #391, opened 2021-07-20
[npm] Npm uninstall should be able to take more than 1 arg


Nutlope:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue

### What CLI tool does this relate to?

npm

### Which statement makes the most sense?

There is a bug with this completion spec

### Issue Details

Npm uninstall only allows 1 argument, but the user should be able to input many args to uninstall various packages at the same time.

Nutlope:
Working on this

Monkeyank:
Closed with #392

PR #392, opened 2021-07-20
feat: fix bugs in npm uninstall

Nutlope:
- Bug fixes

- Various bugs

- Add aliases of npm uninstall
- Changed icon & description of npm uninstall to make more sense
- Made npm uninstall variatic to process multiple uninstalls
- Fixed bug with npm uninstall not working with packages including @, by using a displayName

**Additional info:**
- Addresses #390 and #391

### DIFF
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -264,57 +264,12 @@ export const completionSpec: Fig.Spec = {
       description: "uninstall a package",
       args: [
         {
-          generators: {
-            script: "cat package.json",
-            postProcess: function (out) {
-              if (out.trim() === "") {
-                return [];
-              }
-              try {
-                const packageContent = JSON.parse(out);
-                const dependencies = packageContent["dependencies"];
-                if (dependencies) {
-                  const dps = Object.keys(dependencies);
-                  return dps.map((pkg) => {
-                    const scope = pkg.indexOf("/") + 1;
-                    if (scope !== -1) {
-                      pkg = pkg.substring(scope);
-                    }
-                    const version = pkg.indexOf("@");
-                    if (version !== -1) {
-                      pkg = pkg.substring(version);
-                    }
-                    return {
-                      name: pkg,
-                      icon: `fig://icon?type=file`,
-                      description: "dependency file",
-                    };
-                  });
-                }
-              } catch (e) {}
-              return [];
-            },
-          },
+          name: "package",
+          generators: dependenciesGenerator,
+          variadic: true,
         },
       ],
-      options: [
-        {
-          name: ["-S", "--save"],
-          description: " Package will be removed from your dependencies",
-        },
-        {
-          name: ["-D", "--save-dev"],
-          description: "Package will appear in your `devDependencies`",
-        },
-        {
-          name: ["-O", "--save-optional"],
-          description: "Package will appear in your `optionalDependencies`",
-        },
-        {
-          name: "--no-save",
-          description: "Prevents saving to `dependencies`",
-        },
-      ],
+      options: npmInstallOptions,
     },
     {
       name: "r",
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -34,26 +34,20 @@ const dependenciesGenerator: Fig.Generator = {
       const packageContent = JSON.parse(out);
       const dependencies = packageContent["dependencies"];
       if (dependencies) {
-        const dps = Object.keys(dependencies);
-        return dps
+        return Object.keys(dependencies)
           .filter(function (pkg) {
-            let isListed = false;
-            for (let i = 3; i <= context.length; i++) {
-              if (pkg == context[i - 2]) {
-                isListed = true;
-              }
-            }
+            const isListed = context.some((current) => current === pkg);
             return !isListed;
           })
-          .map((pkg) => {
-            const scope = pkg.indexOf("/") + 1;
-            const version = pkg.indexOf("@");
+          .map((name) => {
+            const scope = name.indexOf("/") + 1;
+            const version = name.indexOf("@");
             const displayName =
-              (scope !== -1 ? pkg.substring(scope) : pkg) ||
-              (version !== -1 ? pkg.substring(version) : pkg);
+              (scope !== -1 ? name.substring(scope) : name) ||
+              (version !== -1 ? name.substring(version) : name);
             return {
-              name: pkg,
-              displayName: displayName,
+              name,
+              displayName,
               description: "dependency",
             };
           });
@@ -63,7 +57,7 @@ const dependenciesGenerator: Fig.Generator = {
   },
 };
 
-const npmInstallOptions = [
+const npmInstallOptions: Fig.Option[] = [
   {
     name: ["-S", "--save"],
     description: " Package will be removed from your dependencies",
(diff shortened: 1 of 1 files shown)


dev/npm.ts
```diff
@@ -24,6 +24,64 @@ const searchGenerator: Fig.Generator = {
...
+              description: "dependency",
+            };
+          });
+      }
+    } catch (e) {}
+    return [];
+  },
+};
+
+const npmInstallOptions = [
```
QuiiBz, line 66: Would love to see a  `Fig.Option[]` type here!

dev/npm.ts
```diff
@@ -24,6 +24,64 @@ const searchGenerator: Fig.Generator = {
...
+    description: "Package will appear in your `devDependencies`",
+  },
+  {
+    name: ["-O", "--save-optional"],
+    description: "Package will appear in your `optionalDependencies`",
+  },
+  {
+    name: "--no-save",
+    description: "Prevents saving to `dependencies`",
+  },
```
QuiiBz, line 82: I personally use yarn, but I think you forgot the `-P` option for peerDependencies

dev/npm.ts
```diff
@@ -265,61 +323,52 @@ export const completionSpec: Fig.Spec = {
...
         {
-          name: "--no-save",
-          description: "Prevents saving to `dependencies`",
+          name: "package",
+          generators: dependenciesGenerator,
+          variadic: true,
         },
       ],
+      options: npmInstallOptions,
     },
```
QuiiBz, line 372: Instead of duplicating those subcommands (uninstall, remove, rm, r), we can set the `name` property to an array. Those will represent the aliases of the subcommands, in this case:
```ts
{
  name: ["uninstall", "remove", "rm", "r"],
  description: "Remove a package",
  ...
```

dev/npm.ts
```diff
@@ -24,6 +24,64 @@ const searchGenerator: Fig.Generator = {
...
+    if (out.trim() === "") {
+      return [];
+    }
+    try {
+      const packageContent = JSON.parse(out);
+      const dependencies = packageContent["dependencies"];
+      if (dependencies) {
+        const dps = Object.keys(dependencies);
+        return dps
+          .filter(function (pkg) {
```
QuiiBz, line 39: Small improvement:
```suggestion
        return Object.keys(dependencies);
          .filter((pkg) => {
```

dev/npm.ts
```diff
@@ -24,6 +24,64 @@ const searchGenerator: Fig.Generator = {
...
+            const scope = pkg.indexOf("/") + 1;
+            const version = pkg.indexOf("@");
+            const displayName =
+              (scope !== -1 ? pkg.substring(scope) : pkg) ||
+              (version !== -1 ? pkg.substring(version) : pkg);
+            return {
+              name: pkg,
+              displayName: displayName,
+              description: "dependency",
+            };
```
QuiiBz, line 58: Small improvement:
```suggestion
          .map((name) => {
            const scope = name.indexOf("/") + 1;
            const version = name.indexOf("@");
            const displayName =
              (scope !== -1 ? name.substring(scope) : name) ||
              (version !== -1 ? name.substring(version) : name);
            return {
              name,
              displayName,
              description: "dependency",
            };
```

dev/npm.ts
```diff
@@ -24,6 +24,64 @@ const searchGenerator: Fig.Generator = {
...
+        const dps = Object.keys(dependencies);
+        return dps
+          .filter(function (pkg) {
+            let isListed = false;
+            for (let i = 3; i <= context.length; i++) {
+              if (pkg == context[i - 2]) {
+                isListed = true;
+              }
+            }
+            return !isListed;
```
QuiiBz, line 46: I think we can simplify this a bit, not tested but something like:
```suggestion
            const isListed = context.some((current) => current === pkg);
            return !isListed;
```

QuiiBz (changes requested):
Nice fixes! I left some code improvement suggestions, feel free to discuss about them

dev/npm.ts
fedeci: It is also strange that the iteration was starting with `i = 3` 🤔

QuiiBz: > It is also strange that the iteration was starting with `i = 3` 🤔

I think it's because we want to skip the first two tokens, which are `npm <uninstall/remove/rm/r>`

fedeci: I thought at that but I believe it is fairly confusing.

dev/npm.ts
Nutlope: I looked into it and the -P option seems to be for just saving as a regular dependency, so it's the default behavior. So unless I'm mistaken, it doesn't seem to be necessary?

dev/npm.ts
Nutlope: Yes but the user experience isn't great when we do this so I explicitly did them separately. See the example below. What do you think?
image

dev/npm.ts
Nutlope: Wow @QuiiBz, great suggestion! I swear I learn something new every time you review my PRs, thanks for making me a better JS programmer! @fedeci yeah, I wanted to skip the first 2 tokens but agreed, it is definitely confusing. Replacing with Tom's suggestion :)

QuiiBz: Thanks for those kind words @Nutlope!

dev/npm.ts
QuiiBz: Personally, I don't think that is an issue. cc @cstrnt

dev/npm.ts
QuiiBz: Found this in the docs, looks like it's not really documented, but `--save-peer` does exist. You're right for `-P`, it's an alias for `--save-prod`, sorry for this! So let's add the `--save-peer` option

Screenshot 2021-07-21 at 08 35 02

dev/npm.ts
```diff
@@ -24,6 +24,58 @@ const searchGenerator: Fig.Generator = {
...
+    }
+    try {
+      const packageContent = JSON.parse(out);
+      const dependencies = packageContent["dependencies"];
+      if (dependencies) {
+        return Object.keys(dependencies)
+          .filter(function (pkg) {
+            const isListed = context.some((current) => current === pkg);
+            return !isListed;
+          })
```
fedeci, line 41: nit:
```suggestion
          .filter((pkg) => !context.some((current) => current === pkg))
```

QuiiBz: Agree for the arrow function, but we should keep `isListed` - we understand better what's happening.

Just a quick note, we can use `every` instead of `some` to remove the logical NOT.

fedeci: I agree on the `every` but would use a comment to understand what we are doing.

dev/npm.ts
Nutlope: Looked into --save-peer, it seems like it doesn't do anything. It's both not documented and doesn't save a package as a peer dependency (just tested it out with a new project). So seems a bit useless to add a command that doesn't actually do anything & is not documented, unless i'm missing something. Thoughts?

dev/npm.ts
Nutlope: It's a very used command so I think it's better to separate. Gonna keep it like this for now

PR #394, opened 2021-07-20
perf: use `Set` when possible

fedeci:
We should prefer using `Sets` in the most cases when evaluating known strings.

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -217,7 +217,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
         //  M dev/kubectl.ts // test file that was not added
         // A  test2.txt // new added and tracked file
         // ?? test.txt // new untracked file
-        const alreadyAdded = ["M", "A"].includes(file.charAt(0));
+        const alreadyAdded = new Set(["M", "A"]).has(file.charAt(0));
 
         file = file.trim();
         const arr = file.split(" ");
--- a/dev/fig.ts
+++ b/dev/fig.ts
@@ -126,12 +126,12 @@ export const completionSpec: Fig.Spec = {
             console.log(data);
             const out = data.split("\n").reduce((acc, curr) => {
               if (
-                [
+                new Set([
                   ".gitignore",
                   "README.md",
                   "package.json",
                   "package-lock.json",
-                ].includes(curr)
+                ]).has(curr)
               )
                 return acc;
               else {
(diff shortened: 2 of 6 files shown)


dev/git.ts
```diff
@@ -217,7 +217,7 @@ const gitGenerators: Record<string, Fig.Generator> = {
         //  M dev/kubectl.ts // test file that was not added
         // A  test2.txt // new added and tracked file
         // ?? test.txt // new untracked file
-        const alreadyAdded = ["M", "A"].includes(file.charAt(0));
+        const alreadyAdded = new Set(["M", "A"]).has(file.charAt(0));
```
fedeci, line 220: This may be the only case where Set is overkill, because creating it from array is `O(n)`.

QuiiBz:
Thanks for this PR! However, I don't think we really need this. The average items in those arrays are not a lot (less than 10 based on the changed files), so this won't matter enough in terms of performance.

It might also be harder for new devs to understand what a `Set` is, whereas an array with `includes` is more common, as we should make sure that specs are the most accessible to write and understand.

Here is a very small benchmark on an array of 9 items: https://www.measurethat.net/Benchmarks/Show/7818/0/sethas-vs-arrayincludes

fedeci:
Consider a couple of aspects:
1. `Set`s should not be considered as an unknown data structure because they are extremely powerful in many situations;
2. Many of those `.includes` are run in a "loop" context (filter, map...) so the performance gains are not that obvious;
3. Even if it is probably debounced in the Swift engine generators are run several times.

QuiiBz:
I agree that `Set`s are really important - but we should set easy standards that everyone knows. It would be interesting to run a benchmark on this to really see if the difference is noticeable for the user (which I don't think so)

fedeci:
Sorry but I have to disagree with you again because we target `ESNext` in `tsc` which exactly where all the unknown and not-yet-in-the-ecmascript-spec features live but we should not use `Set`🤔 
I also don't like the idea that if something is not visible to the end user it should not be optimized. With `Set` we get an optimization almost for free and if in the future those specs will have to handle more than 10 items they will be ready to do so without problems.

QuiiBz:
@cstrnt any thought on this?

QuiiBz:
Closing this since it seems over-engineered and won't provide noticeable performances. We have a lot more optimizations possibles in the autocomplete-engine itself, but thank you very much for this PR!

fedeci:
"over-engineered"

fisker:
The eslint rule [unicorn/prefer-set-has](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-set-has.md) is what you need.

PR #399, opened 2021-07-22
[gradle] Add Spec

Monkeyank:
Feature
Adds `gradle` spec. (Although `./gradlew` doesn't work 😢)

### DIFF
--- a/dev/gradle.ts
+++ b/dev/gradle.ts
@@ -20,7 +20,7 @@ const consoleModeArg: Fig.Arg = {
     {
       name: "verbose",
       description:
-        "enable color and other rich output like the rich, but output task names and outcomes at the lifecycle log level, as is done by default in Gradle 3.5 and earlier",
+        "Enables color and other rich output like the rich, but output task names and outcomes at the lifecycle log level, as is done by default in Gradle 3.5 and earlier",
     },
   ],
 };
@@ -350,8 +350,6 @@ const sharedOptions: Fig.Option[] = [
 ];
 
 export const completion: Fig.Spec = {
-  // ./gradlew doesn't work
-  // name: ["gradle", "./gradlew"],
   name: "gradle",
   description:
     "Gradle is an open-source build automation tool that is designed to be flexible enough to build almost any type of software",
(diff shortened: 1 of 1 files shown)


dev/gradle.ts
```diff
@@ -0,0 +1,538 @@
...
+    },
+    {
+      name: "rich",
+      description:
+        "Enables color and other rich output in the console output, regardless of whether the build process is not attached to a console",
+    },
+    {
+      name: "verbose",
+      description:
+        "enable color and other rich output like the rich, but output task names and outcomes at the lifecycle log level, as is done by default in Gradle 3.5 and earlier",
```
QuiiBz, line 23:
```suggestion
        "Enable color and other rich output like the rich, but output task names and outcomes at the lifecycle log level, as is done by default in Gradle 3.5 and earlier",
```

dev/gradle.ts
```diff
@@ -0,0 +1,538 @@
...
+    args: {
+      name: "directory",
+      template: "folders",
+    },
+  },
+];
+
+export const completion: Fig.Spec = {
+  // ./gradlew doesn't work
+  // name: ["gradle", "./gradlew"],
```
QuiiBz, line 354: I think gradlew should be in a separate spec

dev/gradle.ts
```diff
@@ -0,0 +1,538 @@
...
+          description:
+            "Gradle will execute every task to be executed where all of the dependencies for that task completed without failure",
+        },
+        ...sharedOptions,
+      ],
+    },
+  ],
+  options: [...sharedOptions],
+
+  // There isn't a generator for gradle tasks (takes too long)
```
QuiiBz, line 532: We can read the `build.gradle` file and put a regex to find all the tasks - I don't think that would be slow

QuiiBz (changes requested):
Entering the Java world! Almost a perfect PR, but I think we should add a generator for the tasks. See the comment on this below

dev/gradle.ts
Monkeyank: I don't think the build.gradle contains all the tasks. At least based on some of my projects that I've checked out

QuiiBz: Yes, we can add the default tasks as suggestions and additional tasks from the build.gradle file using a generator

QuiiBz (approved):
LGTM!

PR #400, opened 2021-07-22
[fzf] Add spec

Monkeyank:
Feature

Adds `fzf` spec

### DIFF
--- a/dev/fzf.ts
+++ b/dev/fzf.ts
@@ -46,7 +46,6 @@ export const completion: Fig.Spec = {
       args: {
         name: "index expressions",
         description: "non-zero integer or range expression ([BEGIN]..[END])",
-        variadic: true,
       },
     },
     {
@@ -56,11 +55,10 @@ export const completion: Fig.Spec = {
       args: {
         name: "index expressions",
         description: "non-zero integer or range expression ([BEGIN]..[END])",
-        variadic: true,
       },
     },
     {
-      name: "-d, --delimiter=STR",
+      name: ["-d", "--delimiter"],
       description: "Field delimiter regex (default: AWK-style)",
       args: {
         name: "STR",
@@ -87,7 +85,6 @@ export const completion: Fig.Spec = {
         name: "criteria",
         default: "length",
         suggestions: ["length", "begin", "end", "index"],
-        variadic: true,
       },
     },
     {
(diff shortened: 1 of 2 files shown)


dev/fzf.ts
```diff
@@ -0,0 +1,376 @@
...
+      description: "Do not normalize latin script letters before matching",
+    },
+    {
+      name: ["-n", "--nth"],
+      description:
+        "Comma-separated list of field index expressions for limiting search scope",
+      args: {
+        name: "index expressions",
+        description: "non-zero integer or range expression ([BEGIN]..[END])",
+        variadic: true,
```
QuiiBz, line 49: I don't think that this arg is `variadic`: from the description, the args are separated by a comma, not a space

dev/fzf.ts
```diff
@@ -0,0 +1,376 @@
...
+      },
+    },
+    {
+      name: "--with-nth",
+      description:
+        "Transform the presentation of each line using field index expressions",
+      args: {
+        name: "index expressions",
+        description: "non-zero integer or range expression ([BEGIN]..[END])",
+        variadic: true,
```
QuiiBz, line 59: Maybe the same as above?

dev/fzf.ts
```diff
@@ -0,0 +1,376 @@
...
+      },
+    },
+    {
+      name: "-d, --delimiter=STR",
```
QuiiBz, line 63: Should be:
```suggestion
      name: ["-d", "--delimiter"],
```

dev/fzf.ts
```diff
@@ -0,0 +1,376 @@
...
+    },
+    {
+      name: "--tiebreak",
+      description:
+        "Comma-separated list of sort criteria to apply when the scores are tied",
+      args: {
+        name: "criteria",
+        default: "length",
+        suggestions: ["length", "begin", "end", "index"],
+        variadic: true,
```
QuiiBz, line 90: Same as above?

dev/fzf.ts
```diff
@@ -0,0 +1,376 @@
...
+    {
+      name: "--print-query",
+      description: "Print query as the first line",
+    },
+    {
+      name: "--expect",
+      description: "Comma-separated list of keys to complete fzf",
+      args: {
+        name: "keys",
+        variadic: true,
```
QuiiBz, line 356: Same as above?

QuiiBz (changes requested):
Very cool spec! I only left a review on `fzf.ts` since they both are the same

QuiiBz (approved):
LGTM!

Issue #304, opened 2021-06-15
[Docker] Give priority to containers names rather than ID for autocompletion


crezra:
### Description:
> Please include a detailed description of the issue (and an image or screen recording, if applicable)

Typing a Docker command such as : `docker exec -ti ` makes fig give autocompletion for currently running containers, which is good, however the autocompletion is based on the container ID (i.e. dd424db56428) rather than the container_name (i.e. project_web_1) 

I believe it would be more convenient to autocomplete the name first, and then fallback to the ID. 

### Details:
|macOS|Fig|Shell|
|----|----|----|
|11.4.0|Version 1.0.41 (B206)|/bin/zsh|

<details><summary><code>fig diagnostic</code></summary>
<p>
<pre>Version 1.0.41 (B206)
UserShell: /bin/zsh
Bundle path: /Applications/Fig.app
Autocomplete: true
Settings.json: true
CLI installed: true
CLI tool path: /Users/*****/.fig/bin/fig
Accessibility: true
Number of specs: 98
SSH Integration: true
Tmux Integration: true
Keybindings path: /Users/*****/.fig/user/keybindings
iTerm Integration: true
Hyper Integration: false
VSCode Integration: true
Docker Integration: true
Symlinked dotfiles: false
Only insert on tab: false
Installation Script: true
PseudoTerminal Path: /Users/****_/.composer/vendor/bin:/Users/_****/.zulu/bin:/Users/*****/.nvm/versions/node/v16.1.0/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
SecureKeyboardInput: false
SecureKeyboardProcess: <none>
Current active process: /bin/zsh (667) - ttys000
Current working directory: /Users/*****/**

- - -
Current window identifier: 109/1CA2E47EC-DE58-435E-9282-57AC6DA4657D% (com.googlecode.iterm2)</pre>
</p>
</details>

QuiiBz:
Merged in #402, thanks a lot!

PR #402, opened 2021-07-23
[Docker] show container names rather than the IDs

dereklowlind:
On input from issue https://github.com/withfig/autocomplete/issues/304 changed the Docker autocomplete generator to show container names rather than the IDs it currently shows.


QuiiBz (approved):
Just tested this change, it's really better! Thanks a lot for this PR

PR #404, opened 2021-07-23
[git] Prioritize current branch

QuiiBz:
Feature

The current branch is distinct by a ⭐️  icon, but can be anywhere in the suggestions

Adds a priority of `100` to the current branch so it will always stay at the top

**Additional info:**
Very useful in this case:
Screenshot 2021-07-23 at 09 21 22


Monkeyank (approved):
Nice catch! I remembered this last week, but I forgot to push it.

QuiiBz:
> Nice catch! I remembered this last week, but I forgot to push it.

Thanks! Yeah it's pretty useful

Issue #403, opened 2021-07-23
[git] tag support for checkout


fwesss:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue

### What CLI tool does this relate to?

git

### Which statement makes the most sense?

This completion spec isn't complete

### Issue Details

git checkout does not currently autocomplete on tags though this is something you can do with the checkout subcommand.

https://git-scm.com/book/en/v2/Git-Basics-Tagging#_checking_out_tags

PR #409, opened 2021-07-26
git[checkout]: show tags in checkout suggestions

fwesss:
Feature

git checkout will list branches, files, and commits to checkout but not tags.

Tags will be listed in the suggestions for git checkout.

**Additional info:**
The current branch is normally given top priority with a star icon. However, if we are in a detached HEAD state after checking out a tag, I don't think we want to suggest the current branch. It would just show (detached HEAD state) and isn't actionable suggestion.

Re:https://github.com/withfig/autocomplete/issues/403

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -57,6 +57,10 @@ const postProcessBranches: Fig.Generator["postProcess"] = (out) => {
     const parts = elm.match(/\S+/g);
     if (parts.length > 1) {
       if (parts[0] === "*") {
+        // We are in a detached HEAD state
+        if (elm.includes("HEAD detached")) {
+          return {};
+        }
         // Current branch.
         return {
           name: elm.replace("*", "").trim(),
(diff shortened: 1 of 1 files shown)


fedeci (commented):
Maybe we should assign a higher priority to branches rather than commits or tags.

Nutlope:
> Maybe we should assign a higher priority to branches rather than commits or tags.

I agree with @fedeci. Maybe branches > commits > tags?

fwesss:
I've increased the priority given to branches in postProcessBranches.
Looks like the only time this generator is used cumulatively is when generating suggestions for tags so I don't think this change in prioritization would affect the results in any other commands.

I'm not a super git expert or 100% familiar with this spec though so any suggestions or comments are appreciated!

dev/git.ts
```diff
@@ -3966,10 +3976,11 @@ export const completionSpec: Fig.Spec = {
       args: [
         {
           name: "branch or file",
```
Nutlope, line 3978: maybe change this to: "branch, file, or tag"?

Nutlope:
> I've increased the priority given to branches in postProcessBranches.
> Looks like the only time this generator is used cumulatively is when generating suggestions for tags so I don't think this change in prioritization would affect the results in any other commands.

Just tested it out and it works great! Just added one more minor comment

dev/git.ts
fwesss: Fixed!

Nutlope:
LGTM!

PR #410, opened 2021-07-27
git: use --no-optional-locks with background status cmd

fwesss:
This PR adds in the --no-optional-locks flag everywhere git status is run as a background command.
A index.lock file will not be created when status is run so user git commands will not be blocked from running.

Bug fix

Background `git status` commands create an index.lock file which can interfere with user commands if the repo is large enough. The .lock file is still in place while Fig is generating completions and blocks the user from exec their own command.

Skip creating the .lock file when we run git status.

**Additional info:**
See explanation: https://git-scm.com/docs/git-status#_background_refresh
Re: https://github.com/withfig/fig/discussions/296

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -86,7 +86,7 @@ const postProcessBranches: Fig.Generator["postProcess"] = (out) => {
 const gitGenerators: Record<string, Fig.Generator> = {
   // Commit history
   commits: {
-    script: "git log --oneline",
+    script: "git --no-optional-locks log --oneline",
     postProcess: function (out) {
       const output = filterMessages(out);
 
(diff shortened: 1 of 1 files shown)


dev/git.ts
```diff
@@ -498,6 +499,10 @@ export const completionSpec: Fig.Spec = {
       name: "--no-replace-objects",
       description: "Do not use replacement refs",
     },
+    {
+      name: "--no-optional-locks",
+      description: "Do not perform optional operations that require locks",
```
fedeci, line 504:
```suggestion
      description: "Do not perform optional operations that require lock files",
```

fedeci (approved):

fwesss:
Added --no-optional-locks to all git commands run by Fig.

A common issue is getting suggestions for git checkout in a large repo. We run git branch -l which creates a lock file.

I can't think of a reason that Fig would be modifying a git repo in any way so we should not need to generate a lock file to prevent editing. We should be able to avoid creating a lock for all commands but want to hear if there could be issues I haven't thought of.

mschrage (approved):

QuiiBz (approved):
That's smart! Thanks a lot for this contribution

PR #412, opened 2021-07-28
[tmux] added additional suggestion

Nutlope:
feature
no shortcuts
added additional "shortcut" suggestion for a popular workflow
**Additional info:**

### DIFF
--- a/dev/tmux.ts
+++ b/dev/tmux.ts
@@ -2113,4 +2113,13 @@ export const completion: Fig.Spec = {
       ],
     },
   ],
+  additionalSuggestions: [
+    {
+      name: "new -S 'name'",
+      description: "Create a new session shortcut",
+      insertValue: "new -S '{cursor}'",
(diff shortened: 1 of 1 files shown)


dev/tmux.ts
```diff
@@ -2113,4 +2113,13 @@ export const completion: Fig.Spec = {
       ],
     },
   ],
+  additionalSuggestions: [
+    {
+      name: "new -S 'name'",
+      description: "Create a new session shortcut",
+      insertValue: "new -S '{cursor}'",
+      icon: "fig://template?color=2ecc71&badge=🔥",
+      // type: "shortcut",
```
Monkeyank, line 2122: Maybe remove this comment?

Nutlope: There is only one other "shortcut" in git and it also had that comment. Wanted to keep it for parity in case they wanna change up the API to add that option.

PR #413, opened 2021-07-28
[tfenv] Add tfenv

menningkeneiext:
Add functionality for [tfenv](https://github.com/tfutils/tfenv), a Terraform version manager.


dev/tfenv.ts
```diff
@@ -0,0 +1,84 @@
...
+  servicesgenerators: {
+    script: "tfenv list",
+    postProcess: function (out) {
+      return out.split("\n").map((tfversion) => {
+        return { name: tfversion, description: "Version" };
+      });
+    },
+  },
+};
+
```
mschrage, line 22: The canonical way of including multiple generators is **creating a single dictionary** that contains all the generators for the completion spec:

```
const generators: Record<string, Fig.Generator> = {
   installedVersions: { ... },
   allVersions: { ... }
}
```

And then you can use these generators in a completion spec by referencing them like this:
```
{
    name: "version",
    description: "Possible Terraform Version",
    suggestions: ["latest", "min-required"],
    generators: generators.allVersions
}
```

mschrage (changes requested):
Thanks so much for contributing @menningkeneiext! 🎉🤩

I think this is basically ready to merge in. You've done a great job of including all the subcommands, options and arguments in the spec.

The one thing to change is how the generators work. This is mainly to keep things consistent across all completion specs. See below:

menningkeneiext:
@mschrage Ah, this is very helpful, thank you. Relatively new to coding and my first time coding in JS, so it was a good learning experience.

Added new commits.

dev/tfenv.ts
```diff
@@ -0,0 +1,81 @@
...
+          name: "version",
+          description: "Possible Terraform Version",
+          suggestions: ["latest", "min-required"],
+          generators: generators.allVersions,
+        },
+      ],
+    },
+    {
+      name: "use",
+      description: "Swith to a version to use",
```
QuiiBz, line 38: Small typo:
```suggestion
      description: "Switch to a version to use",
```

mschrage (approved):

PR #416, opened 2021-07-29
[FASTLY] Update Fastly spec to 0.35.0

kailan:
This release includes:

* Support for Compute@Edge JS SDK (Beta)
* New --override-host and --ssl-sni-hostname options
* ACL commands
* ACL entry commands
* Internal refactor
* Fix compute serve --addr option description to reference port requirement

This diff is a bit unwieldy on GitHub, but as always has been generated automatically using [figly](https://github.com/kailan/figly).


QuiiBz (approved):
That's a big update! As always I won't request changes since it's generated using this amazing tool figly!

PR #419, opened 2021-07-29
[expo cli] auto generate scheme

EvanBacon:

The `expo` and `expo-cli` schemas were a bit behind and lacking in features. Given the size of the expo-cli, we have a script that we use to generate the documentation, I've updated that script to support generating a fig file automatically https://github.com/expo/expo-cli/pull/3727 -- the test plan in the expo-cli PR applies to this PR as well. 

**Additional info:**

Some notes about my experience:

 - `xcodebuild -project ios/*.xcodeproj -list -json` doesn't always work when run in the generator, not sure how to debug or improve this. The closest comparison I can think of is VSCode's toast element which shows errors that don't really have a dedicated UI space.
 - The API `https://api.npms.io/v2/search` (based on the npm and yarn files) isn't very robust, it can't find packages like `expo-camera` (failing after `expo-c`). If there's a better API, I'd be happy to implement that.
- `expo run:ios --device` takes too long to load devices, a cold run of Apple's `time xcrun xctrace list devices` can take upwards of 10s (11.822s just now). Because `--device` is optional, other options are loaded in and the devices never show up. Perhaps fig could provide some async API for presenting loading options.
- The [icon docs](https://fig.io/docs/reference/suggestion/icon-api) say to open an issue for missing icons. As you may notice, I use a lot of different icons here. Could be nice to have icons for: login/logout/whoami, cloud/publish, help (--help), version (--version), true/false, number (for port and max-workers), skip (for things like no-install, and yes), and maybe something destructive for clear, and clean type options.
  - I attempted to make a "number" icon using `fig://template?color=2ecc71&badge=#` but it doesn't work, the `#` symbol doesn't render.

All-in-all, I'm really impressed with the DX/contributing of this repo and think the team has done a great job!


dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
         },
         {
           name: ["-t", "--type"],
-          description: "Type of build: [app-bundle|apk].\n",
-          args: {},
+          description: "Type of build: [archive|simulator].",
+          args: {
+            name: "build",
+            isOptional: false,
+          },
```
Monkeyank, line 237: Add `suggestions: ["archive", "simulator"]

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
+            suggestions: [
+              {
+                name: "true",
+                icon: "fig://template?color=2ecc71&badge=✓",
+              },
+              {
+                name: "false",
+                icon: "❌",
+              },
+            ],
```
Monkeyank, line 125: Can we create a top level const encompassing this arg so we can reuse it!

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
+            isOptional: false,
+          },
         },
         {
-          name: ["--clear-push-key"],
+          name: ["--dist-p12-path"],
           description:
-            "Remove Push Notifications Key stored on Expo servers.\n",
-          args: {},
+            "Path to your Distribution Certificate P12 (set password as EXPO_IOS_DIST_P12_PASSWORD environment variable).",
```
Monkeyank, line 297: Seems like this argument should be taking a path rather than a boolean.
Add `template: "filepaths"` and remove the other suggestions.

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
+          name: ["--push-id"],
+          description: "Push Key ID (ex: 123AB4C56D).",
+          args: {
+            name: "push-id",
+            isOptional: false,
+          },
+        },
+        {
+          name: ["--push-p8-path"],
+          description: "Path to your Push Key .p8 file.",
```
Monkeyank, line 323: Seems like this argument should be taking a path rather than a boolean.
Add `template: "filepaths"` and remove the other suggestions.

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
+              {
+                name: "false",
+                icon: "❌",
+              },
+            ],
+          },
+        },
+        {
+          name: ["--provisioning-profile-path"],
+          description: "Path to your Provisioning Profile.",
```
Monkeyank, line 341: Seems like this argument should be taking a path rather than a boolean.
Add `template: "filepaths"` and remove the other suggestions.

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
+            ],
+          },
+          icon: "⏭",
         },
         {
-          name: ["--team-id"],
-          description: "Apple Team ID.\n",
-          args: {},
+          name: ["--keystore-path"],
+          description: "Path to your Keystore.",
```
Monkeyank, line 505: Seems like this argument should be taking a path rather than a boolean.
Add `template` and remove the other suggestions.

Monkeyank (changes requested):
Great PR! Really liked how you used `yarn introspect fig`. 

There's a few places where I believe you can add suggestions for filepaths. If you could just go through once and change those up, that'd be great.

Also I think you can move repeating options/args to top level consts, but since this spec is being generated automatically by `yarn introspect fig` you don't have to bother with it.

@QuiiBz Will probably have some additional comments 🙂

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
-          args: {},
-        },
-        {
-          name: ["--skip-credentials-check"],
-          description: "Skip checking credentials.\n",
-          args: {},
+            "The URL of an externally hosted manifest (for self-hosted apps)",
+          args: {
+            name: "url",
+            isOptional: false,
```
Monkeyank, line 553: Specifying: `isOptional: false`, isn't required as that's the default.

dev/expo-cli.ts
```diff
@@ -1,1546 +1,4491 @@
...
         },
         {
-          name: ["--config"],
-          description: "Specify a path to app.json or app.config.js\n",
-          args: {},
+          name: ["-t", "--type"],
+          description: "Type of build: [app-bundle|apk].",
+          args: {
+            name: "build",
+            isOptional: false,
```
Monkeyank, line 580: `suggestions: ["app-bundle", "apk"]`

Monkeyank (changes requested):

QuiiBz:
Thanks a lot for this PR! Some toughts on your feedback:

> * `expo run:ios --device` takes too long to load devices, a cold run of Apple's `time xcrun xctrace list devices` can take upwards of 10s (11.822s just now). Because `--device` is optional, other options are loaded in and the devices never show up. Perhaps fig could provide some async API for presenting loading options.

We currently set a timeout of 5 seconds for generators to avoid dead-locks. It should be fairly easy to implement a custom timeout for each generator, but the speed of this command would depend on the speed of the user's device. @sullivan-sean any ideas on this?

> * The [icon docs](https://fig.io/docs/reference/suggestion/icon-api) say to open an issue for missing icons. As you may notice, I use a lot of different icons here. Could be nice to have icons for: login/logout/whoami, cloud/publish, help (--help), version (--version), true/false, number (for port and max-workers), skip (for things like no-install, and yes), and maybe something destructive for clear, and clean type options.

We will definitely add those icons to the list!

>   * I attempted to make a "number" icon using `fig://template?color=2ecc71&badge=#` but it doesn't work, the `#` symbol doesn't render.

I think that it's due to the browser interpreting the `#` as a URL fragment. Not sure if the escaped string would work (`%23`). You can also use an URL to a png/svg/ico file instead of the `fig://` schem.

dev/expo-cli.ts
EvanBacon: yup finding a handful of invalid args in the CLI, commander isn't very well typed. Catching these issues is how I'm justifying spending so much time on the generation script :]

EvanBacon:
@QuiiBz thanks for the response, I ended up creating custom icons for everything (as you may have noticed from the bot comment :] ). Thanks for the feedback!

dev/expo-cli.ts
```diff
@@ -1,1546 +1,5408 @@
...
+      priority: 59,
+      args: [
         {
-          name: ["-c", "--clear-credentials"],
-          description: "Clear stored credentials.\n",
-          args: {},
+          isOptional: true,
+          template: "folders",
+          name: "path",
+          variadic: false,
```
QuiiBz, line 103: You don't have to specify `variadic` when it's false - it's assumed by default!

dev/expo-cli.ts
```diff
@@ -1,1546 +1,5408 @@
...
         {
-          name: ["--skip-workflow-check"],
+          name: ["--apple-id"],
           description:
-            "Skip warning about build service bare workflow limitations.\n",
-          args: {},
+            "Apple ID username (please also set the Apple ID password as EXPO_APPLE_PASSWORD environment variable).",
+          args: {
+            name: "login",
+            isOptional: false,
```
QuiiBz, line 255: `isOptional` is false by default so you don't have to specify it

QuiiBz (changes requested):
Small suggestions for useless properties, this PR seems good overall! We will be ready to merge after those changes.

QuiiBz (approved):
LGTM! Thanks a lot for this amazing PR!

PR #422, opened 2021-07-30
[FIG] feat: add theme command to fig spec

cstrnt:
Feature

Completions for `fig theme` don't exist yet

`fig theme` works!

**Additional info:**

### DIFF
--- a/dev/fig.ts
+++ b/dev/fig.ts
@@ -59,6 +59,47 @@ const SETTINGS_GENERATOR: Record<string, Fig.Generator> = {
   "autocomplete.disableForCommands": disableForCommandsGenerator,
 };
 
+interface GithubApiResult {
+  tree: Array<{
+    path: string;
+    type: "blob" | "tree";
+  }>;
+}
+
+const themeGenerator: Fig.Generator = {
+  custom: async (_, executeShellCommand) => {
+    try {
+      const output = await executeShellCommand(
+        'curl -s -H "Accept: application/json" "https://api.github.com/repos/withfig/themes/git/trees/main?recursive=1"'
+      );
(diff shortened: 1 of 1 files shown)


dev/fig.ts
```diff
@@ -59,6 +59,47 @@ const SETTINGS_GENERATOR: Record<string, Fig.Generator> = {
...
+const themeGenerator: Fig.Generator = {
+  custom: async (_, executeShellCommand) => {
+    try {
+      const output = await executeShellCommand(
+        'curl -s -H "Accept: application/json" "https://api.github.com/repos/withfig/themes/git/trees/main?recursive=1"'
+      );
+      const builtInThemes = ((JSON.parse(
+        await executeShellCommand(`cat ${SETTINGS_PATH}`)
+      ) as Settings[])?.find((e) => e.settingName === "autocomplete.theme")
+        ?.options ?? []) as string[];
```
QuiiBz, line 78: This might be a bit hard to read and understand. I would prefer something like (not tested):
```suggestion
      const settings: Settings[] = JSON.parse(
        await executeShellCommand(`cat ${SETTINGS_PATH}`)
      ) || [];
      
      const builtInThemes: string[] = settings.find(({ settingName }) => settingName === "autocomplete.theme")?.options ?? [] 
```

QuiiBz (commented):
Looks good, only a personal readability issue!

mschrage (changes requested):
I think this is overcomplicating things. The theme command is just switching between locally json files locally installed in the `~/.fig/themes` folder. I think we can just list the files in that directory, unless I'm missing something?

cstrnt:
closing this in favor of #445

PR #427, opened 2021-08-02
[codex] WIP

brendanfalk:
Add support for codex on open AI. Need to merge fig's endpoint on its server first


brendanfalk:
@cstrnt @QuiiBz Typescript question + linting question
Typescript: How can I add a type for `window` to this so we don't get errors
Linting: I made a tiny change to git switch. The insert value is not exactly the same as the name but it is showing a linting error... What's going on?

QuiiBz:
That looks amazing!

> Typescript: How can I add a type for `window` to this so we don't get errors

You have to declare in the global scope the Window interface, and typescript will merge it with the default one:
```ts
declare global {
  interface Window {
    fetch: async () => ...
  }
}
```

> Linting: I made a tiny change to git switch. The insert value is not exactly the same as the name but it is showing a linting error... What's going on?

I think the rule doesn't check for spaces at the end. I think the rule does its job here, we should remove the `insertValue` since it only adds a space at the end.

QuiiBz:
> Need to merge fig's endpoint on its server first

Is the endpoint live, so we can fix and merge this PR?

PR #431, opened 2021-08-02
Update gh spec

EliasGcf:

Update the `gh` (GitHub's CLI tool) spec.

### DIFF
--- a/dev/gh.ts
+++ b/dev/gh.ts
@@ -14,43 +14,666 @@ const ghGenerators: Record<string, Fig.Generator> = {
         };
       }),
   },
+  listAlias: {
+    script: "gh alias list",
+    postProcess: (out) => {
+      const aliases = out.split("\n").map((line) => {
+        const [name, content] = line.split(":");
+
+        return { name: name.trim(), content: content.trim() };
+      });
+
+      return aliases.map(({ name, content }) => ({
+        name,
+        displayName: name,
+        description: content,
+        icon: "fig://icon?type=command",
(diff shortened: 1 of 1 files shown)


dev/gh.ts
```diff
@@ -14,43 +14,666 @@ const ghGenerators: Record<string, Fig.Generator> = {
...
+    postProcess: (out) => {
+      const aliases = out.split("\n").map((line) => {
+        const [name, content] = line.split(":");
+
+        return { name: name.trim(), content: content.trim() };
+      });
+
+      return aliases.map(({ name, content }) => ({
+        name,
+        displayName: name,
```
QuiiBz, line 28: There is no need to set `displayName `if it's the same as `name`

dev/gh.ts
```diff
@@ -14,43 +14,666 @@ const ghGenerators: Record<string, Fig.Generator> = {
...
+                "The protocol to use for git clone and push operations",
+              args: {
+                name: "option",
+                suggestions: ["https", "ssh"],
+              },
+            },
+            {
+              name: "editor",
+              description: "The text editor program to use for authoring text",
+              args: { name: "editor" },
```
QuiiBz, line 152: Could we suggest `vim`, `nano`, ... here?

QuiiBz (changes requested):
Thanks a lot for this amazing PR!

EliasGcf:
Hi, @QuiiBz, I've already made the changes.

PR #432, opened 2021-08-03
[cmake] Add Spec

Monkeyank:
Feature
#429 
Adds `cmake` spec

### DIFF
--- a/dev/cmake.ts
+++ b/dev/cmake.ts
@@ -1,0 +1,826 @@
+export const completion: Fig.Spec = {
+  name: "cmake",
+  description:
+    "Command-line interface of the cross-platform buildsystem generator CMake",
+  subcommands: [
+    {
+      name: "--build",
+      description:
+        "Project binary directory to be built and must be first. This is required unless a preset is specified",
+      args: {
+        name: "dir",
+        template: "folders",
+        isOptional: true,
+      },
+      options: [
+        {
+          name: "--preset",
+          description: "Use a build preset to specify build options",
+          args: {
+            name: "preset",
+          },
+        },
+        {
+          name: "--list-presets",
+          description:
+            "Use a build preset to specify build options. The project binary directory is inferred from the configurePreset key",
+        },
+        {
+          name: ["--parallel", "-j"],
+          description:
+            "The maximum number of concurrent processes to use when building. If <jobs> is omitted the native build tool's default number is used",
+          args: {
+            name: "jobs",
+            isOptional: true,
+          },
+        },
+        {
+          name: ["--target", "-t"],
+          description:
+            "Build <tgt> instead of the default target. Multiple targets may be given, separated by spaces",
+          args: {
+            name: "target",
+            variadic: true,
+          },
+        },
+        {
+          name: "--config",
+          description:
+            "For multi-configuration tools, choose configuration <cfg>",
+          args: {
+            name: "cfg",
+          },
+        },
+        {
+          name: "--clean-first",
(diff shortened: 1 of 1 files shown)


dev/cmake.ts
```diff
@@ -0,0 +1,826 @@
...
+          name: "--preset",
+          description: "Use a build preset to specify build options",
+          args: {
+            name: "preset",
+          },
+        },
+        {
+          name: "--list-presets",
+          description:
+            "Use a build preset to specify build options. The project binary directory is inferred from the configurePreset key",
```
QuiiBz, line 26: Wrong description?

dev/cmake.ts
```diff
@@ -0,0 +1,826 @@
...
+            },
+            {
+              name: "command",
+            },
+          ],
+        },
+        {
+          name: "compare_files",
+          description:
+            "Checkif file1 is same as file2. If fails are the same, then returns 0, if not returns 1. In case of invalid arguments, it returns 2",
```
QuiiBz, line 175:
```suggestion
            "Check if file1 is the same as file2. If fails are the same, then returns 0, if not returns 1. In case of invalid arguments, it returns 2",
```

dev/cmake.ts
```diff
@@ -0,0 +1,826 @@
...
+          options: [
+            {
+              name: "c",
+              description:
+                "Create a new archive containing the specified files. If used, the pathname argument is mandatory",
+            },
+            {
+              name: "x",
+              description:
+                "Extract to disk from the archive. The <pathname>... argument could be used to extract only selected files or directories. When extracting selected files or directories, you must provide their exact names including the path, as printed by list (-t)",
```
QuiiBz, line 462: Looks like there is an argument (pathname) for this option?

dev/cmake.ts
```diff
@@ -0,0 +1,826 @@
...
+            },
+            {
+              name: "t",
+              description:
+                "List archive contents. The <pathname>... argument could be used to list only selected files or directories",
```
QuiiBz, line 467: Same as above

dev/cmake.ts
```diff
@@ -0,0 +1,826 @@
...
+          args: {
+            name: "pathname",
+            isOptional: true,
+          },
+        },
+        {
+          name: "time",
+          description: "Run command and display elapsed time",
+          args: {
+            name: "command",
```
QuiiBz, line 529: Should we add `isCommand` here?

dev/cmake.ts
```diff
@@ -0,0 +1,826 @@
...
+      description: "Lists the available presets",
+      args: {
+        name: "type",
+        isOptional: true,
+        suggestions: ["configure", "build", "test", "all"],
+      },
+    },
+  ],
+  // Only uncomment if cmake takes an argument
+  // args: {}
```
QuiiBz, line 825: We can remove this

QuiiBz (changes requested):
Cool spec! Only a few things to fix before merging it.

dev/cmake.ts
Monkeyank: This refers to the optional argument passed to tar

dev/cmake.ts
Monkeyank: Same as above

dev/cmake.ts
QuiiBz: I see, thanks for the clarification!

QuiiBz (approved):
Amazing spec, LGTM!

Issue #223, opened 2021-05-25
`loadSpec` doesn't work for root commands


cstrnt:
### Completion Spec
Fig itself

### Description
As described in #212, the `loadSpec` property isn't picked up if its used in the root object of the spec

### Steps to reproduce
```ts
export const completion: Fig.Spec = {
  name: "code-insiders",
  description: "",
  loadSpec: "code",
};

```

### Expected result
The spec should be loaded just like it works with subcommand

### Actual result
No suggestions are shown

Monkeyank:
Closed with #388 

You can now use `import`!

PR #433, opened 2021-08-03
[hub] as wrapper for git

fwesss:
Feature

Many people use Github's [hub CLI](https://github.com/github/hub) tool as a wrapper and alias for git. When typing git, they expect autocompletions but don't get anything due to Fig reading the command as 'hub' and seeing no applicable spec.

hub gives the same exact suggestions as git. There are additional features that hub provides but I have not added them in.

**Additional info:**
Per hub's README: "Some hub features feel best when it's aliased as git. This is not dangerous; your normal git commands will all work. hub merely adds some sugar."

There shouldn't be any issues using the git completion spec with hub.

I would have liked to use loadSpec for this but it does not work for root commands at this time: https://github.com/withfig/autocomplete/issues/223
This could be easily refactored when that issue is fixed.

### DIFF
--- a/dev/hub.ts
+++ b/dev/hub.ts
@@ -1,0 +1,5027 @@
+const filterMessages = (out: string): string => {
+  return out.startsWith("warning:") || out.startsWith("error:")
+    ? out.split("\n").slice(1).join("\n")
+    : out;
+};
+
+const postProcessTrackedFiles: Fig.Generator["postProcess"] = (
+  out,
+  context
+) => {
+  const output = filterMessages(out);
+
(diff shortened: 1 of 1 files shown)


QuiiBz:
Any reason why not importing the git spec instead of copying it?

fwesss:
> Any reason why not importing the git spec instead of copying it?

That's the first thing I tried but couldn't the import to work. I just tried it on v7 and it works as expected. 

I'll submit a new PR so it works under the new version and doesn't have all this redundant code.

PR #436, opened 2021-08-04
added completion spec for ganache-cli

Nutlope:
add a completion spec
no completion spec
added completion spec
**Additional info:**


dev/ganache-cli.ts
```diff
@@ -0,0 +1,172 @@
+export const completionSpec: Fig.Spec = {
```
QuiiBz, line 1: This should now be exported by default

QuiiBz (changes requested):
Nice spec! We just need to update the export to a default one

brendanfalk (approved):

Issue #438, opened 2021-08-04
[autojump] support


fwesss:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue

### What CLI tool does this relate to?

autojump & j

### Which statement makes the most sense?

This completion spec does not exist

### Issue Details

Autojump is a commonly used tool for navigating visited directories: https://github.com/wting/autojump

PR #441, opened 2021-08-04
Support for autojump and alias j

fwesss:
Feature

There is no support for autojump

Re: https://github.com/withfig/autocomplete/issues/438

This adds support for autojump and its alias j

### DIFF
--- a/dev/j.ts
+++ b/dev/j.ts
@@ -1,43 +1,8 @@
+import autojump from "./autojump";
+
 const completionSpec: Fig.Spec = {
   name: "j",
-  description: "a faster way to navigate your filesystem",
-  options: [
-    {
-      name: ["-h", "--help"],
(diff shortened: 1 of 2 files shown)


dev/j.ts
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      name: ["-s", "--stat"],
+      description: "show database entries and their key weights",
+    },
+    {
+      name: ["-v", "--version"],
+      description: "show version information",
+    },
+  ],
```
QuiiBz, line 40: Looks like this is duplicated from `autojump`. I think we can now import and spread it:
```ts
import autojump from './autojump';

  ...autojump,
  name: "j",
  args: {
    ...
  },
};
```

QuiiBz (approved):
LGTM!

PR #442, opened 2021-08-04
[hub] as wrapper for git

fwesss:
Feature

Many people use Github's hub CLI tool as a wrapper and alias for git. When typing git, they expect autocompletions but don't get anything due to Fig reading the command as 'hub' and seeing no applicable spec.

hub gives the same exact suggestions as git. There are additional features that hub provides but I have not added them in.

Additional info:
Per hub's README: "Some hub features feel best when it's aliased as git. This is not dangerous; your normal git commands will all work. hub merely adds some sugar."

There shouldn't be any issues using the git completion spec with hub.

### DIFF
--- a/dev/hub.ts
+++ b/dev/hub.ts
@@ -1,0 +1,5 @@
+import git from "../dev/git";
+
+const completionSpec: Fig.Spec = git;
+
+export default completionSpec;
--- a/dev/hub.ts
+++ b/dev/hub.ts
@@ -1,5 +1,1 @@
-import git from "../dev/git";
-
-const completionSpec: Fig.Spec = git;
-
-export default completionSpec;
+export { default as Hub } from "./git";


dev/hub.ts
```diff
@@ -0,0 +1,5 @@
+import git from "../dev/git";
```
QuiiBz, line 1: Isn't the path simply `./git`?

dev/hub.ts
```diff
@@ -0,0 +1,5 @@
...
+
+const completionSpec: Fig.Spec = git;
+
+export default completionSpec;
```
QuiiBz, line 5: I think we can take advantage of the `export from` syntax here:
```ts
export { default as Hub } from './git';
```

QuiiBz (changes requested):

dev/hub.ts
fwesss: Much cleaner. Just made the change.

QuiiBz (approved):
LGTM!

PR #443, opened 2021-08-05
fix: [TMUX] session name flag '-s' should be in lowercase

Bou:
Bug fix

fig autocompletes this way:
`tmux new -S 'name'`
Which throws an error:
```sh
$ tmux new -S 'session1' 
tmux: unknown option -- S
```

```sh
$ tmux new -s 'session1' 
# Opens the session correctly.
```

**Additional info:**


QuiiBz (approved):
Thank you a lot for this fix!

PR #448, opened 2021-08-08
feat: add `deprecated` flag to `BaseSuggestion` type

fedeci:
It introduces a new `deprecated` flag to `BaseSuggestion`. 
I have added some examples to clarify the use case.

### DIFF
--- a/dev/docker-compose.ts
+++ b/dev/docker-compose.ts
@@ -200,14 +200,18 @@ const completionSpec: Fig.Spec = {
     },
     {
       args: {
-        name: "(never|always|auto)",
+        name: "print",
+        suggestions: ["never", "always", "auto"],
       },
-      description: "Control when to print ANSI control characters",
+      description: "Control when to print ANSI control characters.",
       name: ["--ansi"],
     },
     {
-      description: "Do not print ANSI control characters (DEPRECATED)",
+      description: "Do not print ANSI control characters.",
       name: ["--no-ansi"],
+      deprecated: {
+        alternative: "docker-compose --ansi never",
+      },
     },
     {
       description: "Print version and exit",
--- a/schemas/fig.d.ts
+++ b/schemas/fig.d.ts
@@ -105,6 +105,30 @@ declare namespace Fig {
      * The "-" suggestion is hidden in the `cd` spec. You will only see it if you type cd -
      */
     hidden?: boolean;
+
+    /**
+     *
+     * Specifies whether a suggestion is deprecated.
+     * It is possible to optionally specify a deprecation message and an alternative command to run.
+     *
+     * `deprecated: true` and `deprecated: { }` will just display the option as deprecated.
+     *
+     */
+    deprecated?:
+      | boolean
+      | {
+          /**
+           * Provide an alternative to the deprecated command.
+           *
+           * @example
+           * alternative: 'npm install'
+           */
+          alternative?: string;
+          /**
+           * Provide a warning or additional info about the deprecation.
+           */
+          message?: string;
+        };
   }
 
   export interface Suggestion extends BaseSuggestion {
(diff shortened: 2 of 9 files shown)


QuiiBz:
Thanks for this idea! I think it's a good one, but we will need to implement it in the autocomplete-engine itself first. cc @mschrage @brendanfalk

fedeci:
Yep, it definitely needs.

schemas/fig.d.ts
```diff
@@ -105,6 +105,30 @@ declare namespace Fig {
...
+    deprecated?:
+      | boolean
+      | {
+          /**
+           * Provide an alternative to the deprecated command.
+           *
+           * @example
+           * alternative: 'npm install'
+           */
+          alternative?: string;
```
sullivan-sean, line 126: I like this idea a lot :) One thing that might help Fig be even more useful here in the future is if we type this as
```suggestion
          alternative?: Omit<BaseSuggestion, "deprecated">;
```
Which adds a little more structure/modularity to the alternative so we could (in the future) add an option in settings to replace deprecated commands and flags with their alternative if they exist, e.g. in the `docker-compose` change you made we could just insert `--ansi never` when the user selects the `--no-ansi` suggestion. This probably wouldn't be great default behavior in case you _really_ want to insert the original suggestion, but could be a nice opt in feature, and makes the spec standard a little more information rich.

fedeci: A possible edge case where that structure is not applicable is for example if an option has moved to a subcommand:
e.g.
`foo bar --remove` -> `foo removebar`
I recognize it is a strange one, but
`alternative?: Omit<BaseSuggestion, "deprecated"> | string` seems probably better. If alt is a string just a warning message is displayed.

sullivan-sean: If the intent of the `alternative` string is to offer replacement text, I think anything we could accomplish with a string could be accomplished with the `insertValue` in `BaseSuggestion`, as this is already a string, right? If it is purely informational, then we could use the `description` field of the `BaseSuggestion` but ultimately this decision could be left up to whoever is implementing the spec for that particular deprecated object.

fedeci: It makes more sense now!

mschrage:
Hey @fedeci! Sorry for the delay on this! Can you solve these merge conflicts and then I'll get these merged ASAP!

fedeci:
Sure, I will also finish updating the specs.

fedeci:
I'll follow up with a PR updating the other options.

QuiiBz:
The feature isn't yet implemented into Fig's engine, so we will need to make and deploy it before merging this PR.

dev/docker-compose.ts
```diff
@@ -199,14 +199,18 @@ const completionSpec: Fig.Spec = {
...
       },
       description: "Control when to print ANSI control characters",
       name: "--ansi",
     },
     {
-      description: "Do not print ANSI control characters (DEPRECATED)",
+      description: "Do not print ANSI control characters",
       name: "--no-ansi",
+      deprecated: {
+        alternative: { insertValue: "--ansi never" },
```
fedeci, line 212: @sullivan-sean has explained this cannot handle the case where a command an option has been replaced by a higher command, but it is a super exclusive case and we may just add a message for it
`foo install --all -> foo install-all`

sullivan-sean:
This looks good to me with one last tweak to the typing. I think the simplest type for deprecated would actually be

```typescript
type BaseSuggestion = {
   ...,
   deprecated?: boolean | Omit<BaseSuggestion, "deprecated">;
}
```

If a `BaseSuggestion` is passed to deprecated, we can use the `insertValue` and `description` fields to achieve what you have called `alternative` and `message`. 

With that I'd be happy to merge this in!

fedeci:
@sullivan-sean I tried what you say some times ago, but it looked somewhat strange stylistically so I decided to keep the current solution. I'll test it again tomorrow.

PR #450, opened 2021-08-08
chore(eslint): add `no-missing-default-export` rule

fedeci:
Based on #330


dev/hub.ts
```diff
@@ -1 +1,2 @@
-export { default as Hub } from "./git";
+import git from "./git";
+export default git;
```
fedeci, line 2:
```suggestion
export { default } from './git';
```

scripts/eslint-plugin-fig-linter/rules/no-missing-default-export.js
```diff
@@ -0,0 +1,25 @@
...
+        hasExport = true;
+      },
+      'ExportSpecifier[exported.name="default"]'(node) {
+        hasExport = true;
+      },
+      "Program:exit"(node) {
+        if (!hasExport) {
+          context.report({
+            node,
+            message: "File must contain one default export.",
```
QuiiBz, line 19: The message can be a bit more expressive:
```suggestion
            message: "File must export default the completion spec object",
```

QuiiBz (changes requested):
Thanks, this rule looks great!

QuiiBz (approved):
Thanks a lot for this one!

PR #452, opened 2021-08-09
Pyenv versions listing

lsh:
Extends the `pyenv global` autocomplete.

`pyenv global` currently simply gives the hint that the argument should be a version already installed.

The new functionality is that `pyenv global` lists from `pyenv versions` (already installed instances).


Nutlope:
LGTM! Could you reset the package.json and commit that?

PR #453, opened 2021-08-09
Migrate to v7 spec format

sullivan-sean:

Migrates spec to new format, with
- More consistent names
  - Boolean `variadic` is now `isVariadic`
  - Boolean `required` is now `isRequired`
  - `exclusive` is now `exclusiveOn` to match `dependsOn`
- More meaningful naming for `filterTerm` (now `getQueryTerm`) that clarifies that when a string is passed, it is as a delimiter to _get_ the query term, and not the query term itself
- Parser directives subheading under subcommand
  - Allows POSIX noncompliance at subcommand granularity
  - Sets up a section for future parsing related options at the subcommand level
- Adds shell context for custom generators
  - Renames "context" to "tokens" for generators to disambiguate shell context from the tokens that the user has typed

**Additional info:**


mschrage (approved):
Looks good! ✨

PR #457, opened 2021-08-10
[git] branch -d & -D arg is variadic

fwesss:
Per `man git-branch`: the -d and -D options take a branch arg that is variadic. You can also use in conjunction with -r to delete remote branches.

This PR makes those arguments variadic and also adds a generator that generates either local or remote branches depending on whether -r is being used.


QuiiBz (approved):
LGTM!

PR #461, opened 2021-08-12
Add pageres spec

Monkeyank:
Feature
Adds `pageres` spec


dev/pageres.ts
```diff
@@ -0,0 +1,133 @@
...
+  name: "pageres",
+  description: "Capture website screenshots",
+  options: [
+    {
+      name: ["--verbose", "-v"],
+      description: "Verbose output to see errors if you need to troubleshoot",
+    },
+    {
+      name: ["--crop", "-c"],
+      description: "Crop to the set height",
```
QuiiBz, line 11: Does this option take an argument?

dev/pageres.ts
```diff
@@ -0,0 +1,133 @@
...
+    {
+      name: "--hide",
+      description: "Hide DOM element. Can be set multiple times",
+      args: {
+        name: "element",
+      },
+    },
+    {
+      name: "--no-crop",
+      description: "Override a global crop option within a group",
```
QuiiBz, line 49: This option can exclude the `--crop` one

dev/pageres.ts
```diff
@@ -0,0 +1,133 @@
...
+    },
+    {
+      name: "--css",
+      description:
+        "Apply custom CSS to the webpage. Specify some CSS or the path to a CSS file",
+      args: {
+        name: "string",
```
QuiiBz, line 56: Looks like we can have a template of `filepaths` here

QuiiBz (changes requested):

dev/pageres.ts
Monkeyank: This doesn't take an argument

QuiiBz (approved):
LGTM!

PR #464, opened 2021-08-12
Rewrite Pyenv versionList generator in a functional style

lsh:
No new changes are made to the functionality of this code, instead it is refactored to what could be considered a more contemporary (and readable) style. This code would also match the style introduced in #452.


QuiiBz (approved):
So much cleaner, thanks! LGTM

PR #465, opened 2021-08-13
Reference revamp: Edited descriptions

Nutlope:
Edited descriptions for the references section

### DIFF
--- a/schemas/fig.d.ts
+++ b/schemas/fig.d.ts
@@ -165,7 +165,7 @@ declare namespace Fig {
      *
      * If your CLI tool takes another CLI command (e.g. time , builtin) or a script
      * (e.g. python, node) and you would like Fig to continue to provide completions for this
-     * script, see `isCommand` and `isScript` in @link https://fig.io/docs/reference/arg | Args}.
+     * script, see `isCommand` and `isScript` in @link https://fig.io/docs/reference/arg | Arg}.
      */
     loadSpec?: string;
     /**
@@ -193,7 +193,7 @@ declare namespace Fig {
      *
      *
      * @example
-     * `-aux` is parsed as a single flag when `flagsArePosixNoncompliant: true`. Normally, this would be chained and parsed as `-a -u -x` if `flagsArePosixNoncompliant` is not set to true.
+     * `-work` from the go spec is parsed as a single flag when `parserDirectives.flagsArePosixNoncompliant` is set to true. Normally, this would be chained and parsed as `-w -o -r -k` if `flagsArePosixNoncompliant` is not set to true.
      */
     parserDirectives?: {
       flagsArePosixNoncompliant?: boolean;
@@ -226,7 +226,7 @@ declare namespace Fig {
      * Signalling that an option is required currently doesn't do anything. However, Fig will handle it in the future.
      *
      * @example
-     * The `"-m"` option of git commit is required
+     * The `-m` option of git commit is required
      *
      */
     isRequired?: boolean;
@@ -244,7 +244,7 @@ declare namespace Fig {
     exclusiveOn?: string[];
     /**
      *
-     * Signals whether an option depends other options. This is definedas an array of strings of the option names.
+     * Signals whether an option depends other options. This is defined as an array of strings of the option names.
      * The default behavior is that an option does NOT depend on any other options.
      * If the user has an unmet dependency for a flag they've already typed, this dependency will have boosted priority in the suggestion list.
      *
@@ -330,14 +330,14 @@ declare namespace Fig {
      */
     isCommand?: boolean;
     /**
-     * The same as the `isCommand` prop, except, you specify a string to prepend to what the user inputs and fig will load the completion spec accordingly. if isModule: "python/", Fig would load up the python/USER_INPUT.js completion spec from the `~/.fig/autocomplete` folder.
+     * The same as the `isCommand` prop, except you specify a string to prepend to what the user inputs and fig will load the completion spec accordingly. if isModule: "python/", Fig would load up the python/USER_INPUT.js completion spec from the `~/.fig/autocomplete` folder.
      *
      * @example
      * For `python -m`, the user can input a specific module such as http.server. Each module is effectively a mini CLI tool that should have its own completions. Therefore the argument object for -m has `isModule: "python/"`. Whatever the modules user inputs, Fig will look under the `~/.fig/autocomplete/python/` directory for completion spec.
      */
     isModule?: string;
     /**
-     * The same as the `isCommand` prop ,except Fig will look for a completion spec in a .fig folder in the user's current working directory.
+     * The same as the `isCommand` prop, except Fig will look for a completion spec in a .fig folder in the user's current working directory.
      *
      * @example
      * `python` take one argument which is a `.py` file. If I have a `main.py` file on my desktop and my current working directory is my desktop, if I type `python main.py` Fig will look for a completion spec in `~/Desktop/.fig/main.py.js`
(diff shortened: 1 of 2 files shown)


schemas/fig.d.ts
```diff
@@ -35,32 +35,31 @@ declare namespace Fig {
...
     /**
-     * The value that's inserted into the terminal when a user presses enter/tab or clicks on a menu item.
-     * You can optionally specify {cursor} in the string and Fig will automatically place the cursor there after insert.
+     * The value that's inserted into the terminal when a user presses enter/tab or clicks on a menu item. You can use `\n` to insert a newline or `\b` to insert a backspace.
+     * You can also optionally specify {cursor} in the string and Fig will automatically place the cursor there after insert.
      * The default is the name prop.
      *
      * @example
-     * For `git commit` the `-m` option has an insert value of `-m '{cursor}'`
+     * For `git commit` subcommand, the `-m` option has an insert value of `-m '{cursor}'`
```
sullivan-sean, line 52:
```suggestion
     * For the `git commit` subcommand, the `-m` option has an insert value of `-m '{cursor}'`
```

schemas/fig.d.ts
```diff
@@ -147,25 +147,25 @@ declare namespace Fig {
...
+     * `aws-s3` refers to the `~/.fig/autocomplete/aws-s3` spec.
      *
      * When is this used? The aws spec is so large that it is slow to load. It needs to be
      * broken up into a separate spec for each subcommand.
      *
-     * If your CLI tool takes another CLI command (e.g. time , builtin... ) or a script
+     * If your CLI tool takes another CLI command (e.g. time , builtin) or a script
      * (e.g. python, node) and you would like Fig to continue to provide completions for this
-     * script, see `isCommand` and `isScript` in {@link https://fig.io/docs/reference/arg | Arg}.
+     * script, see `isCommand` and `isScript` in @link https://fig.io/docs/reference/arg | Args}.
```
sullivan-sean, line 168: I think singular `Arg` here still makes sense

schemas/fig.d.ts
```diff
@@ -188,11 +188,12 @@ declare namespace Fig {
...
      *
-     * Currently, the only parser option is flagsArePosixNoncompliant, which
-     * allows options to have multiple characters even if they only have one hyphen.
+     * Currently, the only parser option that exists is flagsArePosixNoncompliant. This
+     * allows options to have multiple characters even if they only have one hyphen when set to `true`.
+     *
      *
      * @example
-     * `-longflag` is parsed as a single flag when `flagsArePosixNoncompliant: true`
+     * `-aux` is parsed as a single flag when `flagsArePosixNoncompliant: true`. Normally, this would be chained and parsed as `-a -u -x` if `flagsArePosixNoncompliant` is not set to true.
```
sullivan-sean, line 196: When I think `-aux` I think `ps` which is POSIX compliant, maybe it would make sense to use an example from a go cli with `-help` as a clear example of when you'd want to set this to true

schemas/fig.d.ts
```diff
@@ -246,13 +244,13 @@ declare namespace Fig {
     exclusiveOn?: string[];
     /**
      *
-     * Signals whether an option depends other options. Define this as an array of strings of the option names.
-     * The default is an option does NOT depend on any other options.
+     * Signals whether an option depends other options. This is definedas an array of strings of the option names.
```
sullivan-sean, line 247:
```suggestion
     * Signals whether an option depends other options. This is defined as an array of strings of the option names.
```

schemas/fig.d.ts
```diff
@@ -324,37 +323,37 @@ declare namespace Fig {
...
+     * The same as the `isCommand` prop, except, you specify a string to prepend to what the user inputs and fig will load the completion spec accordingly. if isModule: "python/", Fig would load up the python/USER_INPUT.js completion spec from the `~/.fig/autocomplete` folder.
      *
      * @example
      * For `python -m`, the user can input a specific module such as http.server. Each module is effectively a mini CLI tool that should have its own completions. Therefore the argument object for -m has `isModule: "python/"`. Whatever the modules user inputs, Fig will look under the `~/.fig/autocomplete/python/` directory for completion spec.
      */
     isModule?: string;
     /**
-     * Exactly the same as the `isCommand` prop except Fig will look for a completion spec in a .fig folder in the user's current working directory.
+     * The same as the `isCommand` prop ,except Fig will look for a completion spec in a .fig folder in the user's current working directory.
```
sullivan-sean, line 340:
```suggestion
     * The same as the `isCommand` prop, except Fig will look for a completion spec in a .fig folder in the user's current working directory.
```

schemas/fig.d.ts
```diff
@@ -324,37 +323,37 @@ declare namespace Fig {
...
+     * See our docs for more on this @link https://fig.io/docs/tutorials/building-internal-clis | Fig for Teams}
      */
     isScript?: boolean;
 
     /**
      * This will debounce every keystroke event for this particular arg. If there are no keystroke events after 100ms, Fig will execute all the generators in this arg and return the suggestions.
      *
      * @example
-     * NPM install and pip install send debounced network requests after inactive typing from users.
+     * Npm install and pip install send debounced network requests after inactive typing from users.
```
sullivan-sean, line 352:
```suggestion
     * `npm install` and `pip install` send debounced network requests after inactive typing from users.
```

schemas/fig.d.ts
```diff
@@ -383,26 +382,24 @@ declare namespace Fig {
      * In order to generate contextual suggestions for arguments, Fig lets you execute a shell command on the users local device as if it were done in their current working directory.
      * You can either specify
      * 1. a string to be executed (like `ls` or `git branch`)
-     * 2. a function to generate the string to be executed. The function takes in an array of tokens of the user input and should output a string. You use a function when the script you run is dependent upon one of the tokens the user has already input (for instance an app name, a kubernetes tokens etc)
+     * 2. a function to generate the string to be executed. The function takes in an array of tokens of the user input and should output a string. You use a function when the script you run is dependent upon one of the tokens the user has already input (for instance an app name, a kubernetes token etc)
```
sullivan-sean, line 385:
```suggestion
     * 2. a function to generate the string to be executed. The function takes in an array of tokens of the user input and should output a string. You use a function when the script you run is dependent upon one of the tokens the user has already input (for instance an app name, a Kubernetes token etc.)
```

sullivan-sean (approved):
Looks great, especially nice that we're adding links between docs when it makes sense, which I think will improve the DX a lot :)

Left a few nits and only looked at things that came up in the diff, but trust you went through the rest more thoroughly

schemas/fig.d.ts
brendanfalk: agree. also, perhaps have `parserDirectives.flagsArePosixNoncompliant` to be more clear here

schemas/fig.d.ts
```diff
@@ -223,19 +221,19 @@ declare namespace Fig {
...
      *
-     * Signals whether an option is required. The default is an option is NOT required.
+     * Signals whether an option is required. The default value is false, meaning an option is NOT required.
      *
-     * Currently, signalling that an option is required doesn't do anything, however, Fig will handle it in the future
+     * Signalling that an option is required currently doesn't do anything. However, Fig will handle it in the future.
      *
      * @example
-     * The "-m" option of git commit is required
+     * The `"-m"` option of git commit is required
```
brendanfalk, line 229: just make it `-m` not `"-m"`

brendanfalk (commented):
Agree with your changes + all of sean's edits. My last comment would be when something can be a function, we don't have a standard for how we describe the input and output params. Sometimes we use bullets, sometimes we use a string. Descriptions are always random. It should be much more clear and standard.

Look into tsdoc vs jsdoc and speak to tim. But ideally we follow one standard, stick to it, and then make sure the parser we built that takes this standard and converts it into docs understands this standard

mschrage:
Looks good to me!

mschrage:
You just need to resolve small merge conflict with @sullivan-sean and then you're good to merge!

PR #467, opened 2021-08-13
Completed lsd spec

fwesss:
lsd was previously a wrapper for ls but they don't have the same options or arguments.

Completed spec to match lsd --help

### DIFF
--- a/dev/lsd.ts
+++ b/dev/lsd.ts
@@ -1,6 +1,6 @@
 const completionSpec: Fig.Spec = {
   name: "lsd",
-  description: "as ls command with a lot of pretty colors and some other stuff",
+  description: "an ls command with a lot of pretty colors and some other stuff",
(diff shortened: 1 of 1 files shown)


dev/lsd.ts
```diff
@@ -1,233 +1,197 @@
 const completionSpec: Fig.Spec = {
   name: "lsd",
-  description: "list directory contents",
+  description: "as ls command with a lot of pretty colors and some other stuff",
```
Monkeyank, line 3:
```suggestion
  description: "an ls command with a lot of pretty colors and some other stuff",
```

Monkeyank (changes requested):
Small type, otherwise LGTM!

Nutlope (approved):

Nutlope:
LGTM!

PR #468, opened 2021-08-13
Add fnm spec

noftaly:
Feature

Adds a spec for the [`fnm` node version manager](https://github.com/Schniz/fnm)

If you want to compare it to the fnm's help command, it is available [on their github](https://github.com/Schniz/fnm/blob/master/docs/commands.md).

**What is the new behavior (if this is a feature change)?** /

**Additional info:** 
As this is my first contribution to Fig, don't refrain from suggesting how the spec could be improved or what I should maybe do differently. I'm happy to discuss! :)

### DIFF
--- a/dev/fnm.ts
+++ b/dev/fnm.ts
@@ -1,0 +1,328 @@
+interface NodejsVersion {
+  major: number;
+  minor: number;
+  patch: number;
+  original: string;
+  ltsName: string;
+}
+
+// Generators
+const versionGenerator: Fig.Generator = {
+  script: "fnm ls",
+  postProcess: function (out) {
+    return out
+      .split("\n")
+      .reverse()
+      .map((line) => ({
+        name: line.slice(2),
+        description: `Node.js ${line.slice(2)}`,
+      }));
+  },
+};
+
+const NODE_VERSION_REGEX = /v(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)(?: \((?<ltsName>\w+)\))?/iu;
+const parseNodejsVersion = (raw: string): NodejsVersion => {
+  const { major, minor, patch, ltsName } =
+    NODE_VERSION_REGEX.exec(raw)?.groups ?? {};
+  return {
+    major: Number(major),
+    minor: Number(minor),
+    patch: Number(patch),
+    original: raw,
+    ltsName,
+  };
+};
+const uniqBy = <T = unknown>(arr: T[], callback: (a: T, b: T) => boolean) =>
+  arr.reduce((acc, v) => {
+    if (!acc.some((x) => callback(v, x))) acc.push(v);
+    return acc;
+  }, []);
+
+/**
+ * This generator will output the following Node.js' versions:
+ * - Every version of the latest LTS;
+ * - Latest release of all named LTS;
+ * - Latest release of the beta right before the last LTS, or the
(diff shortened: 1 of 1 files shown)


fedeci:
Splitting everything makes useless some linting rules, I'll have to update those.

noftaly:
> Splitting everything makes useless some linting rules, I'll have to update those.

Ah! 😄 I started with the nvm completions to build this one, and it split everything like this

fedeci:
In fact it is definitely cleaner! My fault :)

dev/fnm.ts
```diff
@@ -0,0 +1,234 @@
...
+
+const installIfMissing: Fig.Option = {
+  name: "--install-if-missing",
+  description: "Install the version if it isn't installed yet",
+};
+
+const shell: Fig.Option = {
+  name: "shell",
+  description: "The shell syntax to use. Infers when missing",
+  args: [
```
QuiiBz, line 29: No need to englobe this in an array when there is only one `arg`. Same for underneath

QuiiBz (changes requested):
Thanks a lot for this PR! Almost everything looks perfect.

I think it should be possible to add custom generators for  `install` and `uninstall` (executing `ls-remote` and `ls` respectively in a generator and mapping the result to suggestions). Here is the [documentation](https://fig.io/docs/reference/generator) for it, tell us if you need any help!

dev/fnm.ts
noftaly: sure! I'll commit that. Thanks \:)

noftaly:
> I think it should be possible to add custom generators for install and uninstall (executing ls-remote and ls respectively in a generator and mapping the result to suggestions). Here is the documentation for it, tell us if you need any help!

Great idea! I'll try to add that :)

noftaly:
@QuiiBz I've added it for every command that required a `version` argument.

The only thing is, there are a *lot* of versions (for the remoteVersions). Do you think we should do some filtering?
I thought of maybe showing the 4 latest even majors and the last 2 odd majors (so v10, v12, v14, v15, v16, v17) (because in Node, odd majors are beta, and even majors are LTS).
We could also show the latest minor every time, so only v16.6.2 for example, not v16.6.1 and v16.6.0.

Wdyt?

noftaly:
Quick update: I'll aim for this suggestions:
- Latest patch of all the minor of the last LTS
- All named LTS (latest version)
- Latest patch of latest minor of 1 or 2 latest Beta
- And everything else sorted, but after those (so you still get autocomplete, they are just not visible on top)

Example with Node.js' version of 2021/08/21:
We will have:
v16.7.0
v16.6.2
v16.5.0
v16.4.2
v16.3.0
v16.2.0
v16.1.0
v16.0.0
v15.14.0
v14.17.5 (Fermium)
v12.22.5 (Erbium)
v10.24.1 (Dubnium)
v8.17.0 (Carbon)
v6.17.1 (Boron)
v4.9.1 (Argon)
everything else

If there was a v17, then there will also be latest v17 on top. If there was a v18, v15 would disappear because it's a beta that's too old, but v17 would stay.

I hope it makes sense 😄 I think it's a great compromise to show the latest versions (most commonly used), but also provide completion for every other version.

Nutlope:
@noftaly this looks awesome, nice work!! LGTM :)

Nutlope (approved):

PR #473, opened 2021-08-14
eleventy-cli

yavko:
Adds Eleventy Cli Support
No support
support
**Additional info:**
4th contribution

### DIFF
--- a/dev/eleventy.ts
+++ b/dev/eleventy.ts
@@ -43,15 +43,16 @@ const completionSpec: Fig.Spec = {
       },
     },
     {
-      description: "Run web server on port (default 8080) and watch them too.",
+      description: "Run web server on port and watch them too.",
       name: "--serve",
+      dependsOn: "--port",
     },
     {
-      description: "Write HTML output to this folder.",
+      description: "Run Web Server on specified port.",
       name: "--port",
-      dependsOn: "--serve",
       args: {
         name: "port",
+        default: "8080",
       },
     },
     {
@@ -59,6 +60,7 @@ const completionSpec: Fig.Spec = {
       name: "--formats",
       args: {
         name: "Template Names Seperated by ,",
+        suggestions: ["md", "html", "ejs"],
       },
     },
     {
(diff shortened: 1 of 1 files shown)


dev/eleventy.ts
```diff
@@ -0,0 +1,82 @@
...
+      dependsOn: "--serve",
+      args: {
+        name: "port",
+      },
+    },
+    {
+      description: "Whitelist only certain template types.",
+      name: "--formats",
+      args: {
+        name: "Template Names Seperated by ,",
```
Monkeyank, line 61: I think we can provide a list of suggestions, such as `md,html,ejs`

dev/eleventy.ts
```diff
@@ -0,0 +1,82 @@
...
+    {
+      description: "Run web server on port (default 8080) and watch them too.",
+      name: "--serve",
+    },
+    {
+      description: "Write HTML output to this folder.",
+      name: "--port",
+      dependsOn: "--serve",
+      args: {
+        name: "port",
```
Monkeyank, line 54: Maybe add a `default: "8080"`?

Monkeyank (changes requested):

dev/eleventy.ts
yavko: ok

dev/eleventy.ts
yavko: ok

dev/eleventy.ts
yavko: Idk how tho, I read the docs and i still don't get how

Monkeyank: You can add a prop underneath args. Here's an example:

```
args: {
     name: "Template Names Separated by,",
     suggestions: ["md", "html", "eje"]
}
```

yavko: ok thanks

fedeci: However that does not allow for comma separated values, right?

yavko: idk

QuiiBz (approved):
LGTM, thanks a lot for this PR!

yavko:
YAYAYA PR 4

PR #478, opened 2021-08-14
Autocomplete spec for ionic cli

EthanOrlander:
Adds autocomplete spec for Ionic CLI.

Working on the generator over here:[https://github.com/EthanOrlander/ionic-fig-generator]( https://github.com/EthanOrlander/ionic-fig-generator)
Hooking into the Ionic CLI using `ionic help --json` and programatically generating the spec.
- Turns out `ionic help --json` is incomplete and incorrect, so there's a bit of manual post-processing work 

Please let me know if there's any issues thus far!

### DIFF
--- a/dev/ionic.ts
+++ b/dev/ionic.ts
@@ -29,6 +29,12 @@ const completionSpec = {
         {
           name: ["--configuration", "-c"],
           description: "Specify the configuration to use",
+          args: [
+            {
+              name: "conf",
+              template: ["filepaths", "folders"],
+            },
+          ],
         },
         {
           name: ["--source-map"],
@@ -255,6 +261,7 @@ const completionSpec = {
           args: [
             {
               name: "conf",
+              template: ["filepaths", "folders"],
             },
           ],
         },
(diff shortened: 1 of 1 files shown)


EthanOrlander:
Also anyone is welcome to fill in those args or icons :)

EthanOrlander:
@Nutlope If I want to limit filepaths suggestions to specific extensions, is there an example you can point me to? I can do it myself, but I'd rather use what you've already approved. Specifically, I want to limit suggested files to .ts if TypeScript Capacitor project or .json if JavaScript Capacitor project.

Nutlope:
@EthanOrlander this is looking great so far! FYI, feel free to submit the skeleton as a PR the way it is now (to get feedback on it + get it merged) then submit another one with generators. You may learn a few things that'll be helpful. 

For your question about limiting filepaths to specific extensions, I actually can't think of an example off the top of my head. Probably a generator that runs `ls` as the script and has uses `.filter` in the postprocess function to only return the extensions you want.

EthanOrlander:
@Nutlope Sounds good! So, I'll leave this PR as-is for review, and I'll make another where I'll add in the generators.

QuiiBz (approved):
Thanks a lot for this huge spec! Feel free to open another PR if you want to add generators :)

PR #451, opened 2021-08-09
style(eslint): add `no-useless-arrays` rule

fedeci:
With this PR we ensure that when `name` properties of subcommands or options have just one value they are not wrapped in array.

PR #479, opened 2021-08-14
Add Spec for N

sohcah:
Feature

No autocompletion for N (https://github.com/tj/n)

Full autocompletion for N

**Additional info:**
Includes dynamic suggestions for versions from `n lsr --all`

### DIFF
--- a/dev/n.ts
+++ b/dev/n.ts
@@ -25,13 +25,11 @@ const versionArg: Fig.Arg = {
     {
       script: "n lsr --all",
       postProcess: function (out) {
-        const set = new Set()
-        const versions = out
-          .split("\n")
-          .slice(1)
+        const set = new Set<string>();
+        const versions = out.split("\n").slice(1);
         for (const version of versions) {
           set.add(version); // 16.1.2
-          const splitted = version.split(".")
+          const splitted = version.split(".");
           set.add(splitted[0] + "." + splitted[1]); // 16.1
           set.add(splitted[0]); // 16
         }
--- a/dev/n.ts
+++ b/dev/n.ts
@@ -153,21 +153,33 @@ const completionSpec: Fig.Spec = {
     {
       name: ["-a", "--arch"],
       description: "Override system architecture",
+      args: {
+        name: "Architecture",
+        suggestions: [
+          { name: ["x64", "x86_64"] },
+          { name: ["arm64", "aarch64", "armv8l"] },
+          { name: ["x86", "i386", "i686"] },
+          { name: ["armv6l", "armv7l"] },
+        ],
+      },
     },
     {
       name: ["--insecure"],
       description:
         "Turn off certificate checking for https requests (may be needed from behind a proxy server)",
+      isDangerous: true,
     },
     {
       name: ["--use-xz"],
       description:
         "Override automatic detection of xz support and enable use of xz compressed node downloads.",
+      exclusiveOn: ["--no-use-xz"],
     },
     {
       name: ["--no-use-xz"],
       description:
         "Override automatic detection of xz support and disable use of xz compressed node downloads.",
+      exclusiveOn: ["--use-xz"],
     },
   ],
 };
--- a/dev/n.ts
+++ b/dev/n.ts
@@ -1,0 +1,175 @@
+import node from "./node";
+
+const versionArg: Fig.Arg = {
+  name: "version",
+  suggestions: [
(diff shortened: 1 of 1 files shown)


dev/n.ts
```diff
@@ -0,0 +1,175 @@
...
+      description: "Install a Node.js version",
+      args: [versionArg],
+    },
+    {
+      name: ["rm", "–"],
+      description: "Remove a Node.js version",
+      args: [variadicVersionArg],
+    },
+    {
+      name: ["prune"],
```
fedeci, line 70: nit: see #451 
```suggestion
      name: "prune",
```

dev/n.ts
```diff
@@ -0,0 +1,175 @@
...
+};
+
+const completionSpec: Fig.Spec = {
+  name: "n",
+  description: "Node version management",
+  subcommands: [
+    {
+      name: ["i", "install"],
+      description: "Install a Node.js version",
+      args: [versionArg],
```
fedeci, line 62: nit: both for this and the others
```suggestion
      args: versionArg,
```

dev/n.ts
```diff
@@ -0,0 +1,175 @@
...
+      name: ["ls", "list"],
+      description: "Output downloaded versions",
+    },
+    {
+      name: ["lsr", "ls-remote", "list-remote"],
+      description: "Output matching versions available for download",
+      args: [versionArg],
+      options: [
+        {
+          name: ["--all"],
```
fedeci, line 92: ditto:
```suggestion
          name: "--all",
```

dev/n.ts
```diff
@@ -0,0 +1,175 @@
...
+          .map((i) =>
+            i.split(".").map((_, i, a) => a.slice(0, i + 1).join("."))
+          )
+          .flat();
+        return [...new Set(versions)].map((version) => {
+          return {
+            name: [version, `v${version}`],
+            description: `Node.js ${version}`,
+          };
+        });
```
fedeci, line 40: perf: using a set from the beginning and avoiding redundant operations on strings is a good boost.
```suggestion
        const set = new Set()
        const versions = out
          .split("\n")
          .slice(1)
        for (const version of versions) {
          set.add(version); // 16.1.2
          const splitted = version.split(".")
          set.add(splitted[0] + "." + splitted[1]); // 16.1
          set.add(splitted[0]); // 16
        }
        return Array.from(set).map((version) => {
          return {
            name: [version, `v${version}`],
            description: `Node.js ${version}`,
          };
        });
```

sohcah: Very good idea to use a set from the start
 I had originally not implemented the splitting at all, so didn't use a Set... I had just quickly hacked that in afterwards, which wasn't the wisest idea.

dev/n.ts
```diff
@@ -0,0 +1,178 @@
...
+      description:
+        'Disable curl output. Disable log messages processing "auto" and "engine" labels.',
+    },
+    {
+      name: ["-d", "--download"],
+      description: "Download only",
+    },
+    {
+      name: ["-a", "--arch"],
+      description: "Override system architecture",
```
QuiiBz, line 158: I believe that this option takes an argument

dev/n.ts
```diff
@@ -0,0 +1,178 @@
...
+    },
+    {
+      name: ["--insecure"],
+      description:
```
QuiiBz, line 162: I think we can add [isDangerous](https://fig.io/docs/reference/option#isdangerous) here!

dev/n.ts
```diff
@@ -0,0 +1,178 @@
...
+    },
+    {
+      name: ["--use-xz"],
+      description:
+        "Override automatic detection of xz support and enable use of xz compressed node downloads.",
+    },
+    {
+      name: ["--no-use-xz"],
+      description:
+        "Override automatic detection of xz support and disable use of xz compressed node downloads.",
```
QuiiBz, line 173: Looks like we can use [exclusiveOn](https://fig.io/docs/reference/option#exclusiveon) for those two options.

QuiiBz (changes requested):
Thanks a lot for this PR! `n` looks really cool, I noticed few things to resolve to improve the quality of the spec before merging it.

QuiiBz:
Also, could you run the `lint:fix` script?

QuiiBz (approved):
Thanks a lot for this amazing spec!

PR #481, opened 2021-08-15
Update grep spec

Monkeyank:

**Additional info:**


dev/grep.ts
```diff
@@ -13,107 +13,297 @@ const completionSpec: Fig.Spec = {
...
+    {
+      name: ["-L", "--files-without-match"],
+      description:
+        "Suppress normal output; instead print the name of each input file from which no output would normally have been printed. The scanning will stop on the first match",
+    },
+    {
+      name: ["-l", "--files-with-matches"],
+      description:
+        "Suppress normal output; instead print the name of each input file from which output would normally have been printed. The scanning will stop on the first match. (-l is specified by POSIX .)",
+    },
```
QuiiBz, line 87: We can add [exclusiveOn](https://fig.io/docs/reference/option#exclusiveon) for those options!

dev/grep.ts
```diff
@@ -13,107 +13,297 @@ const completionSpec: Fig.Spec = {
...
+      name: ["-A", "--after-context"],
+      description: "Print num lines of trailing context after each match",
+      args: {
+        name: "NUM",
+      },
+    },
+    {
+      name: ["-B", "--before-context"],
+      description:
+        "Print num lines of leading context before each match.  See also the -A and -C options",
```
QuiiBz, line 164:
```suggestion
        "Print num lines of leading context before each match. See also the -A and -C options",
```

dev/grep.ts
```diff
@@ -13,107 +13,297 @@ const completionSpec: Fig.Spec = {
...
+      description:
+        "Print NUM lines of output context. Places a line containing a group separator (--) between contiguous groups of matches. With the -o or --only-matching option, this has no effect and a warning is given",
+      args: {
+        name: "NUM",
+      },
+    },
+    {
+      name: ["-a", "--text"],
+      description:
+        "Treat all files as ASCII text.  Normally grep will simply print ``Binary file ... matches'' if files contain binary characters. Use of this option forces grep to output lines matching the specified pattern",
```
QuiiBz, line 180:
```suggestion
        "Treat all files as ASCII text. Normally grep will simply print ``Binary file ... matches'' if files contain binary characters. Use of this option forces grep to output lines matching the specified pattern",
```

dev/grep.ts
```diff
@@ -13,107 +13,297 @@ const completionSpec: Fig.Spec = {
...
+      name: "--help",
+      description:
+        "Print a usage message briefly summarizing these command-line options and the bug-reporting address, then exit",
     },
     {
-      name: ["-RIn"],
+      name: ["-E", "--extended-regexp"],
       description:
-        "Search for a pattern [R]ecursively in the current directory, showing matching line [n]umbers, [I]gnoring non-text files",
+        "Interpret PATTERN as an extended regular expression (-E is specified by POSIX .)",
```
QuiiBz, line 23: Extra space after `POSIX`, same in other places underneath.

QuiiBz (changes requested):
So many improvements on this spec, congrats!

QuiiBz (approved):
LGTM!

PR #477, opened 2021-08-14
feat: update `npm uninstall` generator

fedeci:
- Add `devDependencies` and `optionalDependencies` as suggestions for `npm uninstall` command.
- Join `npm uninstall` aliases

PR #482, opened 2021-08-16
Brendan/npm dependencies generator

brendanfalk:
NPM uninstall suggestions
* description: now the version number
* name/displayName: now the full key, rather than stripping out the scope

### DIFF
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -25,30 +25,33 @@ const searchGenerator: Fig.Generator = {
 };
 
 const dependenciesGenerator: Fig.Generator = {
-  script: "cat package.json",
+  script:
+    "until [[ -f package.json ]] || [[ $PWD = '/' ]]; do cd ..; done; cat package.json",
   postProcess: function (out, context) {
     if (out.trim() === "") {
       return [];
     }
     try {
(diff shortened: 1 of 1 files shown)


fedeci:
#477😆

QuiiBz:
As @fedeci mentioned, looks like #477 contains the same changes, and it has already been merged. Should we close this one or is there another feature that I missed in this PR?

dev/npm.ts
```diff
@@ -25,30 +25,36 @@ const searchGenerator: Fig.Generator = {
 };
 
 const dependenciesGenerator: Fig.Generator = {
-  script: "cat package.json",
+  script:
+    "until [[ -f package.json ]] || [[ $PWD = '/' ]]; do cd ..; done; cat package.json",
```
fedeci, line 29: @QuiiBz this one seems like a good feature to add.

QuiiBz: Good catch!

Nutlope: Ah, great catch! Going to close this PR and re-submit one with this single change

Nutlope:
@brendanfalk I'm closing this PR since @fedeci submitted a similar one

PR #484, opened 2021-08-16
Completed grep spec

fwesss:
feature

grep spec was missing many options and the combos should have been in additionalSuggestions.

Completed the spec per `man grep` and moved combo suggestions to additionalSuggestions.


QuiiBz:
Looks like the same changes have been made in #481, I think we should prioritize it. Feel free to re-open this if needed!

PR #485, opened 2021-08-16
fix: update `yarn workspace` generator

### LINKED ISSUE #483: `yarn workspace [name of workspace]` suggests directory names instead of workspace name
### Sanity checks

yarn

### Issue Details

When typing `yarn workspace [name of workspace] *`, the autocompletion suggests the name of workspace directories, but it should be suggesting the values of the `name` field of the corresponding `package.json`s. Otherwise, proceeding will result in errors such as the following:

```sh
Workspace '[name of workspace]' not found. Did you mean any of the following:
...
```

### Example reproduction

In a project directory using yarn workspaces, type `yarn workspace [name of workspace] [command in workspace]`, using autocompletion for both sections within brackets. I'm using the following configuration:

```
yarn v3.0.0
node v14.17.4
fig v1.0.47 (with updated autocompletions as of this message)
```

fedeci:
Closes #483


dev/yarn.ts
```diff
@@ -1119,82 +1099,77 @@ const completionSpec: Fig.Spec = {
     {
       name: "workspace",
       description: "Manage workspace",
-      generateSpec: async (_tokens, executeShellCommand) => {
-        const { postProcess } = scriptList;
-        const subcommands = [];
+      generateSpec: async (_, executeShellCommand) => {
```
fedeci, line 1102: It is really too slow. I tested it on Babel which is definitely not a small workspace but it takes 2s to get the results.

dev/yarn.ts
```diff
@@ -1119,82 +1099,77 @@ const completionSpec: Fig.Spec = {
...
-              }
-            }
-          }
-        } catch (e) {
-          return { name: "workspaces" };
+        for (const workspace of workspaces) {
+          subcommands.push({
+            name: workspace.name,
+            description: "Workspace",
+            // subcommands: [{
```
fedeci, line 1132: I commented this because it seemed that it was rerunning the entire `generateSpec` each time a spec path was entered:
e.g
```bash
yarn workspace ... # runs generateSpec
yarn workspace @babel/core ... # reruns generateSpec
```

dev/yarn.ts
```diff
@@ -1119,60 +1099,56 @@ const completionSpec: Fig.Spec = {
...
+      generateSpec: async (_, executeShellCommand) => {
+        const subcommands: Fig.Subcommand[] = [];
+        const isYarn1 = (await executeShellCommand("yarn -v"))[0] === "1";
 
-        try {
-          const out = await executeShellCommand("cat package.json");
+        const rawWorkspaces = await executeShellCommand(
+          isYarn1
+            ? "yarn workspaces info --json"
+            : "yarn workspaces list --json"
```
fedeci, line 1109: It's possible that it's caused by some yarn lag🤔 
I have a couple ideas of how to tweak it but I am not sure if they will work, suggestions?

dereklowlind: Is it possible to combine the isYarn1 shell call and the if for rawWorkspaces into a single shell script? That might help a bit on the calling shell overhead

fedeci: Sadly the slowness is from the `info` subcommand. Will try it however.

dev/yarn.ts
dereklowlind: Should this be a subcommand with a generator instead of a generateSpec? Using the variadic flag on a generator may prevent the generator from rerunning each time

fedeci: A generator must return an `Arg[]` type and an arg type cannot provide more subcommands, so 
`yarn workspace @babel/core run` would not be configurable.

dev/yarn.ts
fedeci: We can't combine it in a single script as we need to know if it is yarn1 or yarn2 in the TS spec later.

fedeci:
Currently this is absolutely a thought result to achieve for bigger yarn workspace repos because of the slowness in parsing all of the `package.json` files.

QuiiBz:
Closing since fixed in https://github.com/withfig/autocomplete/pull/678, thanks for the inspiration!

PR #488, opened 2021-08-16
feat: add `z` spec

### LINKED ISSUE #486: [z] completions
### Sanity checks

z

### Issue Details

A tools for jumping between directories. Similar to autojump or j.

https://github.com/rupa/z

fedeci:
Closes #486

### DIFF
--- a/dev/z.ts
+++ b/dev/z.ts
@@ -1,0 +1,26 @@
+// https://github.com/rupa/z
+const completionSpec: Fig.Spec = {
+  name: "z",
+  args: {
+    name: "regex",
+    isVariadic: true,
+  },
+  options: [
+    {
+      name: "-c",
+      description:
+        "Restrict matches to subdirectories of the current directory",
+    },
+    { name: "-e", description: "Echo the best match, don't cd" },
+    { name: "-h", description: "Show a brief help message" },
+    { name: "-l", description: "List only" },
+    { name: "-r", description: "Match by rank only" },
+    { name: "-t", description: "Match by recent access only" },
+    {
+      name: "-x",
+      description: "Remove the current directory from the datafile",
+    },
+  ],
+};
+
+export default completionSpec;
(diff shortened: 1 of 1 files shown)


dev/z.ts
```diff
@@ -0,0 +1,26 @@
+// https://github.com/rupa/z
+const completionSpec: Fig.Spec = {
+  name: "z",
+  args: {
+    name: "regex",
+    isVariadic: true,
```
Monkeyank, line 6: Seems like the regex arg is also optional `z [-chlrtx] [regex1 regex2 ... regexn]`

```suggestion
    isVariadic: true,
    isOptional: true,
```

dev/z.ts
```diff
@@ -0,0 +1,26 @@
+// https://github.com/rupa/z
+const completionSpec: Fig.Spec = {
+  name: "z",
```
Monkeyank, line 3: Can we add a description
```suggestion
  name: "z",
  description: "CLI tool to jump around"
```

Monkeyank (changes requested):

dev/z.ts
fedeci: Completely missed it :)

dev/z.ts
fedeci: I am not sure about this. It may just be a syntax misaligned.
@fwesss does it anything if you don't specify args?

dev/z.ts
fedeci, line 3:
```suggestion
  name: "z",
  description: "CLI tool to jump around directories"
```

dev/z.ts
```diff
@@ -0,0 +1,27 @@
...
+  description: "CLI tool to jump around directories"
```
fedeci, line 4:
```suggestion
  description: "CLI tool to jump around directories",
```

dev/z.ts
fwesss: Running `z` with no args outputs the history. Equivalent to running `z -l`.

QuiiBz (approved):
LGTM! Thanks for this spec

PR #490, opened 2021-08-17
Add composer packages autocompletion

duboiss:
Feature

Nothing displayed.

image
image

**Additional info:**
I have taken the code from the [NPM specification](https://github.com/withfig/autocomplete/blob/5b3571dca7951d9bcc3a21ca3bcc1a6665c345ed/dev/npm.ts#L1).


dev/composer.ts
```diff
@@ -63,6 +126,12 @@ const completionSpec: Fig.Spec = {
               isOptional: !arg.is_required,
               default: argDefault,
               isVariadic: arg.is_array,
+              ...(command.name === "require"
+                ? { generators: searchGenerator }
+                : {}),
+              ...(triggersCommands.includes(command.name)
+                ? { generators: packagesGenerator }
+                : {}),
```
fedeci, line 134:
```suggestion
			  generators: command.name === "require" 
			    ? searchGenerator 
			    : triggersCommands.includes(command.name) 
			    ? packagesGenerator 
			    : []
```

dev/composer.ts
```diff
@@ -33,6 +33,66 @@ interface ComposerListOutput {
...
+            name: item.name,
+            description: item.description,
+            icon: "📦",
+          } as Fig.Suggestion)
+      ) as Fig.Suggestion[];
+    } catch (e) {
+      return [];
+    }
+
+    return temp;
```
QuiiBz, line 58: We can remove the `var` and simplify a bit:

```suggestion
    try {
      return JSON.parse(out).results.map(
        (item) =>
          ({
            name: item.name,
            description: item.description,
            icon: "📦",
          } as Fig.Suggestion)
      ) as Fig.Suggestion[];
    } catch (e) {
      return [];
    }
```

dev/composer.ts
```diff
@@ -33,6 +33,66 @@ interface ComposerListOutput {
...
+  },
+};
+
+// generate package list from composer.json file
+const packagesGenerator: Fig.Generator = {
+  script: "cat composer.json",
+  postProcess: function (out) {
+    console.log("debug", out);
```
QuiiBz, line 66: Could you remove this log?

QuiiBz (changes requested):
This is looking amazing! Only two small issues to fix before merging it!

duboiss:
I just realized that the autocompletion of arguments (and packages) is missing for `composer global` command.
I might see about fixing this in another PR.

QuiiBz:
> I just realized that the autocompletion of arguments (and packages) is missing for `composer global` command.
> I might see about fixing this in another PR.

Feel free to do it in this one, we just need to change the title.

Also, could you resolve the merge conflict so we can get it merged?

Nutlope:
Hey @duboiss, I noticed you deleted the composer spec in the /dev folder and only kept the one inside the php folder. Not actually sure why we added it in the php folder, it should be on the outside since that's how Fig looks for a spec (unless we use `loadSpec` which we're not in this case). Please correct me if I'm wrong @QuiiBz!

Could you move the `composer` spec from the `/dev/php`folder to the `/dev` folder please?

Nutlope (approved):

QuiiBz:
> Not actually sure why we added it in the php folder, it should be on the outside since that's how Fig looks for a spec (unless we use `loadSpec` which we're not in this case). Please correct me if I'm wrong @QuiiBz!

That's 100% right! The `php` folder is currently only used for `artisan`, which is used like a sub-command of `php`: `php artisan ...`. That's why it is in this folder, as opposed to `composer` which isn't a sub-command of `php`.

QuiiBz (approved):
LGTM! Thanks a lot for those amazing generators!

PR #492, opened 2021-08-17
feat: Add `shortcuts` spec (macOS Monterey)

rmorey:
Feature

Backend doesn't exist

It does exist

**Additional info:**
The Shortucts app and this utility are currently available in the macOS Monterey Public Beta, slated for release this Fall.

https://www.apple.com/macos/monterey-preview/
https://sixcolors.com/link/2021/06/shortcuts-shaping-up-on-the-mac/

## Examples
 - `shortcuts list`
 - `shortcuts list --folders`
 - `shortcuts run HelloWorld`
 - `shortcuts view HelloWorld`

Screen Shot 2021-08-17 at 5 01 00 PM

 Screen Shot 2021-08-17 at 2 44 21 PM

### DIFF
--- a/dev/shortcuts.ts
+++ b/dev/shortcuts.ts
@@ -13,12 +13,7 @@ const shortcut: Fig.Arg = {
 const completionSpec: Fig.Spec = {
   name: "shortcuts",
   description: "Command-line utility for running shortcuts.",
-  options: [
-    {
-      name: ["--help", "-h"],
-      description: "Show help information.",
-    },
-  ],
+
   subcommands: [
     {
       name: "run",
@@ -35,23 +30,19 @@ const completionSpec: Fig.Spec = {
       options: [
         {
           name: ["-i", "--input-path"],
-          isRequired: false, // Default: stdin
           args: [
             {
               name: "input-path",
               template: "filepaths",
-              isOptional: false,
             },
           ],
           description: "The input to provide to the shortcut.",
         },
         {
           name: ["-o", "--output-path"],
-          isRequired: false,
           args: [
             {
               name: "output-path",
-              isOptional: false,
               template: "filepaths",
             },
           ],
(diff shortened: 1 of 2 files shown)


dev/shortcuts.ts
```diff
@@ -0,0 +1,180 @@
...
+              template: "filepaths",
+            },
+          ],
+        },
+        {
+          name: ["--mode", "-m"],
+          description: "The signing mode. (default: people-who-know-me)",
+          args: [
+            {
+              name: "mode",
```
Monkeyank, line 146: We can add `default: "people-who-know-me"`

Monkeyank (changes requested):
Thanks so much for this completion spec! Just a little nit I found, otherwise LGTM!

dev/shortcuts.ts
```diff
@@ -0,0 +1,180 @@
...
+        {
+          name: "view",
+          description: "View a shortcut in Shortcuts.",
+          icon: "🔍",
+        },
+        {
+          name: "sign",
+          description: "Sign a shortcut file.",
+          icon: "🔏",
+        },
```
fedeci, line 175: It would be great if we could extract this from the `completionSpec` obj.
Like 
```ts
completionSpec.subcommands.filter(command => command.name !== "help").map(({ name, description, icon }) => ({ name, description, icon }))
```

dev/shortcuts.ts
```diff
@@ -10,170 +10,185 @@ const shortcut: Fig.Arg = {
...
-          description: "Sign a shortcut file.",
-          icon: "🔏",
-        },
-      ],
+      subcommands: subcommands.map(
+        (_) =>
+          Object.fromEntries(
+            ["name", "description", "icon"].map((k) => [k, _[k]])
+          ) as Fig.Subcommand
+      ),
```
fedeci, line 191: Just have a look at perf cause this seems to be O(n^2) and is like 70% slower on jsbench.
It may result in problems when the spec is loaded the first time.

QuiiBz (approved):
This is amazing! Thanks a lot for this spec

PR #498, opened 2021-08-19
feat: add fvm completion spec

jonas-zebari:
This PR adds autocompletion support for the fvm CLI (Flutter version manager).
Currently there is no autocomplete for fvm.
Autocomplete now works with all fvm commands and subcommands except the dart proxy command. When I get around to adding Dart autocomplete support I would like to add that to fvm autocomplete as well (the same way the flutter proxy is currently implemented).
**Additional info:**

### DIFF
--- a/dev/fvm.ts
+++ b/dev/fvm.ts
@@ -14,11 +14,11 @@ const completionSpec: Fig.Spec = {
   options: [
     ...globalOptions,
     {
-      name: ["--verbose"],
+      name: "--verbose",
       description: "Print verbose output.",
     },
     {
-      name: ["--version"],
+      name: "--version",
       description: "Current FVM version.",
     },
   ],
@@ -101,16 +101,14 @@ const completionSpec: Fig.Spec = {
             priority: 98,
           },
         ],
-        generators: [
-          {
-            script: "fvm releases",
-            postProcess: function (out): Fig.Suggestion[] {
-              const matches = out.match(semverRegex);
-              const matchesSet = [...new Set(matches)];
-              return matchesSet.map((match) => ({ name: match })).reverse();
-            },
+        generators: {
+          script: "fvm releases",
+          postProcess: function (out): Fig.Suggestion[] {
+            const matches = out.match(semverRegex);
+            const matchesSet = [...new Set(matches)];
+            return matchesSet.map((match) => ({ name: match })).reverse();
           },
-        ],
+        },
       },
       options: [
         ...globalOptions,
@@ -140,7 +138,7 @@ const completionSpec: Fig.Spec = {
       options: [
         ...globalOptions,
         {
-          name: ["--force"],
+          name: "--force",
           description: "Skips version global check.",
         },
       ],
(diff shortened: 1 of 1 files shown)


dev/fvm.ts
```diff
@@ -0,0 +1,154 @@
+const semverRegex = /((([0-9]+)\.([0-9]+)\.([0-9]+)(?:-([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)/gm;
+
+const globalOptions = [
```
QuiiBz, line 3: Could you add a type of `Fig.Option` here?

dev/fvm.ts
```diff
@@ -0,0 +1,154 @@
...
+  subcommands: [
+    {
+      name: "config",
+      description: "Set configuration for FVM",
+      options: [
+        ...globalOptions,
+        {
+          name: ["-c", "--cache-path"],
+          description:
+            "Set the path which FVM will cache the version. Priority over FVM_HOME.",
```
QuiiBz, line 34: Looks like this option takes an argument, which should have a template of `filepaths`.

dev/fvm.ts
```diff
@@ -0,0 +1,154 @@
...
+          name: ["-g", "--git-cache", "--no-git-cache"],
+          description:
+            "ADVANCED: Will cache a local version of Flutter repo for faster version install.",
+        },
+      ],
+    },
+    {
+      name: "dart",
+      description: "Proxies Dart commands",
+      args: {},
```
QuiiBz, line 50: Does the darg subcommand have any arguments?

dev/fvm.ts
```diff
@@ -0,0 +1,154 @@
...
+    },
+    {
+      name: "doctor",
+      description:
+        "Shows information about environment, and project configuration.",
+    },
+    {
+      name: "flavor",
+      description: "Switches between different project flavors.",
+      args: {},
```
QuiiBz, line 60: Does the flavor subcommand have any arguments?

QuiiBz (changes requested):
Thanks a lot for this PR! I left some comments to resolve before merging it

dev/fvm.ts
jonas-zebari: I don't think so, if I understand correctly.  fvm dart is a proxy to the Dart CLI, so it has its own commands and really needs its own spec. I plan on adding the dart spec soon. Should I remove args: {} completely or leave it empty?

fedeci: It is better to remove it for the moment, when we will have the dart spec we can reference it from here.

dev/fvm.ts
jonas-zebari: Yep, added.

Nutlope (changes requested):
Looks good! One small thing, could you run `npm run lint:fix` so the linter test passes?

Nutlope:
Looks like you just did, thanks for the contribution Jonas!

Nutlope (approved):

PR #503, opened 2021-08-21
[WIP] feat: add versioning to Specs

cstrnt:
Feature

There is no way to version Specs

Specs can now be versioned and share code more easily

**Additional info:**


schemas/fig.d.ts
```diff
@@ -198,6 +198,9 @@ declare namespace Fig {
     parserDirectives?: {
       flagsArePosixNoncompliant?: boolean;
     };
+
+    getVersionCommand?: string;
```
sullivan-sean, line 202: Is there a reason this is on `Subcommand` as opposed to `Spec`?

sullivan-sean (commented):
Looks great 🙌  Only change that might be nice is an update to the `README.md` explaining a bit of how versioning works for future contributors

Nutlope:
Hey @cstrnt, heads up that @QuiiBz and I merged a ton of specs in the past 24h so re-pulling from master would be great before merging. Also give me a heads up when this is merged so I can update the docs accordingly. Great work man!

fedeci (commented):
Why not using workspaces (yarn) and make each spec a package that can be downloaded separately from npm. 
Benefits:
- Modularity
- A user could just install what needs

Not strictly related to this PR, but may be good to join the two things.

PR #506, opened 2021-08-22
feat: add laravel completion spec

cipi1965:

This PR adds autocompletion support for laravel (Laravel Installer cli).

Missing support

**Additional info:**

### DIFF
--- a/dev/laravel.ts
+++ b/dev/laravel.ts
@@ -46,6 +46,27 @@ const completionSpec: Fig.Spec = {
           description: 'Installs the latest "development" release',
         },
         {
+          name: "--git",
+          description: "Initialize a Git repository",
+        },
+        {
+          name: "--branch",
+          description: "The branch that should be created for a new repository",
+          args: {
+            name: "BRANCH",
+            default: "main",
+          },
+        },
+        {
+          name: "--github",
+          description: "Create a new repository on GitHub",
+          args: {
+            name: "GITHUB",
+            isOptional: true,
+            default: "false",
+          },
+        },
+        {
           name: "--jet",
           description: "Installs the Laravel Jetstream scaffolding",
         },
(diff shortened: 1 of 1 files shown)


dev/laravel.ts
```diff
@@ -0,0 +1,110 @@
...
+        },
+      ],
+    },
+    {
+      name: "new",
+      description: "Create a new Laravel application",
+      args: {
+        name: "name",
+      },
+      options: [
```
fedeci, line 43: Is the `--git` and github related options missing? https://github.com/laravel/installer/blob/da5e8b471d87c0b4762c3a0a4c35f09521b48db9/src/NewCommand.php

cipi1965: Oops i think i based my work on a old version of the tool 😅

cipi1965: Added

dev/laravel.ts
```diff
@@ -0,0 +1,131 @@
...
+          name: "--format",
+          description: "The output format (txt, xml, json, or md)",
+          args: {
+            suggestions: ["txt", "xml", "json", "md"],
+          },
+        },
+        {
+          name: "--raw",
+          description: "To output raw command list",
+        },
```
QuiiBz, line 18: I think we could extract those options to a top-level const, instead of duplicating them (see https://github.com/withfig/autocomplete/pull/506/files#diff-ba0e472ae0f57d53510879cf1b3e1a8097585a18dee3325b5d0103192cfdd16eR24-R35)

dev/laravel.ts
```diff
@@ -0,0 +1,131 @@
...
+        {
+          name: "--jet",
+          description: "Installs the Laravel Jetstream scaffolding",
+        },
+        {
+          name: "--stack",
+          description: "The Jetstream stack that should be installed",
+          args: {
+            name: "STACK",
```
QuiiBz, line 77: Could we provide suggestions for this argument?

QuiiBz (changes requested):
PHP devs are gonna thank you! Left some comments to resolve before merging it

QuiiBz (approved):

PR #508, opened 2021-08-23
feat: add hyper completion spec

Deep-Codes:

Feature : adds specs for `hyper`

### DIFF
--- a/dev/hyper.ts
+++ b/dev/hyper.ts
@@ -43,6 +43,20 @@ const completionSpec: Fig.Spec = {
     {
       name: ["uninstall", "u", "rm", "remove"],
       description: "Uninstall plugin",
+      args: {
+        name: "plugin",
+        description: "plugin to uninstall",
+        generators: [
+          {
+            script: "hyper list",
+            postProcess: function (out) {
+              return out.split("\n").map((p) => {
+                return { name: p, description: "plugin" };
+              });
+            },
+          },
+        ],
+      },
     },
     {
       name: "version",
--- a/dev/hyper.ts
+++ b/dev/hyper.ts
@@ -1,6 +1,9 @@
 const completionSpec: Fig.Spec = {
   name: "hyper",
   description: "Hyper is an Electron-based terminal",
+  args: {
+    template: "folders",
+  },
   subcommands: [
     {
       name: ["install", "i"],
--- a/dev/hyper.ts
+++ b/dev/hyper.ts
@@ -34,6 +34,11 @@ const completionSpec: Fig.Spec = {
     {
       name: ["search", "s"],
       description: "Search for plugins on npm",
+      args: {
+        isOptional: true,
+        name: "query",
+        description: "your search query",
+      },
     },
     {
       name: ["uninstall", "u", "rm", "remove"],
--- a/dev/hyper.ts
+++ b/dev/hyper.ts
@@ -1,0 +1,56 @@
+const completionSpec: Fig.Spec = {
+  name: "hyper",
+  description: "Hyper is an Electron-based terminal",
+  subcommands: [
+    {
+      name: ["install", "i"],
+      description: "Install a plugin",
+      args: {
+        name: "plugin",
+      },
+    },
+    {
+      name: ["docs", "d", "h", "home"],
+      description: "Open the npm page of a plugin",
+      args: {
+        name: "plugin",
+      },
+    },
+    {
+      name: "help",
+      description: "Display help",
+    },
+    {
+      name: ["list", "ls"],
+      description: "List installed plugins",
+    },
+    {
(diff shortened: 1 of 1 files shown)


fedeci (commented):
If I specify a folder as arg does it open that folder in hyper?
e.g. `hyper node/`

Deep-Codes:
> If I specify a folder as arg does it open that folder in hyper?
> e.g. `hyper node/`

It does adding `folder` as template

Screenshot 2021-08-23 at 8 18 47 PM

dev/hyper.ts
```diff
@@ -0,0 +1,59 @@
...
+      name: ["list", "ls"],
+      description: "List installed plugins",
+    },
+    {
+      name: ["list-remote", "lsr", " ls-remote"],
+      description: "List plugins available on npm",
+    },
+    {
+      name: ["search", "s"],
+      description: "Search for plugins on npm",
```
QuiiBz, line 36: If think this option takes an argument, which is a string to search on npm?

dev/hyper.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+    {
+      name: ["uninstall", "u", "rm", "remove"],
+      description: "Uninstall plugin",
```
QuiiBz, line 40: Same as above but for the name of the plugin

QuiiBz (changes requested):
Awesome! There are some arguments that are missing thought

dev/hyper.ts
fedeci: It can actually be autogenerated via the `list` subcommand.

QuiiBz: Yes! But adding generators isn't mandatory when adding a completely new spec. We can add them later!

dev/hyper.ts
Deep-Codes: i don't think so 

```
  Usage: hyper [options] [command]
  
  Commands:
    <default>                    Launch Hyper
    docs, d, h, home             Open the npm page of a plugin
    help                         Display help
    install, i                   Install a plugin
    list, ls                     List installed plugins
    list-remote, lsr, ls-remote  List plugins available on npm
    search, s                    Search for plugins on npm
    uninstall, u, rm, remove     Uninstall a plugin
    version                      Show the version of hyper
  
  Options:
    -h, --help     Output usage information
    -v, --verbose  Verbose mode (disabled by default)
  
```

QuiiBz: Could you try running `hyper search help`?

Deep-Codes: this is the usage 

`> hyper search` doesn't take an argument 🤔 

Screenshot 2021-08-23 at 9 11 34 PM

Deep-Codes: Yes it is

QuiiBz: Just installed and tried it, and looks like I can add search arguments (`material` here). Maybe we don't have the same version of the cli?
Screenshot 2021-08-23 at 17 46 13

Deep-Codes: yes, it does search according to the argument.

Should i add, args to the search subcommand?

Screenshot 2021-08-23 at 9 21 33 PM

QuiiBz: Yes you can add it, along with the [isOptional](https://fig.io/docs/reference/arg#isoptional) option!

dev/hyper.ts
Deep-Codes: Can you explain What should I do here?

QuiiBz: We were talking about [generators](https://fig.io/docs/reference/generator), which is an amazing feature of Fig. They allow you to dynamically provide suggestions, using a shell command. In this case, we could make a generator for `uninstall`, which would run the `hyper ls` command in the background, parse the result, and output it as suggestions.

You don't need to make generators right now, it's just a cool feature that can be added in the future. Feel free to ask if you need more help on this!

dev/hyper.ts
Deep-Codes: Added ✅

dev/hyper.ts
Deep-Codes: Hey got it running 🚀

Screenshot 2021-08-23 at 10 57 30 PM

Deep-Codes:
Seems like a new lint rule is failing CI 🙂

fedeci:
You can rebase on master, run `npm run lint:fix` and we are good to go!

Deep-Codes:
@fedeci i think this is done 🙂

QuiiBz:
Could you rebase your PR to `master`? There are a lot of past commits above.

QuiiBz (approved):
Just removed the past commits, thanks a lot for this PR!

PR #509, opened 2021-08-23
feat: add afplay completion spec

Deep-Codes:

Feature: adds specs for `afplay`

### DIFF
--- a/dev/afplay.ts
+++ b/dev/afplay.ts
@@ -1,0 +1,57 @@
+const completionSpec: Fig.Spec = {
+  name: "afplay",
+  description: "Audio File Play",
+  args: {
+    template: "filepaths",
+    name: "play the song in the path",
+  },
+  options: [
+    {
+      name: ["-v", "--volume"],
+      description: `Set the volume for playback of the file
+      Apple does not define a value range for this, but it appears to accept
+      0=silent, 1=normal (default) and then up to 255=Very loud.
+      The scale is logarithmic and in addition to (not a replacement for) other volume control(s).`,
+      args: {
+        name: "0-1",
+      },
+    },
+    {
+      name: ["-h", "--help"],
+      description: "print help",
+    },
+    {
+      name: ["--leaks"],
+      description: "Run leaks analysis.",
+    },
+    {
+      name: ["-t", "--time"],
+      description: `Play for \`TIME\` seconds
+      >0 and < duration of \`audio_file\`.`,
+      args: {
+        name: "TIME",
+      },
+    },
+    {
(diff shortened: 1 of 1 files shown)


dev/afplay.ts
```diff
@@ -0,0 +1,57 @@
...
+      description: "print help",
+    },
+    {
+      name: ["--leaks"],
+      description: "Run leaks analysis.",
+    },
+    {
+      name: ["-t", "--time"],
+      description: `Play for \`TIME\` seconds
+      >0 and < duration of \`audio_file\`.`,
```
fedeci, line 30: I think we can add it to the description of the argument.

dev/afplay.ts
```diff
@@ -0,0 +1,57 @@
...
+      args: {
+        name: "TIME",
+      },
+    },
+    {
+      name: ["-r", "--rate"],
+      description: `Play at playback \`RATE\`.
+      practical limits are about 0.4 (slower) to 3.0 (faster).`,
```
fedeci, line 38: Same here.

dev/afplay.ts
```diff
@@ -0,0 +1,57 @@
...
+    template: "filepaths",
+    name: "play the song in the path",
+  },
+  options: [
+    {
+      name: ["-v", "--volume"],
+      description: `Set the volume for playback of the file
+      Apple does not define a value range for this, but it appears to accept
+      0=silent, 1=normal (default) and then up to 255=Very loud.
+      The scale is logarithmic and in addition to (not a replacement for) other volume control(s).`,
```
fedeci, line 14: And here too.

dev/afplay.ts
Deep-Codes: I don't understand why 🤔 

if I add it in the `args`'s description I will get the autocomplete after i write the flag and the description would come for the flag's value.

Screenshot 2021-08-23 at 6 55 19 PM

Doesn't this seem more explanatory? 

Screenshot 2021-08-23 at 6 57 31 PM

fedeci: I don't mean everything, just the constraints part. Like
```js
  name: '-t',
  description: 'Play for TIME seconds,
  args: {
    name: 'TIME',
    description: 'A value greater than 0 but lower than the audio file length'
```

Deep-Codes: Thanks, sorry for the misunderstanding changing it.

Deep-Codes:
@fedeci Done let me know 🙂

Nutlope (approved):

Nutlope:
LGTM, thanks for the contribution @Deep-Codes!

PR #512, opened 2021-08-23
[mv] Fixed 1st arg and options

fwesss:
-R and -P are not options on mv so they were removed and replaced with -f, -i, -n, and -v

First arg should also be variadic


Nutlope (approved):

Nutlope:
Nice use of exclusiveOn! LGTM!

PR #514, opened 2021-08-23
[cp] Finished spec

fwesss:
First arg to copy is variadic. 

All options added per "man cp"


Nutlope:
Crazy how these specs were missing so much, nice work!

PR #515, opened 2021-08-23
[brew] info subcommand completed

fwesss:
Added all options and args for brew info

## Generator for --category
Category has different possible args depending on whether an arg has been listed for the subcommand. 
The generator looks at the context to see if any of the tokens after "brew info" do not start with a '-'. This would mean an arg was supplied so we should show suggestions for providing info on a formula or cask.

### DIFF
--- a/dev/brew.ts
+++ b/dev/brew.ts
@@ -48,8 +48,99 @@ const completionSpec: Fig.Spec = {
       ],
     },
     {
-      name: "info",
+      name: ["info", "abv"],
       description: "Display brief statistics for your Homebrew installation",
+      args: {
+        isVariadic: true,
+        isOptional: true,
+        name: "formula",
+        description: "Formula or cask to summarize",
+        generators: {
+          script:
+            "HBPATH=$(brew --repository); ls -1 $HBPATH/Library/Taps/homebrew/homebrew-core/Formula $HBPATH/Library/Taps/homebrew/homebrew-cask/Casks",
+          postProcess: function (out) {
+            return out.split("\n").map((formula) => {
+              return {
+                name: formula.replace(".rb", ""),
+                description: "formula",
+                icon: "🍺",
+                priority:
+                  (formula[0] >= "0" && formula[0] <= "9") || formula[0] == "/"
+                    ? 0
+                    : 51,
+              };
+            });
+          },
+        },
+      },
+      options: [
+        {
+          name: "--analytics",
+          description:
+            "List global Homebrew analytics data or, if specified, installation and build error data for formula",
+        },
+        {
+          name: "--days",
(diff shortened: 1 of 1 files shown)


Nutlope (changes requested):
Looks good! One small thing: from a user experience perspective I think it'd be better to separate "info" and "abv" into 2 separate subcommands since info is used a lot and it's better for it to stay separate as a standalone subcommand

dev/brew.ts
```diff
@@ -13,6 +13,104 @@ const generators: Record<string, Fig.Generator> = {
   },
 };
 
+// brew info is equiv to brew abv. Everything but 'name' is shared.
+const brewInfo = {
```
QuiiBz, line 17: Could you add the `Fig.Option` type here?

dev/brew.ts
```diff
@@ -13,6 +13,104 @@ const generators: Record<string, Fig.Generator> = {
...
+    {
+      name: "--analytics",
+      description:
+        "List global Homebrew analytics data or, if specified, installation and build error data for formula",
+    },
+    {
+      name: "--days",
+      description: "How many days of analytics data to retrieve",
+      exclusiveOn: ["--analytics"],
+      args: {},
```
QuiiBz, line 52: Could you add a `name` (and maybe `description`) to this argument?

dev/brew.ts
```diff
@@ -13,6 +13,104 @@ const generators: Record<string, Fig.Generator> = {
...
+    },
+    {
+      name: "--category",
+      description: "How many days of analytics data to retrieve",
```
QuiiBz, line 56: Seems like this isn't the right description?

QuiiBz (changes requested):
Nice improvements!

dev/brew.ts
```diff
@@ -49,7 +147,11 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: "info",
-      description: "Display brief statistics for your Homebrew installation",
+      ...brewInfo,
+    },
+    {
+      name: "abv",
+      ...brewInfo,
```
fedeci, line 154: nit: same as https://github.com/withfig/autocomplete/pull/516#discussion_r694740753 IMHO

dev/brew.ts
fwesss: I've changed this to a function that takes a name and outputs a Fig.Suggestion per @fedeci comment below.

dev/brew.ts
fwesss: Good catch!

dev/brew.ts
fwesss: Agreed, much cleaner.

QuiiBz (approved):
LGTM!

PR #516, opened 2021-08-23
[npm] changes to dependencies gen and npm uninstall

Nutlope:
feature

- Npm uninstall has too many acronyms that provides suboptimal user experience
- Package.json doesn't check top level folders if user is running a command from within a nested folder

- Breaking up npm uninstall for better UX
- Change script of dependencies generator taken from Brendan's previous PR

### DIFF
--- a/dev/npm.ts
+++ b/dev/npm.ts
@@ -50,7 +50,8 @@ const workspaceGenerator: Fig.Generator = {
 };
 
 const dependenciesGenerator: Fig.Generator = {
-  script: "cat package.json",
+  script:
+    "until [[ -f package.json ]] || [[ $PWD = '/' ]]; do cd ..; done; cat package.json",
   postProcess: function (out, context) {
     if (out.trim() === "") {
       return [];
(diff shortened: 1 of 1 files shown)


dev/npm.ts
```diff
@@ -723,7 +724,47 @@ const completionSpec: Fig.Spec = {
     { name: "token", description: "manage your authentication tokens" },
     { name: "tst", description: "test a package" },
     {
-      name: ["uninstall", "remove", "rm", "r", "unlink", "un"],
+      name: "uninstall",
```
fedeci, line 727: I disagree with this change because it adds many useless(?) repetitions to the codebase.
I don't really think it will improve the UX since if a user wants to remove a package is well aware that `uninstall` has the same meaning.

Nutlope: Adding repetitions in the codebase doesn't really matter as we're optimizing for UX rather than the best possible codebase (DX) as most of our users will never look inside our codebase.

I see where you're coming from but I disagree. If I'm used to typing `npm rm PACKAGE` and I get this with Fig, it doesn't look great.

image

fedeci: Wouldn't a new `BaseCompletion` flag be better?

dev/npm.ts
```diff
@@ -723,7 +724,47 @@ const completionSpec: Fig.Spec = {
...
+    {
+      name: "un",
+      description: "uninstall a package",
+      args: {
+        name: "package",
+        generators: dependenciesGenerator,
+        isVariadic: true,
+      },
+      options: npmInstallOptions,
+    },
```
QuiiBz, line 765: We can clean this by extracting the `option` object to a top-level const:
```ts
const uninstallOption: Fig.Option = {
  name: "uninstall",
  description: "uninstall a package",
  args: {
    name: "package",
    generators: dependenciesGenerator,
    isVariadic: true,
  },
  options: npmInstallOptions,
}
```
and then spread it while updating the name:
```ts
[
  // uninstall
  {
    ...uninstallOption,
  },
  // remove
  {
    ...uninstallOption,
    name: "remove",
  },
  // r, rm
  {
    ...uninstallOption,
    name: ["r", "rm"],
  },
  ...
]
```

dev/npm.ts
QuiiBz: Agree with @Nutlope that it's better for the UX to separate them. But this idea seems interesting.

QuiiBz (changes requested):

dev/npm.ts
fedeci: Functional approach is better then.
```js
function uninstallNamed(name: string): Fig.Subcommand {
	return {
  		name,
	    description: "uninstall a package",
	    args: {
	      name: "package",
	      generators: dependenciesGenerator,
	      isVariadic: true,
	    },
	    options: npmInstallOptions,
  	}
}
...
subcommands: [
	  ...,
	  uninstallNamed("remove"),
	  uninstallNamed("uninstall"),
]
```

QuiiBz: Both are valid options, the first approach is a lot more used in this codebase.

fedeci: Always in time to change. Imagine if you place `name` and the `object spread` in the opposite position? You would get no result.

PR #519, opened 2021-08-23
draft PR for heroku without gens

Nutlope:
Rescrape heroku plugin with our oclif parser to fix some required arguments being marked optional

Many arguments that are mandatory are put in as optional args

Fix some required arguments being marked optional

### DIFF
--- a/dev/heroku.ts
+++ b/dev/heroku.ts
@@ -1,471 +1,542 @@
-const getAppGenerator: Fig.Generator = {
-  script: "heroku apps --all --json",
-  postProcess: function (out) {
-    try {
-      return JSON.parse(out).map((app) => {
-        return {
-          name: app.name,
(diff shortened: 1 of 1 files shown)


dev/heroku.ts
```diff
@@ -5682,693 +5891,2596 @@ const completionSpec: Fig.Spec = {
...
-          name: "--json",
-          description: "output in json format",
+          name: ["-v", "--verbose"],
+          description: "",
         },
       ],
       args: {
-        name: "space",
+        name: "path",
+        description: "path to plugin",
```
QuiiBz, line 6112: Same as above (is it a `folders` template?)

QuiiBz (changes requested):
Huge improvements on this spec! I left some comments to improve a bit the spec

PR #520, opened 2021-08-24
[touch] re-work per BSD spec

fwesss:
Touch spec was originally written for the GNU version but macOS ships with BSD. This PR changes the touch spec to comply with BSD (man spec on macOS).

Basic functions are the same but some of the options differ.


QuiiBz (approved):
Good catch, LGTM!

PR #521, opened 2021-08-24
fix: Schema declaration example

Mitch528:

Fix example in schema declaration file

Example does not compile

Example compiles

**Additional info:**


QuiiBz (approved):
Really good catch, thanks a lot!

PR #523, opened 2021-08-25
style(eslint): add `conventional-descriptions` rule

fedeci:
Nothing of too innovative but again it should help to standardize the repo.
It only works on `StringLiteral`s so no template literal support.

We may also enforce first letter capitalized and no trailing/leading whitespace.

At the opposite we can enforce trailing dot.


QuiiBz (approved):
Amazing rule, love to have such consistency across specs. LGTM!

Issue #493, opened 2021-08-17
[find] Completion Spec


JosephTLyons:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

find

### Which statement makes the most sense?

This completion spec does not exist

### Issue Details

It would be great if there were completions for [find](https://man7.org/linux/man-pages/man1/find.1.html) command

QuiiBz:
Merged in #528, thanks everyone!

PR #528, opened 2021-08-26
[find] basic spec

fwesss:
Adds basic args and options for `find`

The 2nd arg, expression, is actually a combo of 'primaries' and 'operands'. See `man find` for details on how this works. 
Not sure how best to offer suggestions on expressions based on Fig specs at this time.

This will at least provide basic suggestions for the find tool.

re: https://github.com/withfig/autocomplete/issues/493


QuiiBz (approved):
LGTM!

PR #531, opened 2021-08-27
Add spec for Ag (Silver Searcher)

### LINKED ISSUE #529: [ag] support for the_silver_searcher
### Sanity checks

ag

### Issue Details

ag (the_silver_searcher) is a faster version of ack

https://github.com/ggreer/the_silver_searcher

RSO:
Add all the args from the `man ag` page.

This is my first time writing autocomplete specs so let me know if there's anything I should've done differently!

Fixes #529

### DIFF
--- a/dev/ag.ts
+++ b/dev/ag.ts
@@ -28,11 +28,13 @@ const completionSpec: Fig.Spec = {
     {
       name: "--affinity",
       description: "Set thread affinity (if platform supports it)",
+      exclusiveOn: ["--noaffinity"],
     },
 
     {
       name: "--noaffinity",
       description: "Don't set thread affinity (if platform supports it)",
+      exclusiveOn: ["--affinity"],
     },
 
     {
@@ -78,11 +80,13 @@ const completionSpec: Fig.Spec = {
     {
       name: "--color",
       description: "Print color codes in results",
+      exclusiveOn: ["--nocolor"],
     },
 
     {
       name: "--nocolor",
       description: "Don't print color codes in results",
+      exclusiveOn: ["--color"],
     },
 
     {
@@ -133,21 +137,25 @@ const completionSpec: Fig.Spec = {
     {
       name: "--filename",
       description: "Print file names",
+      exclusiveOn: ["--nofilename"],
     },
 
     {
       name: "--nofilename",
       description: "Don't print file names",
+      exclusiveOn: ["--filename"],
     },
 
     {
       name: ["-f", "--follow"],
       description: "Follow symlinks",
+      exclusiveOn: ["--nofollow"],
     },
 
     {
       name: "--nofollow",
       description: "Don't follow symlinks",
+      exclusiveOn: ["-f", "--follow"],
     },
 
     {
@@ -159,12 +167,14 @@ const completionSpec: Fig.Spec = {
       name: "--group",
       description:
         "Lumps multiple matches in the same file together, and presents them under a single occurrence of the filename",
+      exclusiveOn: ["--nogroup"],
     },
 
     {
       name: "--nogroup",
       description:
         "Refrains from lumping matches in the same file together, and instead places the filename at the start of each match line",
+      exclusiveOn: ["--group"],
     },
 
     {
@@ -188,11 +198,13 @@ const completionSpec: Fig.Spec = {
     {
       name: ["-H", "--heading"],
       description: "Print filenames above matching contents",
+      exclusiveOn: ["--noheading"],
     },
 
     {
       name: "--noheading",
       description: "Don't print filenames above matching contents",
+      exclusiveOn: ["-H", "--heading"],
     },
 
     {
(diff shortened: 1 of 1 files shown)


dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+    },
+
+    {
+      name: "--color",
+      description: "Print color codes in results.",
+    },
+
+    {
+      name: "--nocolor",
+      description: "Don't print color codes in results.",
```
QuiiBz, line 85: We can add [exclusiveOn](https://fig.io/docs/reference/option#exclusiveon) for those two options!

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+    },
+
+    {
+      name: ["-f", "--follow"],
+      description: "Follow symlinks.",
+    },
+
+    {
+      name: "--nofollow",
+      description: "Don't follow symlinks.",
```
QuiiBz, line 150: Same as above for `filename / nofilename` and `follow / nofollow`

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+    {
+      name: "--group",
+      description:
+        "Lumps multiple matches in the same file together, and presents them under a single occurrence of the filename.",
+    },
+
+    {
+      name: "--nogroup",
+      description:
+        "Refrains from lumping matches in the same file together, and instead places the filename at the start of each match line.",
```
QuiiBz, line 167: Same as above

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+    },
+
+    {
+      name: ["-H", "--heading"],
+      description: "Print filenames above matching contents.",
+    },
+
+    {
+      name: "--noheading",
+      description: "Don't print filenames above matching contents.",
```
QuiiBz, line 195: Same as above

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+      description: "Search hidden files. This option obeys ignored files.",
+    },
+
+    {
+      name: "--ignore",
+      description:
+        "Ignore files/directories whose names match this pattern. Literal file and directory names are also allowed.",
+      args: {
+        name: "PATTERN",
+        description: "The pattern to look for",
```
QuiiBz, line 209: I think we can also add a template of `filepaths` and `folders` here

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+      },
+    },
+
+    {
+      name: "--ignore-dir",
+      description: "Alias for --ignore for compatibility with ack.",
+      args: {
+        name: "NAME",
+        description: "The directory to ignore",
```
QuiiBz, line 218: Same as above but only for `folers`

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+    },
+
+    {
+      name: "--multiline",
+      description: "Match regexes across newlines.",
+    },
+
+    {
+      name: "--nomultiline",
+      description: "Don't match regexes across newlines",
```
QuiiBz, line 267: Same as above for [exclusiveOn](https://fig.io/docs/reference/option#exclusiveon)

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+      description:
+        "When  recursing  directories, don't scan dirs that reside on other storage devices. This lets you avoid scanning slow network mounts. This feature is not supported on all platforms.",
+    },
+
+    {
+      name: ["-p", "--path-to-ignore"],
+      description: "Provide a path to a specific .ignore file.",
+      args: {
+        name: "STRING",
+        description: "The path to the ignore file",
```
QuiiBz, line 301: Looks like we can add a template of `filepaths`

dev/ag.ts
```diff
@@ -0,0 +1,433 @@
...
+      },
+    },
+
+    {
+      name: "--pager",
+      description:
+        "Use a pager such as less. Use --nopager to override. This option is also ignored if output is piped to another program.",
+      args: {
+        name: "COMMAND",
+        description: "The pager",
```
QuiiBz, line 311: Could we add [suggestions](https://fig.io/docs/reference/arg#suggestions) to this argument? (just to name a few, more, less, most...)

QuiiBz (changes requested):
Thanks a lot for this spec!

> This is my first time writing autocomplete specs so let me know if there's anything I should've done differently!

That's actually really good! I left some comments to improve further the spec. Also, could you rebase/merge `master`, because we added a new ESLint rule (#523) for descriptions, and I think there are some issues with it in this spec.

RSO:
> > Is there a way to get the autocomplete to show certain options above others?
> 
> Yep! You can use the [`priority`](https://fig.io/docs/reference/option#priority) property.

It seems to me that the priority only is available on the `option` and not on the `arg`? In my case, I'd like the args to be shown above the options if that makes sense!

QuiiBz:
> It seems to me that the priority only is available on the `option` and not on the `arg`? In my case, I'd like the args to be shown above the options if that makes sense!

You're right, `priority` is currently only available for `options`. We might need to add it to `args`.

QuiiBz (approved):
LGTM, thanks a lot for this amazing spec!

PR #532, opened 2021-08-27
Pathspec arg can be a dir

fwesss:
man git-add:
<pathspec>
...Also a leading directory name (e.g.  dir to add dir/file1 and dir/file2) can be given to update the index to match the current state of the directory as a whole (e.g. specifying dir will record not just a file dir/file1 modified in the working tree, a file dir/file2 added to the working tree, but also a file dir/file3 removed from the working tree)...


QuiiBz (approved):
Really good catch, thanks!

PR #534, opened 2021-08-28
feat: add `babel` spec

fedeci:
New `babel` spec


dev/npx.ts
```diff
@@ -1,4 +1,8 @@
 const suggestions: Fig.Suggestion[] = [
+  {
+    name: "babel",
+    icon: "https://raw.githubusercontent.com/babel/logo/master/babel.png",
```
QuiiBz, line 4: Shouldn't we also add it to `yarn`, to allow suggestions for `yarn babel <command...`?

fedeci: I forgot that😄

QuiiBz (approved):
LGTM!

PR #535, opened 2021-08-28
feat: publish types

### LINKED ISSUE #533: Make autocomplete types available as an external npm package
### Sanity checks

`@withfig/autocomplete` itself

Something else

### Issue Details

It may be useful for external tools to be able to download and use types. 
Currently no module is exported from the package so no type is available from the outside and is required to manually copying everything.

fedeci:
Closes #533 
We publish types along with `@withfig/autocomplete` package on npm so that they can be imported from external libraries too.
In the future we may want to move types to an external package. (breaking change needed)


mschrage (changes requested):
I definitely agree that we should pull the type definitions out so that they can easily by included by other libs.

I don't love that every spec has to include `import * as Fig from "../schemas";`. I'm confused why this was not necessary before and but is now?

fedeci:
Yeah, I don't like it either so I did some research and it seems that when a declaration file containes one or more exports it is treated as a module so the namespace is no longer available throughout the project without importing it ([SO](https://stackoverflow.com/a/42026565)). There is no way to achieve both the results without duplicating everything and it does not sound like a viable solution to me.
Now, since TS namespaces are outdated compared to ES modules I chose to use the latter.

fedeci:
Closing as this has been already implemented

PR #536, opened 2021-08-29
feat: add autocomplete spec for `dart`

jonas-zebari:
New autocomplete spec.
No autocomplete for Dart.
Full autocomplete for Dart CLI.
**Additional info:**


QuiiBz (changes requested):
Thanks for this PR! Found some issues to fix before merging, most of them are due to shared option for `--...` and `--no-...`. I left an example on how to fix this issue in a comment below.

jonas-zebari:
There are some more missed cases of --xyz and --no-xyz being in the same option, so I am working on that right now.

QuiiBz (approved):
Amazing spec, thanks a lot for this contribution!

PR #543, opened 2021-08-31
[aws] sso login

danilofuchs:
Implements the only missing suggestion for AWS SSO

https://awscli.amazonaws.com/v2/documentation/api/latest/reference/sso/index.html#cli-aws-sso

### DIFF
--- a/dev/aws/sso.ts
+++ b/dev/aws/sso.ts
@@ -203,6 +203,21 @@ const completionSpec: Fig.Spec = {
       ],
     },
     {
+      name: "login",
+      description:
+        "Refreshes SSO access tokens and logs in. Required setup using ``aws configure sso``",
+      options: [
+        {
+          name: "--profile",
+          description:
+            "The name of the profile to use.\n\nIf not set, will use the default profile",
+          args: {
+            name: "string",
+          },
+        },
+      ],
+    },
+    {
       name: "logout",
       description:
         "Removes the client- and server-side session that is associated with the user",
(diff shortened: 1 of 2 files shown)


dev/aws/sso.ts
```diff
@@ -202,6 +202,21 @@ const completionSpec: Fig.Spec = {
...
+      name: "login",
+      description:
+        "Refreshes SSO access tokens and logs in. Required setup using ``aws configure sso``",
+      options: [
+        {
+          name: "--profile",
+          description:
+            "The name of the profile to use.\n\nIf not set, will use the default profile",
+          args: {
+            name: "string",
```
fedeci, line 215:
```suggestion
            name: "name",
```

dev/aws/sso.ts
```diff
@@ -202,6 +202,21 @@ const completionSpec: Fig.Spec = {
         },
       ],
     },
+    {
+      name: "login",
+      description:
+        "Refreshes SSO access tokens and logs in. Required setup using ``aws configure sso``",
```
QuiiBz, line 208: Looks like we have some extra `:
```suggestion
        "Refreshes SSO access tokens and logs in. Required setup using `aws configure sso`",
```

dev/aws/sso.ts
QuiiBz: Agree with @fedeci

QuiiBz (changes requested):

We only have two comments to resolve before getting this PR merged in.

dev/aws/sso.ts
danilofuchs: I've used other AWS commands as template. It seems like all of them use double backticks

QuiiBz (approved):
LGTM, thanks a lot for this PR!

PR #547, opened 2021-08-31
feat: add `splash-cli` spec

Rawnly:
Added [splash-cli](https://github.com/splash-cli/splash-cli) spec.


dev/splash.ts
```diff
@@ -0,0 +1,314 @@
+const completionSpec: Fig.Spec = {
+  name: "splash-cli",
+  displayName: "Splash",
+  description: "Get stunning wallpapers from Unsplash",
+  subcommands: [
+    {
+      priority: 100,
```
fedeci, line 7: Having fixed priorities for each subcommand does not allow fig to give priority to the most used commands, I would remove all of them when not necessary.

dev/splash.ts
```diff
@@ -0,0 +1,314 @@
...
+          name: "get",
+          args: {
+            name: "config key",
+            description: "Config key",
+            isOptional: true,
+          },
+        },
+        {
+          name: "set",
+        },
```
fedeci, line 21: Can you add a description for this. I also think that it accepts an arg.

dev/splash.ts
Rawnly: Oh ok! Fixed in [`f4b6b8b` (#547)](https://github.com/withfig/autocomplete/pull/547/commits/f4b6b8b5615d83dcf6ede3fc421ee191119f25eb)

dev/splash.ts
```diff
@@ -0,0 +1,309 @@
...
+            {
+              name: "key",
+            },
+            {
+              name: "value",
+            },
+          ],
+        },
+        {
+          name: "help",
```
fedeci, line 52:
```suggestion
          name: "help",
          description: "Show help menu",
```

dev/splash.ts
```diff
@@ -0,0 +1,309 @@
...
+    },
+    {
+      name: "user",
+      subcommands: [
+        {
+          name: "login",
+        },
+        {
+          name: "logout",
+        },
```
fedeci, line 136: Same here

yarn.lock
```diff
@@ -0,0 +1,3059 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
fedeci, line 1: Can you remove this from the PR? We are using npm😄

dev/splash.ts
Rawnly: It doesn't! (I'm the author of the cli 😉)

fedeci: I've seen that later😅

PR #548, opened 2021-08-31
[git] remove insertValue from clone --recurse-submodules

fwesss:
The argument for `git clone --recurse-submodules` is optional but the insertValue of "--recurse-submodules=" made it appear mandatory.

Users would need to backspace the '=' after entering the suggestion or believed that the arg was mandatory.


QuiiBz (approved):
Really good catch!

PR #554, opened 2021-09-02
[pwd] redo per BSD spec

fwesss:
pwd spec was written according to GNU spec so there were some incorrect options.

Re-wrote according to the BSD version (`man pwd` on macOS).


fedeci (approved):

mschrage (approved):
Good catch, Wes!

PR #555, opened 2021-09-02
[git] stash pop arg fix

fwesss:
git stash pop takes an optional arg. However the spec had an insertValue of 'pop {cursor}' which made it seem as if the arg was mandatory.

This makes it so completing on pop will not add a space or make fig popup again.


mschrage (approved):
I have been meaning to fix this for so long!!!! Thanks for getting it done @fwesss

PR #556, opened 2021-09-03
feat: add completion spec for kubecolor

jb3:

Adds support for [kubecolor](https://github.com/dty1er/kubecolor), a tool which colorizes kubectl output. It takes identical arguments to kubectl as under the hood it just invokes kubectl and reformats the output.

No completions are given for kubecolor or kubectl when kubecolor is used (assuming it is aliased to kubectl).

Support completions for both `kubecolor` and an alias from `kubectl` to `kubecolor`.

Screenshot 2021-09-03 at 02 30 23

**Additional info:**

This PR just re-exports the completions from kubectl as kubecolor.

### DIFF
--- a/dev/kubecolor.ts
+++ b/dev/kubecolor.ts
@@ -1,0 +1,2 @@
+// Kubecolor (https://github.com/dty1er/kubecolor) takes identical arguments to kubectl.
+export { default } from "./kubectl";


mschrage (approved):
What a layup PR! (It's always nice to have an easy one to review)

Thanks so much for contributing @jb3 🎉

dev/kubecolor.ts
```diff
@@ -0,0 +1,2 @@
+// Kubecolor (https://github.com/dty1er/kubecolor) takes identical arguments to kubectl.
+export { default } from "./kubectl";
```
fedeci, line 2: Should we update the name and description of the command here, or they are the same as kubectl?

jb3: Typically, kubecolor is used in the exact same fashion as kubectl, in fact kubecolor advises you alias `kubectl` to `kubecolor`, all parameters and documentation are the same, it just passes through all user options to kubectl and recolors the output.

PR #557, opened 2021-09-03
Add linting for unsupported browser APIs

mschrage:
Feature
Completion specs can use Browser APIs that are supported in newer versions of WebKit, but break when run on older machines.
Add `eslint-compat-plugin` to check for the use browser APIs that are not supported across all necessary versions of Safari (WKWebview).


fedeci (approved):
Great! I'll consider adding usage based polyfills too.

QuiiBz (approved):
LGTM!

PR #559, opened 2021-09-04
feat: add deno completion spec

SeparateRecords:
Feature: Adds a completion spec for the Deno CLI

**Additional info**
Fig is currently missing support for options that _require_ an = for _optional_ arguments (for options with required arguments, `insertValue` does a good enough job). Deno uses this quite heavily, eg. `--allow-read` or `--allow-read=README.md`, but not `--allow-read README.md`

For these options, instead of providing _incorrect_ argument suggestions, the argument is commented out since it's optional anyway. This can be changed, though.

### DIFF
--- a/dev/deno.ts
+++ b/dev/deno.ts
@@ -183,12 +183,13 @@ const seedArg: Fig.Option = {
 
 const v8FlagsOption: Fig.Option = {
   name: "--v8-flags",
+  insertValue: "--v8-flags=",
   description: "Set V8 command line options (for help: --v8-flags=--help",
   // requiresEquals: true,
-  // args: {
-  //   name: "V8 flags",
-  //   description: "Flags to pass to V8",
-  // },
+  args: {
+    name: "V8 flags",
+    description: "Flags to pass to V8",
+  },
 };
 
 const watchOption: Fig.Option = {
@@ -292,11 +293,12 @@ const denoTest: Fig.Subcommand = {
     ...runtimeOptions({ perms: true, inspector: true }),
     {
       name: "--ignore",
+      insertValue: "--ignore=",
       description: "Ignore files",
       // requiresEquals: true,
-      // args: {
-      //   name: "Files to ignore",
-      // },
+      args: {
+        name: "Files to ignore",
+      },
     },
     {
       name: "--no-run",
@@ -339,12 +341,13 @@ const denoTest: Fig.Subcommand = {
     },
     {
       name: "--coverage",
+      insertValue: "--coverage=",
       description: "UNSTABLE: Collect coverage profile data into the directory",
       // requiresEquals: true,
-      // args: {
-      //   name: "directory",
-      //   description: "The directory to use for coverage data",
-      // },
+      args: {
+        name: "directory",
+        description: "The directory to use for coverage data",
+      },
     },
     {
       name: ["-j", "--jobs"],
@@ -391,12 +394,13 @@ const denoFmt: Fig.Subcommand = {
     },
     {
       name: "--ignore",
+      insertValue: "--ignore=",
       description: "Ignore formatting particular source files",
       // requiresEquals: true,
-      // args: {
-      //   name: "Files to ignore",
-      //   template: "filepaths",
-      // },
+      args: {
+        name: "Files to ignore",
+        template: "filepaths",
+      },
     },
     watchOption,
   ],
@@ -589,32 +593,35 @@ const denoCoverage: Fig.Subcommand = {
     ...globalOptions,
     {
       name: "--ignore",
+      insertValue: "--ignore=",
       description: "Ignore coverage files",
       // requiresEquals: true,
-      // args: {
-      //   name: "pattern",
-      //   description: "Ignore files matching this regex pattern",
-      // },
+      args: {
+        name: "pattern",
+        description: "Ignore files matching this regex pattern",
+      },
     },
     {
       name: "--include",
+      insertValue: "--include=",
       description: "Include source files in the report",
       isRepeatable: true,
       // requiresEquals: true,
-      // args: {
-      //   name: "pattern",
-      //   description: "Include files matching this regex pattern",
-      // },
+      args: {
+        name: "pattern",
+        description: "Include files matching this regex pattern",
+      },
     },
     {
       name: "--exclude",
+      insertValue: "--exclude=",
       description: "Exclude source files from the report",
       isRepeatable: true,
       // requiresEquals: true,
-      // args: {
-      //   name: "pattern",
-      //   description: "Exclude files matching this regex pattern",
-      // },
+      args: {
+        name: "pattern",
+        description: "Exclude files matching this regex pattern",
+      },
     },
     {
       name: "--lcov",
--- a/dev/deno.ts
+++ b/dev/deno.ts
@@ -1,0 +1,787 @@
+// This file largely follows the same structure (not order) as deno/cli/flags.rs:
+// https://github.com/denoland/deno/blob/main/cli/flags.rs
+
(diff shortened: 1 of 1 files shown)


dev/deno.ts
```diff
@@ -0,0 +1,794 @@
...
+  options: [...globalOptions, ...compileOptions],
+};
+
+const denoBundle: Fig.Subcommand = {
+  name: "bundle",
+  description: "Bundle module and dependencies into a single file",
+  args: [
+    {
+      name: "source",
+      description: "The source file to bundle",
```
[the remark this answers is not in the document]
SeparateRecords: Great point! I didn't even think of that since it can take URLs too, but that doesn't mean it shouldn't suggest files!

dev/deno.ts
```diff
@@ -0,0 +1,794 @@
...
+    },
+    {
+      name: "output",
+      description: "The output file",
```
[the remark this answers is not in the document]
SeparateRecords: Sure, I'll go through and check that there's a generator/template everywhere there can be.

dev/deno.ts
```diff
@@ -0,0 +1,794 @@
...
+  name: "info",
+  description: "Show info about the cache or info related to a source file",
+  options: [
+    ...globalOptions,
+    reloadOption,
+    caFileOption,
+    {
+      ...locationOption,
+      description: "Show files used for origin-bound APIs (like Web Storage)",
+    },
```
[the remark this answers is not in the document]
SeparateRecords: I implemented this locally but I think the root problem here is that the `--location` in `deno info` is a different instruction that shares the same name/argument.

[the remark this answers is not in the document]
SeparateRecords: I've converted this to an object literal in the latest commit (9f6bd9a) - let me know if you're happy with that or you'd rather it be a function :)

SeparateRecords:
There's a few `// TODO`s littered around there (basically just for suggestion improvements).
I'm intending on resolving those in a follow-up PR, but I'm happy to implement them in this one if that would be better.

fedeci:
No problem in implementing those in a follow up PR!

dev/deno.ts
```diff
@@ -0,0 +1,840 @@
...
+        template: "folders",
+      },
+    },
+  ],
+};
+
+const denoLsp: Fig.Subcommand = {
+  name: "lsp",
+  description: "Start the language server",
+  options: [...globalOptions],
```
fedeci, line 532:
```suggestion
  options: globalOptions,
```

dev/deno.ts
```diff
@@ -0,0 +1,840 @@
...
+};
+
+const denoTypes: Fig.Subcommand = {
+  name: "types",
+  description: "Print Deno's runtime TypeScript declarations",
+  options: [...globalOptions],
```
fedeci, line 538:
```suggestion
  options: globalOptions,
```

dev/deno.ts
```diff
@@ -0,0 +1,840 @@
...
+
+const denoCompletions: Fig.Subcommand = {
+  name: "completions",
+  description: "Generate shell completions",
+  args: {
+    name: "shell",
+    description: "Name of the shell to generate completions for",
+    // TODO: Suggest shells
+  },
+  options: [...globalOptions],
```
fedeci, line 611:
```suggestion
  options: globalOptions,
```

dev/deno.ts
```diff
@@ -0,0 +1,840 @@
...
+    name: "script",
+    description: "The JavaScript or TypeScript file to run",
+    template: "filepaths",
+  },
+  options: [
+    ...runtimeOptions({ perms: true, inspector: true }),
+    {
+      ...watchOption,
+      exclusiveOn: ["--inspect", "--inspect-brk"],
+    },
```
fedeci, line 288: nit: this can potentially be refactored to a function accepting `exclusiveOn` array as arg

fedeci (approved):

dev/deno.ts
```diff
@@ -0,0 +1,840 @@
...
+    description: "Allow all permissions",
+  },
+  {
+    name: "--allow-env",
+    description: "Allow environment access",
+    // requiresEquals: true,
+    // args: {
+    //   name: "variable",
+    //   description: "Comma-separated list of environment variables to allow",
+    //   isOptional: true,
```
QuiiBz, line 19: Any reason why this argument and some others underneath are commented? Is it part of the `requiresEquals` stuff?

dev/deno.ts
```diff
@@ -0,0 +1,840 @@
...
+  description: "Upgrade the deno executable",
+  options: [
+    ...globalOptions,
+    {
+      name: "--version",
+      description: "The version to upgrade to",
+      args: {
+        name: "version",
+        // TODO: Suggest Deno versions, ordered most to least recent
+        // This can be gathered by querying the GitHub API
```
QuiiBz, line 552: This can be done using a [generator](https://fig.io/docs/reference/generator)

QuiiBz (changes requested):
Thanks a lot for this amazing spec!

dev/deno.ts
fedeci: He explained why in the PR description :)

dev/deno.ts
fedeci: I think we can tackle all the generators in a separate PR.

dev/deno.ts
QuiiBz: Some arguments aren't commented while the option has `requiresEquals`: https://github.com/withfig/autocomplete/pull/559/files#diff-9aeaaa6e969792dfa6d9b0c59e74739346001bb8fdba93fd96ca7b8688b61a8dR361

dev/deno.ts
QuiiBz: Of course!

dev/deno.ts
SeparateRecords: I think this table should clear it up (sorry if the formatting is wonky, commenting from my phone)

| | Requires equals | doesn't require equals |
| --- | --- | --- |
| Is optional | commented (I don't want to prescribe whether users want to pass an arg to that option) | uncommented |
| Is required | uncommented | uncommented |

dev/deno.ts
SeparateRecords: That's fair — I'll fix that up tonight when I'm back from work.

SeparateRecords:
Had more free time today than expected so I've started implementing some of the TODOs.

SeparateRecords:
Alright, with those generators in I'd be happy to say this PR is feature complete. The remaining todos are more complicated or have to wait, so I'll save _those_ for a future PR.

dev/deno.ts
```diff
@@ -0,0 +1,974 @@
...
+    ttl: 1000 * 60 * 60 * 24,
+  } as Fig.Cache,
+} as const;
+
+/**
+ * Equivalent to the `"filepaths"` template, but boosts the priority of files
+ * that match one of the names provided.
+ */
+const generatePreferredFilepaths = ({
+  names,
```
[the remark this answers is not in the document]
SeparateRecords: Thanks for pointing that out! The performance would probably be fine regardless but it's worth doing anyway.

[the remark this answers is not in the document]
fedeci: Probably yes because the commands are usually run in the root of the project which shouldn't be so polluted of files but in this case it also improves readability.

SeparateRecords:
Ready for review again :)
I'm off to bed and have tomorrow off work, hopefully we can get this finished up!

fedeci (approved):

QuiiBz (approved):
LGTM, thanks a lot for this amazing spec! Love the generators 🚀

PR #565, opened 2021-09-06
trash-cli-fix

yavko:
Add extra stuff to the trash cli spec stuff
No isDangerous or isOptional flags from some options, and fixes some descriptions
Has isDangerous or isOptional flags from some options, and descriptions are fine
**Additional info:**
6th contribution

### DIFF
--- a/dev/trash.ts
+++ b/dev/trash.ts
@@ -17,8 +17,7 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: "-F",
-      description:
-        'Use the Finder API to move items to the trash, this is slower, but most MacOS features such as sounds and the "put back" feature work',
+      description: "Use the Finder API to move items to the trash",
     },
     {
       name: "-l",
--- a/dev/trash.ts
+++ b/dev/trash.ts
@@ -21,11 +21,11 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: "-l",
-description: "List items in the trash"
+      description: "List items in the trash",
     },
     {
       name: "-e",
-description: "Empty the trash"
+      description: "Empty the trash",
       isDangerous: true,
     },
     {
--- a/dev/trash.ts
+++ b/dev/trash.ts
@@ -5,6 +5,7 @@ const completionSpec: Fig.Spec = {
   name: "trash",
   description: "Trash, move files/folders to the trash",
   args: {
+    isOptional: true,
     isVariadic: true,
     template: ["folders", "filepaths"],
   },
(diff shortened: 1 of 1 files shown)


dev/trash.ts
```diff
@@ -5,29 +5,28 @@ const completionSpec: Fig.Spec = {
...
     {
       name: "-v",
-      description:
-        "Be verbose when moving items to the trash showing them as they are moved",
+      description: "Print verbose output while moving items",
     },
     {
       name: "-F",
-      description: `Ask Finder to move the files to the trash, instead of the system API. This is slower, but it utilizes Finder's (e.g. sounds) and ensures that the "put back" feature works.`,
+      description: `Use the Finder API to move items to the trash, this is slower, but most MacOS features such as sounds and the \"put back\" feature work.`,
```
Milo123459, line 20: Not sure about the "style-guide" but I'd recommend using " over ` because you don't need to use template strings.

yavko: i used it since it previously used apostrophes I'll fix it now

dev/trash.ts
```diff
@@ -5,29 +5,28 @@ const completionSpec: Fig.Spec = {
...
       name: "-l",
-      description:
-        "List items currently in the trash. If this argument is no files need to be specified",
+      description: "List items currently in the trash",
     },
     {
       name: "-e",
-      description:
-        "Empty the trash. trash asks for confirmation before this action. If this argument is used, no files need to specified",
+      description: "Empty the trash asking for confirmation",
```
Milo123459, line 28:
```suggestion
description: "Empty the trash"
```

IMHO, you don't need the confirmation part.

dev/trash.ts
yavko: and quotes

dev/trash.ts
```diff
@@ -5,29 +5,28 @@ const completionSpec: Fig.Spec = {
...
     {
       name: "-F",
-      description: `Ask Finder to move the files to the trash, instead of the system API. This is slower, but it utilizes Finder's (e.g. sounds) and ensures that the "put back" feature works.`,
+      description: `Use the Finder API to move items to the trash, this is slower, but most MacOS features such as sounds and the \"put back\" feature work.`,
     },
     {
       name: "-l",
-      description:
-        "List items currently in the trash. If this argument is no files need to be specified",
+      description: "List items currently in the trash",
```
Milo123459, line 24:
```suggestion
description: "List items in the trash"
```

dev/trash.ts
Milo123459, line 20:
```suggestion:
	description: "Use the Finder API to move items to the trash"
```

fedeci: Even whenusing quotes or double quotes always try to avoid (escaping chars) template literals because there is a lint rule that is not yet working with it.

fedeci: I like the part where he explains what drawbacks and features it bring, I don't think it should be removed.

yavko: ok thats is fixed now!

yavko: Idk what do to
i already pushed lemme get opinions

Milo123459: > I like the part where he explains what drawbacks and features it bring, I don't think it should be removed. +1 for the double quotes

Personally i think it's better clean.
This could be an idea for something like a note section.

Milo123459:
Looks like more people are for the change.

yavko:
> Looks like more people are for the change.

Indeed so, @QuiiBz or @mschrage can you guys review and merge? I think we have reached a conclusion.

QuiiBz (approved):
LGTM, thanks for those small changes!

Issue #418, opened 2021-07-29
feat: add support for restricting commands by OS


EvanBacon:
### Sanity checks

- [X] I have searched [github.com/withfig/fig/issues](https://github.com/withfig/fig/issues?q=) and there are no duplicates of my issue
- [X] My issue does not relate to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If it does, please create your issue here: [withfig/autocomplete](https://github.com/withfig/autocomplete/issues?q=)

### Summary

I'd like to be able to prevent the expo-cli schema from showing options like `run:ios` on windows machines (where xcode isn't supported). It'd be nice to be able to show some restriction message. Perhaps something like a media query for the description field which can target win32. 

### Environment

```bash
Version 1.0.46 (B250) [Dev] [U.S.] 
UserShell: /bin/zsh
Bundle path: /Applications/Fig.app
Autocomplete: true
Settings.json: true
CLI installed: true
CLI tool path: /Users/evanbacon/.fig/bin/fig
Accessibility: true
Number of specs: 151
SSH Integration: true
Tmux Integration: true
Keybindings path: /Users/evanbacon/.fig/user/keybindings
iTerm Integration: true 
Hyper Integration: true
VSCode Integration: true
Docker Integration: false
Symlinked dotfiles: false
Only insert on tab: false
Installation Script: true
PseudoTerminal Path: /usr/local/opt/ruby@2.7/bin:/Users/evanbacon/.nvm/versions/node/v14.15.5/bin:/Users/evanbacon/.nix-profile/bin:/Users/evanbacon/.fastlane/bin:/Users/evanbacon/.yarn/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin:/Users/evanbacon/.fig/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin:/Users/evanbacon/Library/Android/sdk/emulator:/Users/evanbacon/Library/Android/sdk/tools:/Users/evanbacon/Library/Android/sdk/tools/bin:/Users/evanbacon/Library/Android/sdk/platform-tools:/Users/evanbacon/Library/Android/sdk/ndk-bundle:/Users/evanbacon/Documents/flutter/bin
SecureKeyboardInput: false
SecureKeyboardProcess: <none>
Current active process: /bin/zsh (18456) - ttys021
Current working directory: /Users/evanbacon/Documents/GitHub/lab/yolo37
[...]

PR #566, opened 2021-09-07
feat: add xargs completion spec

SeparateRecords:
Feature: completion spec for `xargs`

**Additional info:**
This is BSD xargs (since Fig is macOS only right now) - a port to Linux, where GNU is the norm, would require changing this.

### DIFF
--- a/dev/xargs.ts
+++ b/dev/xargs.ts
@@ -1,0 +1,121 @@
+const completionSpec: Fig.Spec = {
+  name: "xargs",
+  description:
+    "Execute a command with whitespace-delimited strings (from stdin) as arguments",
+  options: [
+    {
+      name: "-0",
+      description:
+        "Change xargs to expect NUL (0x00) characters as separators, instead of newlines and spaces. This is expected to be used with the -print0 function of find(1)",
+    },
+    {
+      name: "-E",
+      description: "Use this string as a logic EOF marker",
+      args: {
+        name: "eof-str",
+        description: "The string to use that marks EOF",
+      },
+    },
+    {
+      name: "-I",
+      description: "Replace occurrences of this string with the input",
+      args: {
+        name: "replacement-str",
+        description: "The string to replace",
+      },
+    },
+    {
+      name: "-J",
+      description:
+        "Replace an argument exactly equal to this string with the input",
+      args: {
+        name: "replacement-str",
+        description: "The string to replace",
+      },
+    },
+    {
+      name: "-L",
+      description:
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
According to [this](https://man7.org/linux/man-pages/man1/xargs.1.html) man page there are some long-option aliases, can you add those?

SeparateRecords:
> According to [this](https://man7.org/linux/man-pages/man1/xargs.1.html) man page there are some long-option aliases, can you add those?

There are in the GNU version, but there aren't on macOS since it uses modified BSD programs (which is closer to the POSIX spec): https://opensource.apple.com/source/shell_cmds/shell_cmds-149/xargs/xargs.c.auto.html

The GNU and BSD versions share a subset of options.

fedeci:
That's something that should be solved.

SeparateRecords:
So to clarify, are you saying to remove non-POSIX options that _do_ exist on macOS or add GNU extensions that _don't_ exist on macOS? I think, given that Fig currently only supports macOS, it's better to leave it as-is.

Looking at precedent, the [`ls` spec](https://github.com/withfig/autocomplete/blob/master/dev/ls.ts) doesn't include the GNU long options ([Linux man page](https://man7.org/linux/man-pages/man1/ls.1.html)), same thing with `cp`. Haven't checked any others.

fedeci:
I mean to fix it a deeper level beginning to allow commands/options...etc to be specific for some platforms
/cc @mschrage

SeparateRecords:
Agreed - I think it would be good to have a plan in place now for platform-specific completions, since that will have to be dealt with at some point. If we cross that bridge now, it'll be a lot easier.

I'm happy to either wait with this PR until there's a solution in place, or to take on the burden of updating it later.

QuiiBz:
> I mean to fix it a deeper level beginning to allow commands/options...etc to be specific for some platforms

 We're only focusing on macOS for now, but that's definitely something that will be implemented in the future, as Fig will be available on others platforms.

Based on your initial comment @SeparateRecords, this spec is BSD-compliant, so it's all fine!

SeparateRecords:
Fixed a minor wording issue that was nagging at me. Ready for review any time :)

SeparateRecords:
Thanks for the input @mschrage! I think that would definitely work, some sugar over that would be lovely. I see #418 is open - I'll leave some thoughts over there :)

QuiiBz (approved):
LGTM, thanks a lot for this spec 🚀

PR #568, opened 2021-09-07
feat(deno): add generator for doc filters

SeparateRecords:
Feature: Improves `deno doc` suggestions 🎉

---

<details>
<summary><b>Screenshots</b> (warning: very compelling)</summary>
<br>

_Fig. 1. Top-level results for a file_
Results for `deno doc src/models.ts`

_Fig. 2. Results for members of `API`_
Results for `deno doc src/models.ts API.`

_Fig. 3. Private symbols work, too!_
Results for `deno doc --private scripts/redact.ts`

Aside: I can't use `src/models.ts` with `--private` because it causes a stack overflow in Deno! So that's fun.

</details>

---

Suggests files to doc, doesn't suggest a filter.

Uses the scope (the file) to suggest a filter as you type. It really does Just Work(tm).

**Additional info:**
It's a relatively complicated generator but it runs decently fast. It also isn't extracted to its own variable, because it isn't usable outside of this one particular command. Having the extra context above it makes it significantly easier to understand.

I basically never use `deno doc` because searching it is a pain and I can't remember what I named that type... this generator just suggests what I'm looking for while I'm typing!

**What I need advice on:**
The icon used for suggestions is currently `fig://icon?type=string`, which isn't _super_ fitting. I don't intend on going full IntelliSense with the icons, though.

### DIFF
--- a/dev/deno.ts
+++ b/dev/deno.ts
@@ -768,24 +768,17 @@ const denoDoc: Fig.Subcommand = {
           // The names are added to a set because duplicates are common.
           const names = new Set<string>();
 
-          // More nodes should be visible if --private was used. This is two
-          // separate loops because the logic is fundamentally different - one
-          // loop just adds all the nodes, while the other is a filter. It also
-          // avoids checking showPrivateSymbols in each iteration, since that
-          // value won't change.
-          if (showPrivateSymbols) {
-            for (const node of suggestNodes) {
+          // More nodes should be visible if --private was used.
+          for (const node of suggestNodes) {
+            if (showPrivateSymbols) {
               names.add(node.name);
+              continue;
             }
-          } else {
-            for (const node of suggestNodes) {
-              // Deno includes imports in the JSON regardless of privacy,
-              // probably so it's easier to build a module graph. However,
-              // imports are (by definition) private, so they must be removed.
-              if (node.kind === "import") continue;
-
-              names.add(node.name);
-            }
+            // Deno includes imports in the JSON regardless of privacy,
+            // probably so it's easier to build a module graph. However,
+            // imports are (by definition) private, so they must be removed.
+            if (node.kind === "import") continue;
+            names.add(node.name);
           }
 
           // Deno uses the name <TODO> when it gets confused, which is common in
--- a/dev/deno.ts
+++ b/dev/deno.ts
@@ -581,11 +581,36 @@ const denoLint: Fig.Subcommand = {
   ],
 };
 
-// The following region was adapted from denoland/deno_doc:
-// https://github.com/denoland/deno_doc/blob/main/lib/types.d.ts
-// It is not stripped further to make updating it easier.
-
 //#region Documentation types
+
+type Named = { name: string };
+type Kind<T = string> = { kind: T };
+
+type DocNodeFunction = Named & Kind<"function">;
+type DocNodeVariable = Named & Kind<"variable">;
+type DocNodeEnum = Named & Kind<"enum">;
+type DocNodeClass = Named &
+  Kind<"class"> & {
+    classDef: {
+      properties: Named[];
+      methods: (Named & Kind)[];
+    };
+  };
+type DocNodeTypeAlias = Named & Kind<"typeAlias">;
+type DocNodeNamespace = Named &
+  Kind<"namespace"> & {
+    namespaceDef: {
+      elements: DocNode[];
+    };
+  };
+type DocNodeInterface = Named &
+  Kind<"interface"> & {
+    interfaceDef: {
+      properties: Named[];
+      methods: (Named & Kind)[];
+    };
+  };
+type DocNodeImport = Named & Kind<"import">;
 
 type DocNode =
   | DocNodeFunction
@@ -596,95 +621,6 @@ type DocNode =
   | DocNodeNamespace
   | DocNodeInterface
   | DocNodeImport;
-
-interface DocNodeBase {
-  kind: DocNodeKind;
-  name: string;
-}
-
-type DocNodeKind =
-  | "function"
-  | "variable"
-  | "enum"
-  | "class"
-  | "typeAlias"
-  | "namespace"
-  | "interface"
-  | "import";
-
-interface DocNodeFunction extends DocNodeBase {
-  kind: "function";
-}
-
-interface DocNodeVariable extends DocNodeBase {
-  kind: "variable";
-}
-
-interface DocNodeEnum extends DocNodeBase {
-  kind: "enum";
-  enumDef: EnumDef;
-}
-
-interface DocNodeClass extends DocNodeBase {
-  kind: "class";
-  classDef: ClassDef;
-}
-
-interface DocNodeTypeAlias extends DocNodeBase {
-  kind: "typeAlias";
-}
-
-interface DocNodeNamespace extends DocNodeBase {
-  kind: "namespace";
-  namespaceDef: NamespaceDef;
-}
-
-interface DocNodeInterface extends DocNodeBase {
-  kind: "interface";
-  interfaceDef: InterfaceDef;
-}
-
-interface DocNodeImport extends DocNodeBase {
-  kind: "import";
-}
-
-interface ClassDef {
-  properties: ClassPropertyDef[];
-  methods: ClassMethodDef[];
-}
-
-interface ClassMethodDef {
-  name: string;
-}
-
-interface ClassPropertyDef {
-  name: string;
-}
-
-interface EnumDef {
-  members: EnumMemberDef[];
-}
-
-interface EnumMemberDef {
-  name: string;
-}
-
-interface InterfaceDef {
-  methods: InterfaceMethodDef[];
-  properties: InterfacePropertyDef[];
-}
-
-interface InterfaceMethodDef {
-  name: string;
-}
-
-interface InterfacePropertyDef {
-  name: string;
-}
-
-interface NamespaceDef {
-  elements: DocNode[];
-}
 
 //#endregion
 
--- a/dev/deno.ts
+++ b/dev/deno.ts
@@ -1117,10 +1117,12 @@ const completionSpec: Fig.Spec = {
     {
       name: "--version",
       description: "Prints version information, including TypeScript and V8",
+      exclusiveOn: ["-V"],
     },
     {
       name: "-V",
       description: "Prints Deno's version",
+      exclusiveOn: ["--version"],
       priority: 25,
     },
   ],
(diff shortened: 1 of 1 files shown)


dev/deno.ts
```diff
@@ -579,6 +579,14 @@ const denoLint: Fig.Subcommand = {
   ],
 };
 
+function docNameToSuggestion(name: string): Fig.Suggestion {
```
SeparateRecords, line 582: This function is poorly named.

dev/deno.ts
```diff
@@ -600,11 +608,101 @@ const denoDoc: Fig.Subcommand = {
...
+      // This generator helps you construct a filter by suggesting top level
+      // nodes from the provided scope, and more specific nodes after a period
+      generators: {
+        // Options can be provided in any order, so the second last element
+        // isn't guaranteed to be the scope. The solution is to modify the
+        // array of tokens to insert --json. However, since there can only be
+        // one occurrence of that flag, it has to be guarded.
+        script: (tokens) => {
+          // The last element is always the `filter`, which must be removed
+          const commandUntilNode = tokens.slice(0, -1);
```
SeparateRecords, line 623: Still refers to "node" instead of filter.

SeparateRecords:
`--private` has finally stopped crashing the cli which has allowed me to find another issue, needs to be resolved before this can be merged

SeparateRecords:
Should be fine now.

SeparateRecords:
I've decided to restructure the generator. It's easier to understand and should be about as efficient, but it's a little bit longer. I'm not near a Mac tonight so I can't test it. This PR has been set to a draft until I've had the chance to test it and update my branch.

SeparateRecords:
(reminder for myself later)
There's a better URL for the versions JSON, so that generator may as well be updated too.

`curl -sLF https://cdn.deno.land/std/meta/versions.json`

SeparateRecords:
Better URL for the `deno upgrade --version` generator, may as well update this later too:
`curl -sL https://cdn.deno.land/deno/meta/versions.json`

dev/deno.ts
```diff
@@ -600,11 +620,142 @@ const denoDoc: Fig.Subcommand = {
...
+        //    b. If there is a dot, everything up to the dot becomes the name.
+        //       All nodes with that name are collected, and the array is
+        //       populated with their children.
+        // 2. Those nodes are filtered, and their names are added to a set.
+        // 3. Those names in that set are transformed into suggestions.
+        postProcess: (out, tokens) => {
+          // The output for `deno doc --json` is `DocNode[]` - the types:
+          // https://github.com/denoland/deno_doc/blob/dbf9e21/lib/types.d.ts
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          let docNodes: any[];
```
QuiiBz, line 668: Would it be possible to type properly this variable?

dev/deno.ts
```diff
@@ -600,11 +620,142 @@ const denoDoc: Fig.Subcommand = {
...
+          } else {
+            const filterName = filterToken.slice(0, firstDotIndex);
+
+            // It's not uncommon that there'd be multiple occurrences of the
+            // same name with different values, eg. overloads and interface
+            // merging. Deno's builtin types actually do this with the `Deno`
+            // interface. Typically this will only be one or two nodes.
+            const foundNodes = docNodes.filter((node) => {
+              return node.name === filterName;
+            });
```
QuiiBz, line 704: We can simplify this a bit:
```suggestion
            const foundNodes = docNodes.filter(({ name }) => name === filterName);
```

dev/deno.ts
```diff
@@ -600,11 +620,142 @@ const denoDoc: Fig.Subcommand = {
...
+            }
+          } else {
+            for (const node of suggestNodes) {
+              // Deno includes imports in the JSON regardless of privacy,
+              // probably so it's easier to build a module graph. However,
+              // imports are (by definition) private, so they must be removed.
+              if (node.kind === "import") continue;
+              suggestionNames.add(node.name);
+            }
+          }
```
QuiiBz, line 743: We can avoid duplicating the `for` loop to make this a bit cleaner:
```suggestion
          if (showPrivateSymbols) {
          for (const node of suggestNodes) {
            // Deno includes imports in the JSON regardless of privacy,
            // probably so it's easier to build a module graph. However,
            // imports are (by definition) private, so they must be removed.
            if (!showPrivateSymbols && node.kind === "import") continue;
            suggestionNames.add(node.name);
          }
```

QuiiBz (changes requested):
Thanks a lot for this amazing stuff! Left some comments to improve the spec before merging it

dev/deno.ts
SeparateRecords: Personally I'd like that! However, that line goes over 80 characters, so the pre-commit formatter puts the condition on the line below. Reading `const foundNodes = docNodes.filter((node) =>` feels awkward, like I've been left hanging, waiting for an expression, a block... something.
Here's what I prefer:
a. The entire callback goes on the next line, which is what it was before (but it's awkward to read)
b. The callback starts on the same line as `.filter` and uses a block instead of an expression

Regardless of what we end up with, it'll be a purely stylistic choice - semantically it's identical. Lmk your thoughts!

dev/deno.ts
SeparateRecords: It's cleaner, but it ends up checking the `showPrivateSymbols` on each iteration instead of just once. That's probably an inconsequential difference, but it's a bunch of work that _can_ be skipped. If you think the trade off is worth it I'm happy to implement!

dev/deno.ts
SeparateRecords: Initially I'd copied the type definitions into the source, but it's a _pretty big_ file. Next time I'm around a Mac I'll try to gut it to include only the minimum type information :)

QuiiBz: Thanks!

dev/deno.ts
QuiiBz: Looking at the recent commits, I find the new solution pretty clean - no need to wrap the callback to a new line while being a single-liner, that's sweet! We can keep it as it is right now.

dev/deno.ts
QuiiBz: How many `node` can be contained inside `suggestNodes`? I don't think that an `if` would cause any issues even if there are thousands of nodes, and I found to not repeat the `for` loop is cleaner. What do you think of it?

dev/deno.ts
```diff
@@ -600,11 +663,146 @@ const denoDoc: Fig.Subcommand = {
...
+          } else {
+            for (const node of suggestNodes) {
+              // Deno includes imports in the JSON regardless of privacy,
+              // probably so it's easier to build a module graph. However,
+              // imports are (by definition) private, so they must be removed.
+              if (node.kind === "import") continue;
+
+              names.add(node.name);
+            }
+          }
```
fedeci, line 789: Is it really worth it? I mean, we are checking for a boolean value which is the simplest operation.

fedeci (approved):

dev/deno.ts
SeparateRecords: I honestly do think that the intention becomes less clear. Don't get me wrong, I don't have a problem with changing this, but it's just less obvious (to me personally) that there are really two separate things happening.

This is what is really going on, and the reason I think the two loops is clearer.
It's putting the names directly into the set to skip creating yet another array.
```typescript
let dups: string[];
if (showPrivateSymbols) {
  dups = suggestNodes
    .map((node) => node.name);
} else {
  dups = suggestNodes
    .filter((node) => node.kind !== "import")
    .map((node) => node.name);
}
const names = new Set(dups);
```

I've written out the suggested code so I can add it when I'm home later:

```suggestion
          for (const node of suggestNodes) {
            if (showPrivateSymbols) {
              names.add(node.name);
              continue;
            }
            // Deno includes imports in the JSON regardless of privacy,
            // probably so it's easier to build a module graph. However,
            // imports are (by definition) private, so they must be removed.
            if (node.kind === "import") continue;

            names.add(node.name);
          }
```

dev/deno.ts
```diff
@@ -72,14 +73,12 @@ function generateFilepathsMatch(init: {
...
+  script: `curl -sL '${VERSIONS_URL}'`,
   cache: { ttl: 1000 * 60 * 60 * 24 },
   postProcess: (out) => {
     try {
-      const versions = JSON.parse(out);
-      return (versions.cli as string[]).map((version) => ({
-        // Currently, the JSON does have a leading 'v', but that may not always
-        // be the case. Removing a leading 'v' instead of slicing is resilient.
+      const data = JSON.parse(out);
+      return (data.versions as string[]).map((version) => ({
```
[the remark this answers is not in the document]
SeparateRecords: Marginally, yeah. Ran a benchmark with 1000 items in the array: the array-map-regex-replace takes about 0.05ms, whereas array-map-slice takes about 0.015. It's an even less meaningful difference with the actual versions data (currently 148 tags)

I think that the resilience of the regex is worth the additional 0.035ms. If a future version is tagged without the leading `v`, it will be fine. With a slice, the leading digit will be removed if there's no 'v'.

I didn't notice I'd deleted the comment explaining my reasoning, though, so that's good to know.

dev/deno.ts
```diff
@@ -768,24 +780,17 @@ const denoDoc: Fig.Subcommand = {
...
-          } else {
-            for (const node of suggestNodes) {
-              // Deno includes imports in the JSON regardless of privacy,
-              // probably so it's easier to build a module graph. However,
-              // imports are (by definition) private, so they must be removed.
-              if (node.kind === "import") continue;
-
+          // More nodes should be visible if --private was used.
+          for (const node of suggestNodes) {
+            if (showPrivateSymbols) {
```
fedeci, line 793: I even prefer refactoring again to join that call.
```suggestion
            if (showPrivateSymbols || node.kind !== "import") {
```

fedeci (approved):

dev/deno.ts
SeparateRecords: I think leaving it as-is makes it much easier to read, even if it's longer. @QuiiBz thoughts?

SeparateRecords:
`deno` 1.14 is being released tomorrow, so best to hold off on this PR. It includes the lastest `deno_doc` that outputs another top level kind of doc node (moduleDoc) that will have to be added to the types and filtered out. I'll test tonight in the Canary build and update when I know this works.

SeparateRecords:
Looks like it's all working to me. Hopefully we can finish this up today/tomorrow!

fedeci:
Thanks @SeparateRecords, it seems ok now! I will re-review the PR later to see the other changes

QuiiBz (approved):
LGTM, love this new generator 🎉

PR #573, opened 2021-09-08
[FIG-TEAMS] add initial spec

cstrnt:
Feature

The Spec doesn't exist

There is a spec for fig-teams

**Additional info:**


QuiiBz (approved):
Everything looks good, excited to have Fig for teams! Feel free to merge when needed

dev/npx.ts
```diff
@@ -45,6 +45,10 @@ const suggestions: Fig.Suggestion[] = [
     name: "typeorm",
     icon: "https://avatars.githubusercontent.com/u/20165699?s=200&v=4",
   },
+  {
+    name: "fig-teams",
+    icon: "https://fig.io/icons/fig-light.png",
+  },
```
fedeci, line 51: Yarn needs this one too

QuiiBz: I don't think it's needed: `yarn <package> ...` only runs if you are currently in a project that has this `package` in its dependencies, instead of `npx` which allows you to run any package anywhere on your system.

fedeci: We should remove `create-next-app`, `create-react-native-app`, and `create-react-app` from there, then.

PR #574, opened 2021-09-08
[INTERNAL] update spec format to allow versioning

cstrnt:

**Additional info:**


schemas/fig.d.ts
```diff
@@ -29,8 +29,14 @@ declare namespace Fig {
   // both set to void by default
   export type StringOrFunction<T = void, R = void> = string | Function<T, R>;
 
+  export type VersionedSpec = (version: string | undefined) => Spec;
```
QuiiBz, line 32: This can be simplified a bit,  maybe overkill:
```suggestion
  export type VersionedSpec = (version?: string) => Spec;
```

fedeci:
We partially implemented this PR in the types and I would close this one since we still need to elaborate how versioning should be made properly.
/cc @mschrage

mschrage:
Makes sense @fedeci!

PR #579, opened 2021-09-10
Include both `HEAD` and `HEAD~` as git suggestions

peterkhayes:

Bug fix / polish

Some git commands suggest `HEAD` as an argument; others suggest `HEAD~`.

This standardizes all git commands that suggest one of these to suggest both. In particular, this resolves a nasty issue where `git reset --hard` ONLY suggested `HEAD~`. As a result, by hitting enter twice, the user would automatically throw away the latest commit.

image

**Test plan:**
I tried the commands in my terminal and everything seemed to work.

### DIFF
--- a/dev/git.ts
+++ b/dev/git.ts
@@ -471,11 +471,18 @@ const addOptions: Fig.Option[] = [
   },
 ];
 
-const head = {
-  name: "HEAD",
-  icon: "🔻",
-  description: "Reset multiple commits",
-};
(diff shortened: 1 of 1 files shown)


mschrage (approved):
Thanks so much for contributing @peterkhayes ! 🎉

I'm going to add you to with withfig org, so you can add a badge to your Github profile if you'd like :)

PR #581, opened 2021-09-10
[sfdx] spec generated

fwesss:
Salesforce CLI spec generated using https://github.com/withfig/oclif-plugin

https://developer.salesforce.com/tools/sfdxcli


mschrage (approved):
LGTM 🤩

PR #582, opened 2021-09-13
Bazel-update description and args

ly0708:

Updated the description in bazel.ts making it more clear.
Changed using args instead of insertValue. 

The description is not clear. 

**Additional info:**

### DIFF
--- a/dev/bazel.ts
+++ b/dev/bazel.ts
@@ -73,7 +73,8 @@ const completionSpec: Fig.Spec = {
     {
       name: "--noautodetect_server_javabase",
       exclusiveOn: ["autodetect_server_javabase"],
-      description: "Opposite of --autodetect_server_javabase",
+      description:
+        "Does not fall back to the local JDK for running the bazel server and instead exits",
     },
     {
       name: "--batch",
@@ -83,7 +84,7 @@ const completionSpec: Fig.Spec = {
     {
       name: "--nobatch",
       exclusiveOn: ["--batch"],
-      description: "Opposite of --batch",
+      description: "Run with a server",
     },
     {
       name: "--batch_cpu_scheduling",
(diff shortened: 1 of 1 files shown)


dev/bazel.ts
```diff
@@ -273,7 +293,7 @@ const completionSpec: Fig.Spec = {
       name: "--nounlimit_coredumps",
       exclusiveOn: ["--unlimit_coredumps"],
       description:
-        "Raises the soft coredump limit to the hard limit to make coredumps of the server (including the JVM) and the client possible under common conditions",
+        "Don't raises the soft coredump limit to the hard limit to make coredumps of the server (including the JVM) and the client possible under common conditions",
```
fedeci, line 296: typo:
```suggestion
        "Don't raise the soft coredump limit to the hard limit to make coredumps of the server (including the JVM) and the client possible under common conditions",
```

dev/bazel.ts
```diff
@@ -239,8 +256,10 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: "--server_jvm_out",
-      insertValue: "--server_jvm_out={cursor}",
       description: "The location to write the server's JVM's output",
+      args: {
+        name: "Path",
```
fedeci, line 261: Does it accept a `filepaths` template?

ly0708: I am sure here is the doc - 
`--server_jvm_out=<path> default: see description
The location to write the server's JVM's output. If unset then defaults to a location in output_base.
Tags: affects_outputs, loses_incremental_state`

ly0708: What do you mean by filepaths?

fedeci: Some path on the machine on which the command is run onto.

fedeci: Like `~/Desktop/logs/out.log`

ly0708: I see. then I think it does accept "filepaths template".

fedeci (approved):
Thanks!

QuiiBz (approved):
LGTM, thanks a lot for those amazing new descriptions!

PR #585, opened 2021-09-15
feat(deno): update spec to 1.14

SeparateRecords:
This PR adds support for new flags added in Deno CLI 1.14, released earlier today.
This includes 1 new generator (`generateLintRules`). It also sets `isPersistent` on the global options.

Relevant issue (not a blocker): withfig/fig#454


dev/deno.ts
```diff
@@ -543,6 +569,40 @@ const denoFmt: Fig.Subcommand = {
...
+    },
+    {
+      name: "--options-single-quote",
+      description: "Use single quotes",
+    },
+    {
+      name: "--options-prose-wrap",
+      description: "Define how prose should be wrapped (default: always)",
+      args: {
+        name: "wrap",
```
fedeci, line 598:
```suggestion
        name: "wrap",
        default: "always",
```

SeparateRecords: Oh, that's neat! I didn't know about `default`. I'll play around with that later. Looking at the docs for that also lead to me finding `isPersistent`, so I'll set that on the global options. Thanks!

SeparateRecords: I've had a chance to test this now. I've added the `default` prop, but I don't think it's necessary to remove `(default: always)` from the description because the value of the `default` isn't visible in the popup. If/when that's changed, I'll fix it.
For now, it's more useful for the user to be able to see, in the description, that "always" is the default.

dev/deno.ts
```diff
@@ -95,6 +95,25 @@ const generateRunnableFiles = generateFilepathsMatch({
...
+    const suggestions: Fig.Suggestion[] = [];
+    for (const line of lines) {
+      if (line.startsWith(" - ")) {
+        suggestions.push({
+          name: line.slice(3),
+          icon: "fig://icon?type=string",
+        });
+      }
+    }
+    return suggestions;
```
QuiiBz, line 113: I think we can improve the readability of this:
```suggestion
    return out
      .split("\n")
      .filter(line => line.startsWith(" - "))
      .map(line => ({
        name: line.slice(3),
        icon: "fig://icon?type=string",
      }));
```

dev/deno.ts
QuiiBz: In the future, we might show `default` in the UI, so that would be awesome if you can keep both!

QuiiBz (changes requested):

SeparateRecords:
In implementing @QuiiBz's suggestions I realized that [`--rules` and `--json` can actually be used together](https://github.com/denoland/deno/blob/c04117134ecb07fb6379bfe1057c9da8e0c4f206/cli/tools/lint.rs#L159-L168), which was added by Bartek in denoland/deno#8384. This means we can get a description for each rule, too!

Each description is valid markdown (statically included in the binary from [this directory](https://github.com/denoland/deno_lint/tree/main/docs/rules). This means it's fine to take everything up to the first paragraph separator (`\n\n`), and use that as the description. Also, since Fig doesn't render inline code, I'm removing the surrounding back-ticks.

image

SeparateRecords:
Merge conflicts should be resolved now. Ready for another review :)

QuiiBz (approved):
LGTM, thank you so much 🎉

PR #586, opened 2021-09-15
feat: update `tailwindcss` spec to latest version

fedeci:


dev/tailwindcss.ts
```diff
@@ -32,16 +32,22 @@ const buildOptions: Fig.Option[] = [
...
   {
     name: "--purge",
+    description:
+      "Content paths to use for removing unused classes. [Deprecated]: use `--content` instead",
+  },
+  {
+    name: "--purge",
+    description:
+      "Content paths to use for removing unused classes. [Deprecated]: use `--content` instead",
+  },
```
QuiiBz, line 42: Looks like this option is duplicated

QuiiBz (changes requested):

dev/tailwindcss.ts
```diff
@@ -32,16 +32,22 @@ const buildOptions: Fig.Option[] = [
   },
   {
     name: "--purge",
+    description:
+      "Content paths to use for removing unused classes. [Deprecated]: use `--content` instead",
+  },
+  {
+    name: "--purge",
+    description:
+      "Content paths to use for removing unused classes. [Deprecated]: use `--content` instead",
```
fedeci, line 41:
```suggestion
    name: "--content",
    description: "Content paths to use for removing unused classes instead",
```

dev/tailwindcss.ts
fedeci: 🤦‍♂️

dev/tailwindcss.ts
```diff
@@ -32,16 +32,21 @@ const buildOptions: Fig.Option[] = [
...
+  },
+  {
+    name: "--content",
+    description: "Content paths to use for removing unused classes instead",
+  },
+  {
+    name: "--content",
     description: "Content paths to use for removing unused classes",
   },
```
QuiiBz, line 45: Looks like there are still two options that are the same 😅

QuiiBz (changes requested):

dev/tailwindcss.ts
fedeci: Damn😆

dev/tailwindcss.ts
```diff
@@ -32,16 +32,21 @@ const buildOptions: Fig.Option[] = [
...
   {
     name: "--purge",
+    description:
+      "Content paths to use for removing unused classes. [Deprecated]: use `--content` instead",
+  },
+  {
+    name: "--content",
+    description: "Content paths to use for removing unused classes instead",
+  },
```
fedeci, line 41:
```suggestion
```

dev/tailwindcss.ts
fedeci: Fixed

QuiiBz (approved):
LGTM 🚀

PR #587, opened 2021-09-16
Binary target

lsh:
Adding an option to deal with cargo's `--bin` option on run. This means when a `Cargo.toml` has multiple binary targets, the different targets are listed.


dev/cargo.ts
```diff
@@ -25,6 +25,18 @@ const testList: Fig.Generator = {
   getQueryTerm: ":",
 };
 
+const binList: Fig.Generator = {
+  script: function () {
+    return `cargo read-manifest`;
+  },
+  postProcess: function (data: string) {
+    const manifest = JSON.parse(data);
+    return manifest.targets
```
fedeci, line 34: Can `targets` ever be `undefined`? If so we should perform an earlier chech and return an empty array.

lsh: Pushed a fix just in case, though from what I gather having no targets will actually cause an error with cargo.

QuiiBz (approved):
Amazing generator, thanks a lot for this PR!

PR #591, opened 2021-09-17
[git-reset] Fix option args and completed spec

fwesss:
The spec for git reset used insertValue and had args for the --mixed, --merge, --soft, --hard, and --keep options. 

Only --mixed takes an arg and it's optional. I've removed the insertValue properties and moved their args to the parent subcommand.

Also added missing options referenced in man git-reset.


QuiiBz (approved):
Nice fix! Feel free to merge after resolving the merge conflict :)

PR #592, opened 2021-09-18
feat(subl): add options

SeparateRecords:
This PR adds options to `subl`

### DIFF
--- a/src/subl.ts
+++ b/src/subl.ts
@@ -2,16 +2,27 @@ const generateProjects: Fig.Generator = {
   template: "filepaths",
   filterTemplateSuggestions: (suggestions) => {
     const out: Fig.Suggestion[] = [];
+
+    // This is a combined filter/map that also mutates suggestion objects.
     for (const suggestion of suggestions) {
+      // Folders should always be emitted regardless of their name.
       if (suggestion.type === "folder") {
         out.push(suggestion);
         continue;
       }
-      if (/^\.sublime-(project|workspace)$/.test(suggestion.name)) {
-        suggestion.priority = 75;
+
+      // With the filepaths template, if it wasn't a folder, it's a file.
+      // Files should only be emitted if they're a sublime project/workspace.
+      // These should appear before folders, so their priority is increased.
+      if (
+        suggestion.name === ".sublime-project" ||
+        suggestion.name === ".sublime-workspace"
+      ) {
+        suggestion.priority = 76;
         out.push(suggestion);
       }
     }
+
     return out;
   },
 };
--- a/src/subl.ts
+++ b/src/subl.ts
@@ -1,10 +1,101 @@
+const generateProjects: Fig.Generator = {
+  template: "filepaths",
+  filterTemplateSuggestions: (suggestions) => {
+    const out: Fig.Suggestion[] = [];
+    for (const suggestion of suggestions) {
+      if (suggestion.type === "folder") {
+        out.push(suggestion);
+        continue;
+      }
+      if (/^\.sublime-(project|workspace)$/.test(suggestion.name)) {
+        suggestion.priority = 75;
+        out.push(suggestion);
+      }
+    }
+    return out;
+  },
+};
+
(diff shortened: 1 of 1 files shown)


src/subl.ts
```diff
@@ -1,10 +1,83 @@
...
+        out.push(suggestion);
+        continue;
+      }
+      if (/^\.sublime-(project|workspace)$/.test(suggestion.name)) {
+        suggestion.priority = 75;
+        out.push(suggestion);
+      }
+    }
+    return out;
+  },
```
QuiiBz, line 16: I would prefer a cleaner solution, something like:
```suggestion
  filterTemplateSuggestions: (suggestions) => suggestions.map(suggestion => {
    if (suggestion.type === "folder") {
      return suggestion;
    }

    return {
      ...suggestion,
      priority: /^\.sublime-(project|workspace)$/.test(suggestion.name)
        ? 75
        : suggestion.priority
    }
  })
```

QuiiBz (changes requested):
Thanks for this PR!

src/subl.ts
```diff
@@ -1,10 +1,83 @@
...
+        name: "command",
+      },
+    },
+    {
+      name: ["-n", "--new-window"],
+      description: "Open a new window",
+    },
+    {
+      name: ["-a", "--add"],
+      description: "Add folders to the current window",
```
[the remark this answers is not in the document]
SeparateRecords: Not by itself, it's just a modifier. When it's present, ST will add the files/dirs to the current window

src/subl.ts
```diff
@@ -1,10 +1,83 @@
...
+      args: {
+        name: "project",
+        generators: generateProjects,
+      },
+    },
+    {
+      name: "--command",
+      description: "Run the given command",
+      args: {
+        name: "command",
```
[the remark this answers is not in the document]
SeparateRecords: It's not the kind of command Fig provides completions for, though - it's one of [these](https://docs.sublimetext.io/reference/commands.html#about-paths-in-command-arguments). Should it still be set?

src/subl.ts
SeparateRecords: I think the actual problem here was my lack of comments explaining the intention :)
This loop also removes the files that _don't_ match the regex, so it's a combined filter/map. The equivalent would be this, which is slower. Debatably clearer, but comments can fix that in the imperative version.

```typescript
const generateProjects: Fig.Generator = {
  template: "filepaths",
  filterTemplateSuggestions: (suggestions) =>
    suggestions
      .filter((suggestion) => {
        if (suggestion.type === "folder") {
          return true;
        }
        return /^\.sublime-(project|workspace)$/.test(suggestion.name);
      })
      .map((suggestion) => {
        if (suggestion.type === "folder") {
          return suggestion;
        }
        return { ...suggestion, priority: 75 };
      }),
};
```

Lmk if you'd rather go ahead with this version!

src/subl.ts
```diff
@@ -1,10 +1,83 @@
...
+      },
+    },
+    {
+      name: ["-n", "--new-window"],
+      description: "Open a new window",
+    },
+    {
+      name: ["-a", "--add"],
+      description: "Add folders to the current window",
+    },
```
SeparateRecords, line 51: @QuiiBz -n and -a aren't mutually exclusive, but when -n is present -a is silently ignored. Should these be marked as `exclusiveOn`?

src/subl.ts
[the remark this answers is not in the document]
QuiiBz: Thanks for the clarification! So no, no need to set `isCommand`, you're 100% right!

src/subl.ts
QuiiBz: That would be awesome if you could add some comments!

src/subl.ts
QuiiBz: Totally! You can add `-a` to `exclusiveOn` for the `"-n", "--new-window"` option

mschrage (approved):
🙌

PR #594, opened 2021-09-20
feat(whois): add options

SeparateRecords:
Removes non-existent options and adds all the other options

Incorrect options are suggested, and none of the actual options are

**Additional info:**
Because macOS, it's another BSD spec. Will need to be updated in the future.

### DIFF
--- a/src/whois.ts
+++ b/src/whois.ts
@@ -1,13 +1,95 @@
 const completionSpec: Fig.Spec = {
   name: "whois",
+  description: "Query a database for information about a domain registrant",
   args: {
     name: "name",
+    isVariadic: true,
   },
-  description: "Client for the whois directory service",
   options: [
-    { name: ["-H", "-p", "PORT"], description: "Connect to PORT" },
-    { name: "--verbose", description: "Be verbose" },
-    { name: "--help", description: "Display online help" },
+    {
+      name: "-a",
+      description:
+        "Use the American Registry for Internet Numbers (ARIN) database",
+    },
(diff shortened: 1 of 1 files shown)


src/whois.ts
```diff
@@ -1,13 +1,95 @@
 const completionSpec: Fig.Spec = {
   name: "whois",
+  description: "Query a database for information about a domain registrant",
   args: {
     name: "name",
+    isVariadic: true,
   },
-  description: "Client for the whois directory service",
   options: [
-    { name: ["-H", "-p", "PORT"], description: "Connect to PORT" },
```
QuiiBz, line 8: Looks like the `-p` option exists, along with `-h` for host:
Screenshot 2021-09-22 at 12 41 19

Could you add them back?

QuiiBz (changes requested):

src/whois.ts
SeparateRecords: Already added (L39 and L69) 😄 

Although, I didn't notice that was a pipe between `-c` and `-h` - they are indeed exclusive, so I'll fix that.

QuiiBz: I'm dumb, sorry 😅 Alright!

SeparateRecords:
Btw this is ready for review again 😅 I probably should have said so sooner haha

fedeci (approved):

QuiiBz (approved):
LGTM, a big improvement on this spec, thank you so much!

PR #595, opened 2021-09-20
feat(serve): add autocompletion spec

Pkingo:
feature: added autocompletion for serve

None.

Added autocompletion for serve in accordance with their help section

**Additional info:**
This is my first contribution to fig, and I'm by no means an expert in serve. So if you have any suggestions or something you would like to see different, please let me know.

### DIFF
--- a/dev/serve.ts
+++ b/dev/serve.ts
@@ -1,6 +1,3 @@
-// To learn more about Fig's autocomplete standard visit: https://fig.io/docs/concepts/cli-skeleton
-
-// The below is a dummy example for git. Make sure to change the file name!
 const completionSpec: Fig.Spec = {
   name: "serve",
   description: "Static file serving and directory listing",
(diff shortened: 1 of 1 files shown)


dev/serve.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      description:
+        "Specify a URI endpoint on which to listen - more than one may be specified to listen in multiple places",
+      args: {
+        name: "listen_uri",
+        suggestions: [
+          {
+            name: "port",
+            description: "E.g. serve --listen 1234",
+            type: "arg",
```
QuiiBz, line 28: You can safely remove `type: "arg"` from all the suggestions

QuiiBz (changes requested):
Thank you so much for this PR!

dev/serve.ts
Pkingo: Great. Thanks 💪

QuiiBz (approved):

PR #603, opened 2021-09-23
pdfunite spec

lsh:
Re submission of the PDFUnite spec.


QuiiBz (approved):
Thanks for opening a new PR with a clean history :) LGTM!

PR #608, opened 2021-09-26
feat: add htop completion spec

jhrcook:
New feature: add htop completion spec

There is no current behavior.

The new behavior is to show completion options for the `htop` command.

**Additional info:**
None.

### DIFF
--- a/src/htop.ts
+++ b/src/htop.ts
@@ -1,0 +1,83 @@
+const completionSpec: Fig.Spec = {
+  name: "htop",
+  description: "Improved top (interactive process viewer)",
+
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show help for htop",
+    },
+    {
+      name: ["--no-color", "-C"],
+      description: "Use a monochrome color scheme",
+    },
+    {
+      name: ["--delay", "-d"],
+      description: "Delay between updates, in tenths of sec",
+      args: {
+        name: "delay",
+        suggestions: ["10"],
+        isOptional: false,
+      },
+    },
+    {
+      name: ["--filter", "-F"],
+      description: "Filter commands",
+      args: {
+        name: "filter",
+        isOptional: false,
+      },
+    },
+    {
+      name: ["--highlight-changes", "-H"],
+      description: "Highlight new and old processes",
+      args: {
+        name: "delay",
+        isOptional: true,
+      },
+    },
+    {
+      name: ["--no-mouse", "-M"],
+      description: "Disable the mouse",
+    },
+    {
+      name: ["--pid", "-p"],
+      description: "Show only the given PIDs",
+      args: {
+        name: "PID",
+        isOptional: false,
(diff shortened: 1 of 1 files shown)


src/htop.ts
```diff
@@ -0,0 +1,83 @@
...
+        isOptional: false,
+      },
+    },
+    {
+      name: ["--highlight-changes", "-H"],
+      description: "Highlight new and old processes",
+      args: {
+        name: "delay",
+        isOptional: true,
+      },
```
fedeci, line 37: is this supposed to be here?

fedeci (approved):

src/htop.ts
jhrcook: Yes,  it is an optional delay for argument for how frequently to update the highlighting. Should I add a description to that argument to explain that?

fedeci: Yes please!

jhrcook: Done

src/htop.ts
```diff
@@ -0,0 +1,84 @@
+const completionSpec: Fig.Spec = {
+  name: "htop",
+  description: "Improved top (interactive process viewer)",
+
```
QuiiBz, line 4: Extra line not needed:
```suggestion
```

src/htop.ts
```diff
@@ -0,0 +1,84 @@
...
+    {
+      name: ["--no-color", "-C"],
+      description: "Use a monochrome color scheme",
+    },
+    {
+      name: ["--delay", "-d"],
+      description: "Delay between updates, in tenths of sec",
+      args: {
+        name: "delay",
+        suggestions: ["10"],
```
QuiiBz, line 19: Maybe we could add 1 and 60?

src/htop.ts
```diff
@@ -0,0 +1,84 @@
...
+        name: "filter",
+        isOptional: false,
+      },
+    },
+    {
+      name: ["--highlight-changes", "-H"],
+      description: "Highlight new and old processes",
+      args: {
+        name: "delay",
+        description: "Delay between updates of highlights, in tenths of sec",
```
QuiiBz, line 36: Could we add the same suggestions as above?

src/htop.ts
```diff
@@ -0,0 +1,84 @@
...
+    {
+      name: ["--no-mouse", "-M"],
+      description: "Disable the mouse",
+    },
+    {
+      name: ["--pid", "-p"],
+      description: "Show only the given PIDs",
+      args: {
+        name: "PID",
+        isOptional: false,
```
QuiiBz, line 49: You can remove `isOptional` when it's `false`. Same for other places underneath

QuiiBz (changes requested):
Thanks a lot for this PR! Left few comments before merging it :)

QuiiBz (approved):
LGTM, thank you so much for this contribution 🚀

Issue #545, opened 2021-08-31
[npm] Autocomplete for --legacy-peer-deps not available


akashshyamdev:
### Description:
> Please include a detailed description of the issue (and an image or screen recording, if applicable):
Here's the autocomplete that appears when `--legacy-` is entered:
![Screenshot 2021-08-31 at 14 22 33](https://user-images.githubusercontent.com/56759828/131473033-1e19fde4-46eb-4f4e-b4a2-87ebe8ff10ba.png)

As you can see, only `--legacy-bundling` is shown and not `--legacy-peer-deps` which is a common configuration option. Is there a reason for this? 

 
### Details:
|macOS|Fig|Shell|
|-|-|-|
|11.5.2|Version 1.0.47 (B257)|/bin/zsh|
<details><summary><code>fig diagnostic</code></summary>
<p>
<pre>Version 1.0.47 (B257) [ABC - India] 
UserShell: /bin/zsh
Bundle path: /Applications/Fig.app
Autocomplete: true
Settings.json: true
CLI installed: true
CLI tool path: /Users/akash/.fig/bin/fig
Accessibility: true
Number of specs: 185
SSH Integration: false
Tmux Integration: true
Keybindings path: /Users/akash/.fig/user/keybindings
iTerm Integration: true 
Hyper Integration: false
VSCode Integration: true
Docker Integration: false
Symlinked dotfiles: false
Only insert on tab: false
Installation Script: true
PseudoTerminal Path: /usr/local/share/npm/bin:/Users/akash/.rbenv/shims:/Users/akash/anaconda3/bin:/Users/akash/anaconda3/condabin:/Library/Frameworks/Python.framework/Versions/3.9/bin:/Library/Frameworks/Python.framework/Versions/3.8/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Library/Apple/usr/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/Users/akash/Library/Android/sdk/platform-tools:/usr/ucb:/usr/local/mysql/bin
SecureKeyboardInput: false
SecureKeyboardProcess: <none>
Current active process: /usr/local/Cellar/node/15.11.0/bin/node (27137) - ttys004
Current working directory: /Users/akash/Code/Projects/microfrontend
Current window identifier: 37693/code:26117% (com.microsoft.VSCode)</pre>
</p>
</details>

QuiiBz:
Merged in #610, thanks!

PR #610, opened 2021-09-27
[npm] Add --legacy-peer-deps for npm install

fwesss:
--legacy-peer-deps is a common option used with npm install but it is not in the current spec.

This option tells npm to skip installing peer dependencies. It is most commonly use when building a react app with react >= 17 and installing packages that don't explicitly have react >= 17 listed as a peer dependency.

This flag tells npm to behave as it did in versions v4 - v6 when installing packages.

https://stackoverflow.com/questions/66239691/what-does-npm-install-legacy-peer-deps-do-exactly-when-is-it-recommended-wh

Re: https://github.com/withfig/autocomplete/issues/545

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -230,6 +230,11 @@ const completionSpec: Fig.Spec = {
             "Causes npm to install the package such that versions of npm prior to 1.4, such as the one included with node 0.8, can install the package",
         },
         {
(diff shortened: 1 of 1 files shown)


QuiiBz (approved):
Good catch, LGTM 🎉

Issue #613, opened 2021-09-28
Missing script: "create-example"


Hongbo-Miao:
When I follow [Get Started](https://github.com/withfig/autocomplete/blob/e798be751ff90eb2d1cbc67fa85c5edc8d84e97e/README.md#-get-started), this step

```sh
npm run create-example
```

gives me error

> Missing script: "create-example"

When I check package.json file, it also does not have `create-example` in scripts section

https://github.com/withfig/autocomplete/blob/e798be751ff90eb2d1cbc67fa85c5edc8d84e97e/package.json#L9-L19

fedeci:
It has been replaced by the `create-spec` subcommand (`npm run create-spec [name of the spec]`). I will update the README!

PR #614, opened 2021-09-28
docs: update `README` and `package.json`

fedeci:
#613

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ git remote add upstream https://github.com/withfig/autocomplete.git
 npm install
 
 # Create an example spec (call it "abc")
-npm run create-example
+npm run create-spec abc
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -81,9 +81,6 @@ npm run dev
 
 ```bash
 
-# Create a new spec from a boilerplate template
-npm run create-boilerplate
-
```
QuiiBz, line 86: We should also remove it from the `fig` key in `package.json`, and update `create-example` to `create-spec`

QuiiBz (changes requested):

README.md
fedeci: Yes, i found that a couple hours ago but forgot to change.

QuiiBz (approved):
LGTM!

PR #615, opened 2021-09-29
feat: add `php please` spec for Statamic 3

martyf:
Spec addition: Statamic 3's `please` command

Currently `please` is not available.

Add's Statamic 3's `please` spec based on the command list json output

**Additional info:**
This is replicating the setup of the `artisan` spec for Laravel - perfect fit as Statamic is built on top of Laravel. Full credit to the original authors of the `artisan` implementation for this - this PR operates the same way, but runs the `please` command instead. This includes:
- only adding the spec if `please` is in the current directory, 
- pulls the available options from the `php please list --format=json` command

### DIFF
--- a/src/php/please.ts
+++ b/src/php/please.ts
@@ -1,0 +1,52 @@
+const completionSpec: Fig.Spec = {
+  name: "please",
+  description: "Statamic Please command",
+  generateSpec: async (tokens, executeShellCommand) => {
+    var out = await executeShellCommand("php please list --format=json");
+    const subcommands = [];
+
+    try {
+      const commandDefinition = JSON.parse(out);
+
+      commandDefinition.commands.map((command) => {
+        subcommands.push({
+          name: command.name,
+          description: command.description,
+
+          args: Object.keys(command.definition.arguments).map((argumentKey) => {
+            const argument = command.definition.arguments[argumentKey];
+
+            return {
+              name: argument.name,
+              description: argument.description,
+              isOptional: !argument.is_required,
+            };
+          }),
+          options: Object.keys(command.definition.options).map((optionKey) => {
+            const option = command.definition.options[optionKey];
+            const names = [option.name];
+
+            if (option.shortcut !== "") {
+              names.push(option.shortcut);
+            }
+
+            return {
+              name: names,
+              description: option.description,
+            };
+          }),
+        });
+      });
+    } catch (err) {
+      //
+    }
+
+    return {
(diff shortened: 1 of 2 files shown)


src/php/please.ts
```diff
@@ -0,0 +1,52 @@
+const completionSpec: Fig.Spec = {
+  name: "please",
+  description: "Statamic Please command",
+  generateSpec: async (tokens, executeShellCommand) => {
+    var out = await executeShellCommand("php please list --format=json");
```
QuiiBz, line 5: Could you replace this with a `const`?
```suggestion
    const out = await executeShellCommand("php please list --format=json");
```

src/php/please.ts
```diff
@@ -0,0 +1,52 @@
...
+
+            return {
+              name: names,
+              description: option.description,
+            };
+          }),
+        });
+      });
+    } catch (err) {
+      //
```
QuiiBz, line 41: We should add a `console.error(err);` in the catch, to be able to debug the spec in the future

QuiiBz (changes requested):
Thanks for this first PR! We have only two small issues to resolve before merging this 🚀

martyf:
Thanks @QuiiBz - I've just updated the PR with these two tweaks.

src/php/please.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    try {
+      const commandDefinition = JSON.parse(out);
+
+      commandDefinition.commands.map((command) => {
+        subcommands.push({
+          name: command.name,
+          description: command.description,
+
+          args: Object.keys(command.definition.arguments).map((argumentKey) => {
```
fedeci, line 16: nit: we can directly use `Object.values` to simplify this logic

src/php/please.ts
```diff
@@ -0,0 +1,52 @@
...
+            const argument = command.definition.arguments[argumentKey];
+
+            return {
+              name: argument.name,
+              description: argument.description,
+              isOptional: !argument.is_required,
+            };
+          }),
+          options: Object.keys(command.definition.options).map((optionKey) => {
```
fedeci, line 25: Same as above

martyf:
Tweaked to use `Object.values` instead of `Object.keys`

fedeci (approved):

QuiiBz (approved):
This looks amazing, thank you so much for this contribution 🎉

PR #619, opened 2021-09-29
feat(`npm`): suggest global deps if global flag is specified

### LINKED ISSUE #616: [npm]: correctly suggest packages when using `-g` flag
### Sanity checks

- [x] I would like to work on this.

npm

### Issue Details

When using for example `npm -g rm <package>` we do not provide the correct suggestion.

fedeci:
Closes #616

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -1,3 +1,15 @@
+function uninstallSubcommand(named: string | string[]): Fig.Subcommand {
+  return {
+    name: named,
+    description: "Uninstall a package",
+    args: {
+      name: "package",
+      generators: dependenciesGenerator,
+      isVariadic: true,
+    },
+    options: npmInstallOptions,
+  };
+}
 // GENERATORS
 export const npmSearchGenerator: Fig.Generator = {
   script: function (context) {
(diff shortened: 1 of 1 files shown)


src/npm.ts
```diff
@@ -1,3 +1,15 @@
+function uninstallSubcommand(named: string | string[]): Fig.Subcommand {
```
QuiiBz, line 1: Any reason why using `named` instead of `name`?

src/npm.ts
```diff
@@ -53,14 +65,13 @@ const workspaceGenerator: Fig.Generator = {
...
-    "until [[ -f package.json ]] || [[ $PWD = '/' ]]; do cd ..; done; cat package.json",
-  postProcess: function (out, context = []) {
-    if (out.trim() === "") {
-      return [];
-    }
-
-    try {
+  trigger: (newToken) => newToken === "-g" || newToken === "--global",
+  custom: async function (tokens, executeShellCommand) {
+    console.error(tokens);
```
QuiiBz, line 70: We can remove this `console.error`

QuiiBz (changes requested):
Amazing feature! Would it be possible to port it to `yarn` and `pnpm`?

fedeci:
Do yarn and pnpm allow global commands?

src/npm.ts
fedeci, line 70:
```suggestion
```

QuiiBz:
> Do yarn and pnpm allow global commands?

They install binaries the same way as `npm` do, that would be helpful for `yarn global remove ...` & `pnpm remove --global ...`

fedeci:
pnpm seems to be already using the npm generator

QuiiBz:
> pnpm seems to be already using the npm generator. Yarn already had the feature.

Nice to hear! But it seems like `yarn` isn't working properly. We might handle this in another PR.

Screenshot 2021-09-30 at 21 17 24

fedeci:
It is working for me🤔 Try to install a package e.g. express globally and see if works

QuiiBz:
Found why it's not working: the generator uses `cat`, which I aliased to [`bat`](https://github.com/sharkdp/bat), and the output isn't the same. The simplest way to fix this is to append a `\` to the command, so the shell will execute the right binary, and not the aliased one. I'm going to open a PR since this issue is happening in multiple specs.

QuiiBz (approved):
LGTM!

PR #624, opened 2021-09-30
feat: Add `wc` spec

QuiiBz:
Feature

No spec for `wc`

Add a spec for `wc`

**Additional info:**

### DIFF
--- a/src/wc.ts
+++ b/src/wc.ts
@@ -1,0 +1,31 @@
+const completionSpec: Fig.Spec = {
+  name: "wc",
+  description: "World, line, character, and byte count",
+  options: [
+    {
+      name: "-c",
+      description: "Output the number of bytes to the standart input",
+    },
+    {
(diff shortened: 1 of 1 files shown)


src/wc.ts
```diff
@@ -0,0 +1,31 @@
+const completionSpec: Fig.Spec = {
+  name: "wc",
+  description: "World, line, character, and byte count",
+  options: [
+    {
+      name: "-c",
+      description: "Output the number of bytes to the standart input",
```
fedeci, line 7: typo: there are other occurrences of this error
```suggestion
      description: "Output the number of bytes to the standard input",
```

fedeci (approved):
LGTM apart from a typo

src/wc.ts
QuiiBz: Good catch!

PR #625, opened 2021-10-01
added all amplify commands to spec

Dayton-IX:
Full spec for amplify CLI autocomplete support

Only amplify env is supported

All amplify commands are supported

**Additional info:**

### DIFF
--- a/src/amplify.ts
+++ b/src/amplify.ts
@@ -8,11 +8,205 @@ const envNameGenerator: Fig.Generator = {
   },
 };
 
+const amplifyCategories = [
+  {
+    name: "notifications",
+  },
+  {
+    name: "analytics",
+  },
+  {
+    name: "api",
+  },
+  {
+    name: "auth",
+  },
+  {
+    name: "function",
+  },
+  {
+    name: "hosting",
+  },
+  {
+    name: "interactions",
+  },
+  {
+    name: "predictions",
+  },
+  {
+    name: "storage",
+  },
+  {
+    name: "xr",
+  },
+];
+
+const categoryCommands = [
+  {
+    name: "add",
+    description:
+      "Adds a resource for an Amplify category in your local backend",
+  },
+  {
+    name: "update",
+    description:
+      "Update resource for an Amplify category in your local backend",
+  },
+  {
+    name: "remove",
+    description:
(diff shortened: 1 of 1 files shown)


src/amplify.ts
```diff
@@ -8,11 +8,205 @@ const envNameGenerator: Fig.Generator = {
   },
 };
 
+const amplifyCategories = [
```
QuiiBz, line 11: Could you add a type of `Fig.Subcommand[]` here? And a description for each subcommand would be nice to have!

src/amplify.ts
```diff
@@ -8,11 +8,205 @@ const envNameGenerator: Fig.Generator = {
...
+  },
+  {
+    name: "storage",
+  },
+  {
+    name: "xr",
+  },
+];
+
+const categoryCommands = [
```
QuiiBz, line 44: Same as above for the type :)

src/amplify.ts
```diff
@@ -8,11 +8,205 @@ const envNameGenerator: Fig.Generator = {
...
+    description:
+      "Removes a resource for an Amplify category in your local backend",
+  },
+  {
+    name: "push",
+    description:
+      "Provisions cloud resources with the latest local developments",
+    subcommands: [
+      {
+        name: "--y",
```
QuiiBz, line 66: Is the name really `--y`, and not `-y` instead? That seems a bit strange to me

src/amplify.ts
```diff
@@ -8,11 +8,205 @@ const envNameGenerator: Fig.Generator = {
...
+      description:
+        "Configures the attributes of your project for amplify-cli, such as switching front-end framework and adding/removing cloud-provider plugins",
+    },
+    {
+      name: "push",
+      description:
+        "Provisions cloud resources with the latest local developments",
+      subcommands: [
+        {
+          name: "--y",
```
QuiiBz, line 94: Same as above

QuiiBz (changes requested):
Thanks a lot for this contribution! I left some comments to resolve before getting this PR merged in

src/amplify.ts
Dayton-IX: yeah it really is --y!

QuiiBz (approved):
LGTM, thank you so much for this contribution!

PR #635, opened 2021-10-02
feat: spec for R and Rscript commands for the R programming language

jhrcook:

Feature: spec for R and Rscript commands for the R programming language

The `R` and `Rscript` commands are currently not supported.

Add support for the `R` and `Rscript` commands for the R programming language.

**Additional info:**

I am not experienced with Javascript nor Typescript, so if there are ways of improving the generators I made, please let me know or make the changes yourself. 
Also, if you do accept this PR, can you please add the label `hacktoberfest-accepted` so I can get credit for [Hacktoberfest](https://hacktoberfest.digitalocean.com)?


src/R.ts
```diff
@@ -0,0 +1,1475 @@
...
+      },
+    },
+  ]),
+  // Only uncomment if R takes an argument
+  // args: {}
+};
+
+export default completionSpec;
+
+// TODO: at end, add icons where appropriate
```
QuiiBz, line 1475: Is this todo completed?

QuiiBz (changes requested):
Thanks for this contribution, you did a great job for this spec! Love the generators, the spec is really well made. I left some comments to further improve the quality of it.

src/Rscript.ts
```diff
@@ -0,0 +1,99 @@
...
+    },
+    {
+      name: "--default-packages",
+      description: "Comma separated list of default packages",
+      requiresEquals: true,
+      icon: "📦",
+      args: {
+        name: "packages",
+        description: "A comma-separated set of package names, or 'NULL'",
+        isOptional: false,
```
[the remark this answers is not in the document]
jhrcook: Sorry, I keep including this, I try to keep it straight in my head, but get turned around a bit between the default behavior in `options` vs. `args`. Perhaps it could help to explicitly state the default behavior in the documentation? E.g. change "True if an argument is optional.” to "True if an argument is optional (default is false).”?

src/Rscript.ts
```diff
@@ -0,0 +1,99 @@
...
+                ...file,
+                priority: isRFile && 76,
+              };
+            });
+        },
+      },
+    },
+    {
+      name: "args",
+      description: "Arguments to pass to the script",
```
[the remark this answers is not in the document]
jhrcook: Yes

src/R.ts
```diff
@@ -0,0 +1,1475 @@
...
+    description: "Remove files created during a previous run",
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "R",
+  description: "Start R or invoke an R tool using CMD",
+  subcommands: [
+    {
+      name: "RHOME",
```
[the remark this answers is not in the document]
jhrcook: Yes, some are and some are not.

src/R.ts
```diff
@@ -0,0 +1,1475 @@
...
+                name: ["--debug", "-d"],
+                description: "Turn on debugging messages",
+              },
+              {
+                name: ["--library", "-l"],
+                description: "Install packages to library tree LIB",
+                requiresEquals: true,
+                args: {
+                  name: "LIB",
+                  description: "Path to the R library tree to install to",
```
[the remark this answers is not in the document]
jhrcook: Thank you for suggesting this and I think it would be even better to suggest the actual libraries that R knows about. I can get this information using the `.libPaths()` command in R, so I made a generator specifically for finding R libraries. I’ll add it to other options/arguments that call for a library path.

src/R.ts
```diff
@@ -0,0 +1,1475 @@
...
+                name: "--no-docs",
+                description: "Do not install HTML, LaTeX or examples help",
+              },
+              {
+                name: "--html",
+                description: "Build HTML help",
+              },
+              {
+                name: "--no-html",
+                description: "Do not build HTML help",
```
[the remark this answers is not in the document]
jhrcook: Thank you for letting me know about this feature. I will look through the rest of the options to add this attribute where appropriate.

src/R.ts
```diff
@@ -0,0 +1,1475 @@
...
+          name: "LINK",
+          description: "Perform the specified linkcmd",
+          options: helpAndVersionOptions,
+          args: {
+            name: "linkcmd",
+            description: "Command for linkcmd",
+          },
+        },
+        {
+          name: "Rprof",
```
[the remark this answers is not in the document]
jhrcook: I double checked and both `Rprof` and `rprof` work, but the former (capitalized) is listed in the help doc. Which sould I set or can I set both (though I think this would be more confusing than helpful)?

src/R.ts
jhrcook: Thank you for the reminder!

jhrcook:
> Thanks for this contribution, you did a great job for this spec! Love the generators, the spec is really well made. I left some comments to further improve the quality of it.

Thank you for the kind words and useful suggestions! I will take care of them now.

src/R.ts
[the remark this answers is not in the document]
QuiiBz: I would prefer having it in lowercase

src/rscript.ts
```diff
@@ -0,0 +1,105 @@
+const completionSpec: Fig.Spec = {
+  name: "Rscript",
```
QuiiBz, line 2: Same as above, it would be better in lowercase!

src/r.ts
```diff
@@ -0,0 +1,1536 @@
...
+  },
+  {
+    name: "--preclean",
+    description: "Remove files created during a previous run",
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "R",
+  description: "Start R or invoke an R tool using CMD",
```
QuiiBz, line 125: Could you also set this in lowercase?

src/R.ts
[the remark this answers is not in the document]
jhrcook: Done

src/r.ts
```diff
@@ -0,0 +1,1536 @@
...
+  icon: "https://www.r-project.org/favicon-32x32.png",
+  subcommands: [
+    {
+      name: "RHOME",
```
QuiiBz, line 129: Let’s put all subcommands/options in lowercase

QuiiBz (changes requested):

src/rscript.ts
jhrcook: While invoking `rscript` does work, it is only because when R is installed, there is a symlink from the `Rscript` command:

```
ll /usr/local/bin/Rscript
lrwxr-xr-x  1 root  admin    53B Aug 15 10:35 /usr/local/bin/Rscript -> /Library/Frameworks/R.framework/Resources/bin/Rscript
```

```
ll /usr/local/bin/rscript
lrwxr-xr-x  1 root  admin    53B Aug 15 10:35 /usr/local/bin/rscript -> /Library/Frameworks/R.framework/Resources/bin/Rscript
```

This is at least true for MacOS, but I can’t guarantee that is true for Linux or Windows. As far as I’m aware, Fig doesn’t care if the spec is named `rscript` or `Rscript`, but this may add some confusion where it isn’t needed.

QuiiBz: > This is at least true for MacOS

We’re currently only focusing on macOS so that’s fine. I find it a bit strange that a CLI name can be case-sensitive 🤔

src/r.ts
jhrcook: I’m not sure if this is a good idea because the command `r` in zsh is another command and is the same as `fc -e -.` where "The `fc` command controls the interactive history mechanism."

source: https://zsh.sourceforge.io/Doc/Release/Shell-Builtin-Commands.html

From what I can tell, the `r` command in zsh re-runs the previous command. But, simillar as with `Rscript` vs. `rscript`, if Fig doesn’t care, then it should be fine.

src/r.ts
jhrcook: This won't work in all cases. Both `RHOME` and `CMD` must be uppercase (they aren’t recognized when lowercase), but every subcommand under `R CMD …` works fine as lowercase. I can change these all to be lowercase, but I just want to mention again that they are currently recorded in the Fig spec as they are in the documentation. Can you confirm that you want them all lowercase still?

src/rscript.ts
jhrcook: > I find it a bit strange that a CLI name can be case-sensitive 🤔
 
Me too, it actually took me awhile to realize because I kept using `r` after using `R` so they seemed to be the same.

```bash
❯ which r
r: shell built-in command
❯ which R
/usr/local/bin/R
```

src/r.ts
QuiiBz: We should lowercase them when they work in lowercase, but if `RHOME` and `CMD` only work in uppercase, then we should keep them in uppercase. That CLI is definitely a bit strange 😅

jhrcook: Welcome to the world of programming in R – I didn’t actually know that all the different forms would work, so at least we were both confused/frustrated together. Ill get on these changes in a minute.

src/rscript.ts
jhrcook: Sorry, I meant to put this comment for the other conversation. Any conclusion though on `Rscript` vs. `rscript`?

src/r.ts
jhrcook: Any conclusion on this? Personally, I think we need to stick with `R` because on my compute `r` is a different command.

PR #640, opened 2021-10-05
feat: add sequelize completion spec

pondpiu:
Feature

No spec for sequelize yet.

Add spec for sequelize

**Additional info:**

### DIFF
--- a/src/sequelize.ts
+++ b/src/sequelize.ts
@@ -1,0 +1,85 @@
+const subCommandOptions: Fig.Option[] = [
+  { name: "--env", description: "The environment to run the command in" },
+  { name: "--config", description: "The path to the config file" },
+  {
+    name: "--options-path",
+    description: "The path to a JSON file with additional options",
+  },
+  {
+    name: "--migrations-path",
+    description: "The path to the migrations folder",
+  },
+  { name: "--seeders-path", description: "The path to the seeders folder" },
+  { name: "--models-path", description: "The path to the models folder" },
+  {
+    name: "--url",
+    description: "The database connection string to use",
+  },
+  {
+    name: "--debug",
+    description: "When available show various debug information",
+  },
+];
+
+const subCommands: Fig.Subcommand[] = [
+  { name: "db:migrate", description: "Run pending migrations" },
+  {
+    name: "db:migrate:schema:timestamps:add",
+    description: "Update migration table to have timestamps",
+  },
+  {
+    name: "db:migrate:status",
+    description: "List the status of all migrations",
+  },
+  { name: "db:migrate:undo", description: "Reverts a migration" },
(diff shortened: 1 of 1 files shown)


src/sequelize.ts
```diff
@@ -0,0 +1,254 @@
+const commonOptions: Fig.Option[] = [
+  {
+    name: "--env",
+    description: "The environment to run the command in",
+    args: { name: "environment" },
+  },
+  {
+    name: "--config",
+    description: "The path to the config file",
+    args: { name: "file path" },
```
QuiiBz, line 10: Looks like this argument can have a [`template`](https://fig.io/docs/reference/arg#template) of `filepaths`

src/sequelize.ts
```diff
@@ -0,0 +1,254 @@
...
+  },
+  {
+    name: "--options-path",
+    description: "The path to a JSON file with additional options",
+    args: { name: "file path" },
```
QuiiBz, line 15: Same as above

src/sequelize.ts
```diff
@@ -0,0 +1,254 @@
...
+  },
+  {
+    name: "--migrations-path",
+    description: "The path to the migrations folder",
+    args: { name: "folder path" },
```
QuiiBz, line 20: Same as above but for a template of `folders`. Repeating a few times underneath, could you fix them all?

src/sequelize.ts
```diff
@@ -0,0 +1,254 @@
...
+    name: "--template",
+    description: "Pass template option to dialect, PostgreSQL only",
+    args: { name: "template option" },
+  },
+];
+
+const initOptions: Fig.Option[] = [
+  {
+    name: "--force",
+    description: "Will drop the existing config folder and re-create it",
```
QuiiBz, line 82: Would love to see [`isDangerous`](https://fig.io/docs/reference/option#isdangerous) here!

QuiiBz (changes requested):
Thanks for this contribution!

pondpiu:
@QuiiBz changed as suggested

src/sequelize.ts
```diff
@@ -0,0 +1,255 @@
+const commonOptions: Fig.Option[] = [
+  {
+    name: "--env",
+    description: "The environment to run the command in",
+    args: { name: "environment" },
+  },
+  {
+    name: "--config",
+    description: "The path to the config file",
+    args: { template: "filepaths" },
```
QuiiBz, line 10: Could you add back the `name` property to arguments that have a `template`? Sorry if I misspoke

QuiiBz (changes requested):

pondpiu:
@QuiiBz `name` property has been added

PR #699, opened 2021-10-30
feat: add Redwood cli completion spec

simoncrypta:
Feature: add autocompletion for `Redwood`.
Feature: add `create-redwood-app` for yarn create.

No autocompletion for `Redwood`.

Add autocompletion for `Redwood`.

**Additional info:**
Reference : [RedwoodJS CLI Doc ](https://redwoodjs.com/docs/cli-commands)

### DIFF
--- a/src/redwood.ts
+++ b/src/redwood.ts
@@ -36,7 +36,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "options",
             description: "Options to pass to Webpack Dev Server",
-            isOptional: false,
           },
         },
       ],
@@ -106,7 +105,7 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "side",
             description: "Select side to build",
-            isOptional: false,
+
             suggestions: [
               {
                 name: "api",
@@ -164,7 +163,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the cell component",
-            isOptional: false,
           },
         },
         {
@@ -173,7 +171,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the component",
-            isOptional: false,
           },
         },
         {
@@ -182,7 +179,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the Function",
-            isOptional: false,
           },
         },
         {
@@ -191,7 +187,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the layout component",
-            isOptional: false,
           },
         },
         {
@@ -201,7 +196,6 @@ const completionSpec: Fig.Spec = {
             {
               name: "name",
               description: "Name of the page",
-              isOptional: false,
             },
             {
               name: "path",
@@ -217,7 +211,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the scaffold",
-            isOptional: false,
           },
         },
         {
@@ -227,7 +220,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the sdl",
-            isOptional: false,
           },
         },
         {
@@ -236,7 +228,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the service",
-            isOptional: false,
           },
         },
       ],
@@ -252,7 +243,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the cell component",
-            isOptional: false,
           },
           options: [
             {
@@ -285,7 +275,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the component",
-            isOptional: false,
           },
           options: [
             {
@@ -313,7 +302,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the component",
-            isOptional: false,
           },
         },
         {
@@ -322,7 +310,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the directive",
-            isOptional: false,
           },
           options: [
             {
@@ -347,7 +334,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the Function",
-            isOptional: false,
           },
           options: [
             {
@@ -367,7 +353,6 @@ const completionSpec: Fig.Spec = {
           args: {
             name: "name",
             description: "Name of the layout component",
-            isOptional: false,
           },
           options: [
             {
@@ -401,7 +386,6 @@ const completionSpec: Fig.Spec = {
             {
               name: "name",
               description: "Name of the page",
-              isOptional: false,
             },
             {
               name: "path",
@@ -437,7 +421,6 @@ const completionSpec: Fig.Spec = {
             name: "model",
             description:
               "Model to scaffold. You can also use <path/model> to nest files by type at the given path directory (or directories). For example, redwood g scaffold admin/post",
-            isOptional: false,
           },
           options: [
             {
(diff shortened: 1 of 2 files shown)


src/redwood.ts
```diff
@@ -0,0 +1,914 @@
+import prismaSpec from "./prisma";
+
+const prismaCommands = (prismaSpec as Fig.Subcommand).subcommands;
```
QuiiBz, line 3: Is this type casting mandatory?

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+      },
+      options: [
+        {
+          name: ["--forward", "--fwd"],
+          description:
+            "String of one or more Webpack Dev Server config options",
+          args: {
+            name: "options",
+            description: "Options to pass to Webpack Dev Server",
+            isOptional: false,
```
QuiiBz, line 39: You can remove every occurrences of `isOptional ` when it's `false` - it's the default value (see the [docs](https://fig.io/docs/reference/arg#isoptional))

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+          description: "Destroy a layout component",
+          args: {
+            name: "name",
+            description: "Name of the layout component",
+            isOptional: false,
+          },
+        },
+        {
+          name: "page",
+          description: "Destroy a  page and route component",
```
QuiiBz, line 199:
```suggestion
          description: "Destroy a page and route component",
```

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+          description: "Generate a cell component",
+          args: {
+            name: "name",
+            description: "Name of the cell component",
+            isOptional: false,
+          },
+          options: [
+            {
+              name: ["--force", "-f"],
+              description: "Overwrite existing filess",
```
QuiiBz, line 260: I think we should add [`isDangerous`](https://fig.io/docs/reference/option#isdangerous) here

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+          description: "Generate a component component",
+          args: {
+            name: "name",
+            description: "Name of the component",
+            isOptional: false,
+          },
+          options: [
+            {
+              name: ["--force", "-f"],
+              description: "Overwrite existing filess",
```
QuiiBz, line 293: Same as above, it seems like is repeating a few times

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+      description: "Upgrade all @redwoodjs packages via an interactive CLI",
+      options: [
+        {
+          name: ["--dry-run", "-d"],
+          description: "Check for outdated packages without upgrading",
+        },
+        {
+          name: ["--tag", "-t"],
+          description:
+            'Choices are "canary", "rc", or a specific version (e.g. "0.19.3"). WARNING: Unstable releases in the case of "canary" and "rc", which will force upgrade packages to the most recent release of the specified tag',
```
QuiiBz, line 599: Could we add [`suggestions`](https://fig.io/docs/reference/arg#suggestions)? And I think that this option takes an argument

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+        },
+        {
+          name: "--socket",
+          description:
+            "The socket the server should run. This takes precedence over port",
+        },
+        {
+          name: "--apiRootPath",
+          description: "The root path where your api functions are served",
+        },
```
QuiiBz, line 639: I think these three options should have arguments, and the last one can have a [`template`](https://fig.io/docs/reference/arg#template) of `folders`

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+          "Which side(s) to test, and/or a regular expression to match against your test files to filter by",
+        isOptional: true,
+        suggestions: [
+          {
+            name: "api",
+          },
+          {
+            name: "web",
+          },
+        ],
```
QuiiBz, line 684: In those cases where each suggestion only have a name, we can simply use an array of string:
```suggestion
        suggestions: [
           "api",
          "web",
        ],
```

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+          name: "--open",
+          description: "Open Storybook in your browser on start",
+        },
+        {
+          name: "--build",
+          description: "Build Storybook",
+        },
+        {
+          name: "--port",
+          description: "Which port to run Storybook on (defaults to 7910)",
```
QuiiBz, line 736: I think this option also takes an argument

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+              },
+              {
+                name: "goTrue",
+              },
+              {
+                name: "magicLink",
+              },
+              {
+                name: "netlify",
+              },
```
QuiiBz, line 782: Ditto

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+              {
+                name: "render",
+              },
+              {
+                name: "vercel",
+              },
```
QuiiBz, line 847: Ditto

src/redwood.ts
```diff
@@ -0,0 +1,914 @@
...
+                suggestions: [
+                  {
+                    name: "none",
+                  },
+                  {
+                    name: "postgresql",
+                  },
+                  {
+                    name: "sqlite",
+                  },
```
QuiiBz, line 867: Ditto (the last one :))

src/yarn.ts
```diff
@@ -334,11 +336,19 @@ const completionSpec: Fig.Spec = {
...
+              name: ["redwood", "rw"],
+              loadSpec: "redwood",
+              icon: "fig://icon?type=package",
+            }
+          : {
+              name,
+              loadSpec: name,
+              icon: "fig://icon?type=package",
+            }
+      );
```
QuiiBz, line 351: Not sure if this is the best approach, I would prefer something simplier:
```suggestion
      .map((name) => ({
        name: name === "@redwoodjs/core" ? ["redwood", "rw"] : name,
        loadSpec: name,
        icon: "fig://icon?type=package",
      }));
```

QuiiBz (changes requested):
Thanks a lot for this contribution! I left some comments to resolve before merging this

src/yarn.ts
simoncrypta: Agree that is way cleaner, but since it also needs for `loadSpec`, I don't know if is better to repeat the ternary operator or keep it like that ? I let you decide !

src/redwood.ts
simoncrypta: I don't know, I take the blitz spec as reference for that, since we have the same beaver for Prisma.
https://github.com/withfig/autocomplete/blob/e448d42855f362b65ea9f38adf7964d6dd3e7c01/src/blitz.ts#L3

simoncrypta:
Thank you, @QuiiBz, for this complete review ! I left comments for 2 conversations, let me know what you think. Also, I don't really understand what you mean when you say "Ditto", sorry I'm not a native English speaker.

simoncrypta:
Oh I understand now! I didn't see some of the reviews. I will fix this

src/redwood.ts
QuiiBz: Could you try to remove the type casting and see if typescript doesn't complain? IMO that would be a lot cleaner is that works!

src/yarn.ts
QuiiBz: I think using a ternary operation is both smaller and easier to read, so we can go with this!

QuiiBz:
> I left comments for 2 conversations, let me know what you think.

Just replied to them, will merge this amazing PR when those two small changes are done!

src/redwood.ts
simoncrypta: Typescript need to know that `prismaSpec`is a `Fig.Subcommand`, but I don't think we should have a `const` for that.

simoncrypta:
> Just replied to them, will merge this amazing PR when those two small changes are done!

Done 👍 Thank you again for this review, I appreciate the effort you put on it 😊

QuiiBz (commented):
LGTM, thank you so much for this contribution 🎉  And thanks for those kind words!

QuiiBz (approved):

Issue #694, opened 2021-10-25
[vercel] Spec is missing some commands


piotrski:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

vercel

### Which statement makes the most sense?

This completion spec isn't complete

### Issue Details

`vercel link` and `vercel alias` commands seam to be missing from the spec

https://vercel.com/docs/cli#commands/link
https://vercel.com/docs/cli#commands/alias

PR #701, opened 2021-10-31
feat(vercel): Added completion for vercel's alias, link, billing

alexnaiman:
Added completion for for vercel's alias, link, billing
Missing completion specs for alias, link, billing. See #694 
Completion for vercel's alias, link, billing
**Additional info:**
none

### DIFF
--- a/src/vercel.ts
+++ b/src/vercel.ts
@@ -653,6 +653,73 @@ const completionSpec: Fig.Spec = {
       name: "whoami",
       description: "Shows the username of the currently logged in user",
     },
+    {
+      name: "alias",
+      description:
+        "Apply custom domains based on git branches, or other heuristics",
+      subcommands: [
+        {
+          name: "set",
+          description: "Assign a custom domain to a deployment",
+          args: [
+            {
+              name: "deployment url",
+            },
+            {
+              name: "custom domain",
+            },
+          ],
+        },
+        {
+          name: "rm",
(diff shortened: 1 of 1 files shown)


src/vercel.ts
```diff
@@ -653,6 +653,73 @@ const completionSpec: Fig.Spec = {
...
+      name: "alias",
+      description:
+        "Apply custom domains based on git branches, or other heuristics",
+      subcommands: [
+        {
+          name: "set",
+          description: "Assign a custom domain to a deployment",
+          args: [
+            {
+              name: "deployment url",
```
QuiiBz, line 666: The name of the argument should be one word if possible, and then you can add a `description` to explain a bit more about it. Can you change this for each argument underneath?

As example:
```suggestion
              name: "url",
              description: "The URL of the deployment to assign to a domain",
```

QuiiBz (changes requested):
Thanks a lot for this PR! There's only one issue that needs to be fixed at few places

src/vercel.ts
mschrage: I wonder if we can set up a linter for this? cc @fedeci

QuiiBz: That would be a good idea!

QuiiBz (approved):
LGTM, thanks a lot for this contribution!

PR #705, opened 2021-11-03
feat: dog spec

gusfune:

This adds specs for `dog` as requested on issue #702

### DIFF
--- a/src/dog.ts
+++ b/src/dog.ts
@@ -1,5 +1,6 @@
 const args: Fig.SingleOrArray<Fig.Arg> = {
   name: "arguments",
+  description: "Human-readable host names, nameservers, types, or classes",
 };
 
 const options: Fig.Option[] = [
@@ -80,7 +81,7 @@ const options: Fig.Option[] = [
 const completionSpec: Fig.Spec = {
   name: "dog",
   description: "A command-line DNS client",
-  args: {},
+  args,
   options,
   subcommands: [
     {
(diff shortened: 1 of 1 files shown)


src/dog.ts
```diff
@@ -0,0 +1,144 @@
+const args: Fig.SingleOrArray<Fig.Arg> = {
+  name: "arguments",
+};
+
+const options: Fig.Option[] = [
+  {
+    name: ["-q", "--query"],
+    insertValue: "{HOST}",
```
QuiiBz, line 8: Not sure what you wanted to achieve with `insertValue` here?

src/dog.ts
```diff
@@ -0,0 +1,144 @@
...
+  {
+    name: "--time",
+    description: "Print how long the response took to arrive",
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "dog",
+  description: "A command-line DNS client",
+  args: {},
```
QuiiBz, line 84: You can safely remove this if `dog` doesn't take any argument at the root-level

QuiiBz (changes requested):

src/dog.ts
gusfune: Dog can take arguments or no arguments at root-level. That's why I left this option. Shall I remove it?

QuiiBz: In this case, you should add information to each possible argument, such as the name, description, if it's variadic or not etc

QuiiBz: Still waiting for this, and then we'll be ready to merge!

gusfune: Sorry for the delay.
Working this out in a commit soon.

QuiiBz (approved):
Thanks for fixing the issues! LGTM 🚀

PR #707, opened 2021-11-04
Feature - expand DBT support. 

lilyroberts:
Adding suggestions for CLI options with enumerated arg values. Adding build and parse commands.  Adding new flags and arguments to various commands. 

Feature expansion.

Limited DBT CLI support.

Expanded support for DBT CLI.

### DIFF
--- a/src/dbt.ts
+++ b/src/dbt.ts
@@ -16,8 +16,109 @@ const completionSpec: Fig.Spec = {
           description: "Treat incremental models as table models",
         },
         {
-          name: ["-X", "--fail-fast"],
+          name: ["-x", "--fail-fast"],
           description: "Exit immediately if a single model fails to build",
+        },
+        {
+          name: "--use-colors",
+          description: "Default value -- colorize run logs",
+        },
+        {
+          name: "--no-use-colors",
+          description: "Disable log colorizing",
+        },
+        {
+          name: "--profiles-dir",
+          description: "",
+          args: {
+            name: "profiles directory",
+            description: "Directory containing profiles.yml",
+            template: "folders",
+            suggestions: ["profiles"],
+          },
+        },
+        {
+          name: "--profile",
+          description: "",
+          args: {
+            name: "Profile name",
+            description: "Name of profile to use when running dbt project",
+            suggestions: ["default"],
+          },
+        },
+        {
+          name: "--target",
+          description: "",
+          args: {
+            name: "Target profile",
+            description: "Target profile configuration (dev/prod)",
+            suggestions: ["dev", "prod"],
+          },
+        },
+        {
+          name: "--vars",
+          description: "Variable values to override dbt_project.yml",
+          args: {
+            name: "Variable values",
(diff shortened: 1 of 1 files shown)


src/dbt.ts
```diff
@@ -3,9 +3,111 @@ const completionSpec: Fig.Spec = {
...
+          name: "--full-refresh",
+          description: "Treat incremental models as table models",
+        },
+        {
+          name: ["-x", "--fail-fast"],
+          description: "Exit immediately if a single model fails to build",
+        },
+        {
+          name: "--profiles-dir",
+          description: "",
```
QuiiBz, line 19: Could you add a description for this option?

src/dbt.ts
```diff
@@ -3,9 +3,111 @@ const completionSpec: Fig.Spec = {
...
+          args: {
+            name: "profiles directory",
```
QuiiBz, line 21: Argument's `name` should ideally have a single word, in this case, `directory`. Could you update this for every argument underneath?

src/dbt.ts
```diff
@@ -3,9 +3,111 @@ const completionSpec: Fig.Spec = {
...
+            description: "Directory containing profiles.yml",
+            template: "folders",
+            suggestions: ["profiles"],
+          },
+        },
+        {
+          name: "--profile",
+          description: "",
```
QuiiBz, line 29: Missing description, looks like it happens a few times underneath

src/dbt.ts
```diff
@@ -3,9 +3,111 @@ const completionSpec: Fig.Spec = {
...
+            isVariadic: true,
+            suggestions: [
+              "path:",
+              "tag:",
+              "config:",
+              "test_type:",
+              "test_name:",
+            ],
+          },
+        },
```
QuiiBz, line 89: We can use [`exclusiveOn`](https://fig.io/docs/reference/option#exclusiveon) for these two options

src/dbt.ts
```diff
@@ -16,11 +118,117 @@ const completionSpec: Fig.Spec = {
...
           description: "Exit immediately if a single model fails to build",
         },
+        {
+          name: "--use-colors",
+          description: "Default value -- colorize run logs",
+        },
+        {
+          name: "--no-use-colors",
+          description: "Disable log colorizing",
+        },
```
QuiiBz, line 131: Same as above for `exclusiveOn`

QuiiBz (commented):
Thanks a lot for this first contribution! I left some comments to resolve before merging this. Also, it looks like a lot of options are shared, could you move them to a top-level const to avoid duplicating them?

QuiiBz:
Hi @lilroberts, just pinging you to know if you still want to resolve the code review, or if we should handle it ourselves. Thanks!

mschrage:
Hey @QuiiBz can you make these changes?

QuiiBz (approved):
Thanks for the initial PR @lilyroberts!

PR #712, opened 2021-11-07
feat(ni): add ni specs

QuiiBz:
Feature

No specs for ni

Add specs for ni (ni, nr, nu, nrm)

**Additional info:**
https://github.com/antfu/ni

### DIFF
--- a/src/ni.ts
+++ b/src/ni.ts
@@ -41,6 +41,10 @@ const completionSpec: Fig.Spec = {
         },
       ],
     },
+    {
+      name: ["-h", "--help"],
+      description: "Output usage information",
+    },
   ],
   args: {
     name: "package",
--- a/src/ni.ts
+++ b/src/ni.ts
@@ -1,0 +1,52 @@
+import { npmSearchGenerator } from "./npm";
+
+const completionSpec: Fig.Spec = {
+  name: "ni",
+  description: "Use the right package manager - install",
+  options: [
+    {
+      name: "-g",
+      description:
+        "Operates in 'global' mode, so that packages are installed into the prefix folder instead of the current working directory",
+    },
+    {
+      name: "-D",
+      description: "Package will appear in your `devDependencies`",
+    },
+    {
+      name: "-P",
+      description: "Save package to your `peerDependencies`",
+    },
+    {
+      name: "-O",
+      description: "Save package to your `optionalDependencies`",
+    },
+    {
+      name: "--frozen",
+      description: "Don't generate a lockfile and fail if an update is needed",
+    },
+    {
+      name: "-C",
+      description: "Change directory",
+      args: [
+        {
+          name: "directory",
+          description: "The directory to move",
+          template: "folders",
+        },
+        {
+          name: "target",
+          description: "The target directory",
+          template: "folders",
+        },
+      ],
+    },
+  ],
+  args: {
+    name: "package",
+    generators: npmSearchGenerator,
+    debounce: true,
+    isVariadic: true,
+  },
+};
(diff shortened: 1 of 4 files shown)


src/ni.ts
```diff
@@ -0,0 +1,52 @@
+import { npmSearchGenerator } from "./npm";
+
+const completionSpec: Fig.Spec = {
+  name: "ni",
+  description: "Use the right package manager - install",
+  options: [
```
fedeci, line 6: Doesn't `ni` have a `--help` opt?

src/nrm.ts
```diff
@@ -0,0 +1,35 @@
+import { dependenciesGenerator } from "./npm";
+
+const completionSpec: Fig.Spec = {
```
fedeci, line 3: Same for the --help

src/nu.ts
```diff
@@ -0,0 +1,12 @@
+const completionSpec: Fig.Spec = {
+  name: "nu",
+  description: "Use the right package manage - upgrade",
+  options: [
```
fedeci, line 4: ditto `--help`

src/nrm.ts
```diff
@@ -0,0 +1,35 @@
...
+  description: "Use the right package manage - remove",
+  options: [
+    {
+      name: "-g",
+      description:
+        "Operates in 'global' mode, so that packages are installed into the prefix folder instead of the current working directory",
+    },
+    {
+      name: "-D",
+      description: "Package will appear in your `devDependencies`",
```
fedeci, line 14:
```suggestion
      description: "Package will be removed from your `devDependencies`",
```

src/nrm.ts
```diff
@@ -0,0 +1,35 @@
+import { dependenciesGenerator } from "./npm";
+
+const completionSpec: Fig.Spec = {
+  name: "nrm",
+  description: "Use the right package manage - remove",
+  options: [
+    {
+      name: "-g",
+      description:
+        "Operates in 'global' mode, so that packages are installed into the prefix folder instead of the current working directory",
```
fedeci, line 10:
```suggestion
        "Operates in 'global' mode, so that packages are removed from the prefix folder instead of the current working directory",
```

src/nrm.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    {
+      name: "-D",
+      description: "Package will appear in your `devDependencies`",
+    },
+    {
+      name: "-P",
+      description: "Save package to your `peerDependencies`",
```
fedeci, line 18:
```suggestion
      description: "Remove package from your `peerDependencies`",
```

src/nrm.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    {
+      name: "-O",
+      description: "Save package to your `optionalDependencies`",
```
fedeci, line 22:
```suggestion
      description: "Remove package from your `optionalDependencies`",
```

fedeci (changes requested):

src/ni.ts
QuiiBz: It's not built-in, but fallbacks to the current package manager. Not sure if we should add it or not

fedeci: I think all the three package managers have the same option name for help, I would add it just in case someone needs it and doesn't find it. I'll leave the burden to you however and approve it 😄

fedeci (approved):
Great!

src/ni.ts
QuiiBz: Alright!

PR #715, opened 2021-11-09
feat(deno): update spec to 1.16

SeparateRecords:
Spec update

Not up to date.

Updates Deno spec to 1.15 & 1.16. This modernises the spec a bit (sets `requiresEquals` where appropriate) and, because requiresEquals optional arguments are supported now, adds an environment variable generator for `--allow-env`.

**Additional info:**
Deno 1.16 technically isn't out yet, but will be in less than 24 hours. It doesn't actually introduce anything in the CLI, all the new stuff was in 1.15.
(Still saying it's updated to 1.16 so future-me doesn't get confused when updating to 1.17)

Also, I've been gone for a while for a bunch of _really_ stupid reasons that I've resolved now, all to do with my flaky setup. Hopefully I'll be back to being active in this community :)

### DIFF
--- a/src/deno.ts
+++ b/src/deno.ts
@@ -1,8 +1,9 @@
 // This file largely follows the same structure (not order) as deno/cli/flags.rs:
 // https://github.com/denoland/deno/blob/main/cli/flags.rs
 
-// All objects marked with '// requiresEquals: true' are Clap args with '.require_equals(true)'
-// TODO: When fig supports this option (or something like it), uncomment the arguments.
+// Fig doesn't automatically insert an '=' where an option's argument requires
+// an equals and the argument isn't optional. That's why you'll see a lot of
+// `insertValue: "--name="` in this spec.
 
 /**
  * Equivalent to the `"filepaths"` template, but boosts the priority of files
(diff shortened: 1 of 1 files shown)


SeparateRecords:
Hah! 1.16 was released _while I was writing the PR description_.

image

SeparateRecords:
Reverting to draft - didn't realise that Fig spawns a new shell so `env` won't show the user's variables. Will fix tomorrow

src/deno.ts
```diff
@@ -4,9 +4,6 @@
 // All objects marked with '// requiresEquals: true' are Clap args with '.require_equals(true)'
 // TODO: When fig supports this option (or something like it), uncomment the arguments.
```
QuiiBz, line 5: Can we remove this TODO comment?

src/deno.ts
```diff
@@ -811,7 +829,6 @@ const denoDoc: Fig.Subcommand = {
               throw new Error(`Output data was JSON, but was not an array`);
             }
           } catch (err) {
-            console.error("Returning early due to error:", err);
```
QuiiBz, line 814: That can still be useful if anything goes wrong, to make debugging easier

QuiiBz (changes requested):
Thanks for this contribution!

src/deno.ts
SeparateRecords: Thanks for noticing that. I removed that suggestion from the find&replace results and totally forgot to remove it 😆

QuiiBz (approved):
LGTM, thanks for keeping this spec up to date!

PR #716, opened 2021-11-09
[fastly] Update `fastly` spec to 1.1.0

kailan:


QuiiBz (approved):
LGTM, thanks for keeping `fastly` updated!

PR #727, opened 2021-11-16
fix: -f option definition for docker volume rm

giovannibenussi:
Bug fix

When you type `docker volume rm`, Fig auto-suggests to use the `-f, --force` flag as expected. However, when you select that option from the autocompletion menu, you get `docker volume rm -f,\ --force` instead of `docker volume rm -f` as expected.

Fig inserts the correct value for the `-f` option: `docker volume rm -f`

**Additional info:**
 \-

### DIFF
--- a/src/docker.ts
+++ b/src/docker.ts
@@ -5344,7 +5344,7 @@ const completionSpec: Fig.Spec = {
           },
           options: [
             {
-              name: "-f, --force",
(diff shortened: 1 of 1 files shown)


fedeci (approved):

QuiiBz (approved):
Quick and easy fix, thanks a lot!

PR #732, opened 2021-11-18
added ls-remote to git spec

zp4rker:
feature

`ls-remote` is missing in spec

`ls-remote` is now in spec

**Additional info:**


QuiiBz (approved):
LGTM, thanks for this small contribution!

PR #733, opened 2021-11-18
feat: ansible + ansible-playbook completion specs

zp4rker:
Adds `ansible` and `ansible-playbook` completion specs. Implements https://github.com/withfig/autocomplete/issues/621

Not present.

Implements them.

**Additional info:**
This is only 2 out of 12+ different ansible commands (all separate binaries), these are by far the most used commands out of the lot. I may implement more in the future but don't have a need for it at the moment (aka laziness). May possibly also add generators for enhancement in the future such as giving host pattern suggestions based on inventory but not right now (again due to laziness).


src/ansible-playbook.ts
```diff
@@ -0,0 +1,232 @@
+const completionSpec: Fig.Spec = {
+  name: "ansible-playbook",
+  description:
+    "Tuns Ansible playbooks, executing the defined tasks on the targeted hosts",
+  args: {
+    name: "pattern",
+    description: "Host pattern",
+    isOptional: false,
```
QuiiBz, line 8: You can remove `isOptional` when the value is `false`, it's the default value!

src/ansible-playbook.ts
```diff
@@ -0,0 +1,232 @@
...
+      description: "Specify the vault identity to use",
+      args: {
+        name: "vault ID",
+      },
+    },
+    {
+      name: "--vault-password-file",
+      description: "Specify a vault password file",
+      args: {
+        name: "vault password file",
```
QuiiBz, line 71: I think this argument can have a [template](https://fig.io/docs/reference/arg#template) of `filepaths`

src/ansible-playbook.ts
```diff
@@ -0,0 +1,232 @@
...
+      description: "Override the connection timeout in seconds",
+      args: {
+        name: "timeout",
+      },
+    },
+    {
+      name: ["--connection", "-c"],
+      description: "Connection type to use",
+      args: {
+        name: "connection type",
```
QuiiBz, line 216: Could we provide [suggestions](https://fig.io/docs/reference/arg#suggestions) for this argument?

src/ansible.ts
```diff
@@ -0,0 +1,236 @@
+const completionSpec: Fig.Spec = {
+  name: "ansible",
+  description:
+    "Define and run a single Ansible task 'playbook' against a set of hosts",
+  args: {
+    name: "pattern",
+    description: "Host pattern",
+    isOptional: false,
```
QuiiBz, line 8: Same as above, you can remove this

src/ansible.ts
```diff
@@ -0,0 +1,236 @@
...
+    {
+      name: "--list-hosts",
+      description: "Outputs a list of matching hosts; does not execute",
+    },
+    {
+      name: "--playbook-dir",
+      description:
+        "Since this tool does not use playbooks, use this as a substitute playbook directory.This sets the relative path for many features including roles/ group_vars/etc",
+      args: {
+        name: "playbook dir",
```
QuiiBz, line 24: Same as above for the template, but a template of `folders`

src/ansible.ts
```diff
@@ -0,0 +1,236 @@
...
+      description: "Specify the vault identity to use",
+      args: {
+        name: "vault ID",
+      },
+    },
+    {
+      name: "--vault-password-file",
+      description: "Specify a vault password file",
+      args: {
+        name: "vault password file",
```
QuiiBz, line 43: Same as above for the template of `filepaths`

QuiiBz (changes requested):

src/ansible-playbook.ts
zp4rker: It's not as simple as a static list of suggestions for this one, it's based on what connection plugins are installed. Could possibly be implemented in the future using generators.

QuiiBz (approved):
LGTM, thanks a lot for these amazing specs!

PR #734, opened 2021-11-18
feat: add scc spec

SeparateRecords:
Adds a spec for scc (https://github.com/boyter/scc)

There's one generator in there, the other generators will require more work so if I do implement them it'll definitely be a separate PR.


src/scc.ts
```diff
@@ -0,0 +1,376 @@
...
+          { name: "embedded", icon: "fig://icon?type=string" },
+        ],
+      },
+    },
+    {
+      name: "--count-as",
+      description:
+        'Count extension as language [e.g. jsp:htm,chead:"C Header" maps extension jsp to html and chead to C Header]',
+      // TODO: contextual suggestions (this will use `languagesGenerator`)
+      //       (trigger on ':', remove on ',')
```
QuiiBz, line 105: Can you tackle this TODO?

src/scc.ts
```diff
@@ -0,0 +1,376 @@
...
+          { name: "sql-insert", icon: "fig://icon?type=string" },
+        ],
+      },
+    },
+    {
+      name: "--format-multi",
+      description:
+        "Have multiple format output overriding --format [e.g. tabular:stdout,csv:file.csv,json:file.json]",
+      // TODO: contextual suggestions.
+      // suggest formats immediately, then files on `:`, then reset to formats on `,`
```
QuiiBz, line 155: Same as above

QuiiBz (changes requested):
Thanks for this contribution, love the types and comments, that really helps!

src/scc.ts
SeparateRecords: As in to have these generators done for this PR? As I said in the PR message I think it would be better to add these complicated generators in a follow up PR.
For this particular option, the syntax scc expects is `ext:Name,ext2:"Spaced Name"` so there's a fair amount of logic that would have to go into that generator (and I just don't have the time to implement that right now, but will next week)

SeparateRecords: Was just about to sleep and figured I should clarify, I'm happy to leave this PR open until I have time if you'd prefer that!

QuiiBz: Thanks for the clarification, I think we can merge this PR - and the generators can be done in a new one as you said!

QuiiBz (approved):
LGTM, looking for the generators PR :)

PR #745, opened 2021-11-22
feat: update `expo` and `expo-cli` completion spec to 4.13.0

kbrandwijk:
Updated completion spec for Expo CLI

Completion for expo-cli@4.12.3

Completion for expo-cli@4.13.0

**Additional info:**
-


src/expo-cli.ts
```diff
@@ -111,18 +211,7 @@ const completionSpec: Fig.Spec = {
           description: "Clear all credentials stored on Expo servers",
           args: {
             name: "boolean",
-            suggestions: [
```
fedeci, line 114: This is a generation error, are you using a custom tool to autogenerate the spec? I would suggest you to do not add `suggestions` prop if it is an empty array so our [merge](https://github.com/withfig/autocomplete-tools/blob/main/packages/autocomplete-tools/merge/index.ts) tool can work properly. I still need to add proper docs for it but if you have doubts configuring it, please tag me.

kbrandwijk: Yes, the generation script is made by the folks over at Expo, in merely ran it on the latest version of their CLI, so I didn't do any manual post-processing.

I can manually go over it and remove the empty ones, or if it doesn't break anything, I can instead submit a PR on their repo to the script for future versions, based on your comment.

fedeci: Yeah, opening a PR to fix the problem at the source seems the best solution to me. I suggest you to wait for that PR to be merged and then regenerate and merge the spec because having to manually resolve all the problems would be really time consuming.

kbrandwijk: Hmmm I have just compared this new version to the old one, and apparently, it just hasn't generated any suggestions at all. So the problem isn't that it keeps empty ones, the problem is that it hasn't generated them. I'll pull this back into draft until I get that sorted out.

fedeci: It can't generate suggestions, they were manually added. What we can do to avoid adding those manually on every update is not to generate empty suggestions and using the merge tool :)

kbrandwijk: Apparently, something was commented out in the generation. This now matches what it was before, no more empty suggestions.

fedeci: Great!

fedeci:
Let's wait for @QuiiBz review, if it is autogenerated it LGTM. Is expo changing all that stuff in a minor version or is it just a property ordering problem?

QuiiBz:
> Is expo changing all that stuff in a minor version or is it just a property ordering problem?

They may have added some subcommands, so that's why the ordering has moved. LGTM!

kbrandwijk:
> > Is expo changing all that stuff in a minor version or is it just a property ordering problem?
> 
> They may have added some subcommands, so that's why the ordering has moved. LGTM!

Yeah, most of it is ordering, the _actual_ changes were minimal, I just ran into a minor difference, figured out it was autogenerated anyways, so thought: why not...

fedeci:
Great work @kbrandwijk!

QuiiBz (approved):

Issue #605, opened 2021-09-24
[maven] Support


fwesss:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

maven

### Which statement makes the most sense?

This completion spec does not exist

### Issue Details

Maven is a build tool typically used with java.

https://maven.apache.org/ref/3.8.1/maven-embedder/cli.html

zp4rker:
@Cool-Runningz sorry if you already started on it, but I noticed its been a while so I decided to do it 😬 

PR submitted @ https://github.com/withfig/autocomplete/pull/746

PR #746, opened 2021-11-23
feat: add mvn (Maven) completion spec

zp4rker:
New completion spec

Non-existent

Existent

**Additional info:**
Resolves https://github.com/withfig/autocomplete/issues/605

### DIFF
--- a/src/mvn.ts
+++ b/src/mvn.ts
@@ -1,0 +1,235 @@
+const completionSpec: Fig.Spec = {
+  name: "mvn",
+  description: "Maven - a Java based project management and comprehension tool",
+  subcommands: [
+    {
+      name: "my_subcommand",
+      description: "Example subcommand",
+      subcommands: [
+        {
+          name: "my_nested_subcommand",
+          description:
+            "Nested subcommand, example usage: 'mvn my_subcommand my_nested_subcommand'",
+        },
+      ],
+    },
+  ],
+  options: [
+    {
+      name: ["--also-make", "-am"],
+      description: "Also build projects required by project list",
+    },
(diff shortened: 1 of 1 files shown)


src/mvn.ts
```diff
@@ -0,0 +1,259 @@
...
+        name: "server password",
+      },
+    },
+    {
+      name: ["--file", "-f"],
+      description:
+        "Force the use of an alternate POM file (or directory with pom.xml)",
+      args: {
+        name: "path",
+        suggestions: ["filepaths", "folders"],
```
QuiiBz, line 76: This should actually be:

```suggestion
        template: ["filepaths", "folders"],
```

src/mvn.ts
```diff
@@ -0,0 +1,259 @@
...
+      name: ["--show-version", "-V"],
+      description: "Display version information",
+    },
+    {
+      name: ["--debug", "-X"],
+      description: "Produce execution debug output",
+    },
+  ],
+  args: {
+    name: "goal/phase",
```
QuiiBz, line 220: Could you also add a description for this argument?

QuiiBz (changes requested):

src/mvn.ts
zp4rker: I blame GitHub Copilot 😂 

Fixed.

src/mvn.ts
zp4rker: Thanks for noticing, fixed!

QuiiBz (approved):
LGTM, thanks a lot for this contribution!

Issue #748, opened 2021-11-24
[ssh] known_hosts generator producing duplicates


zp4rker:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [X] I would like to work on this.

### What CLI tool does this relate to?

ssh

### Which statement makes the most sense?

There is a bug with this completion spec

### Issue Details

The generator which reads from `known_hosts` is currently producing duplicates - 2 duplicates in my experience (making it a total of 3 suggestions of the same thing).

![Screen Shot 2021-11-24 at 8 47 31 pm](https://user-images.githubusercontent.com/13144755/143216978-aef84bfa-614f-4d14-81b7-c04a3863834d.png)

Submitting a PR for it right after submitting this issue.

PR #749, opened 2021-11-24
[ssh] fix duplicates issue with known_hosts generator

zp4rker:
Bug fix

`known_hosts` generator produces duplicates.
Screen Shot 2021-11-24 at 8 47 31 pm

All suggestions from `known_hosts` are unique

**Additional info:**
Resolves https://github.com/withfig/autocomplete/issues/748

### DIFF
--- a/src/ssh.ts
+++ b/src/ssh.ts
@@ -5,16 +5,17 @@ const knownHosts: Fig.Generator = {
   postProcess: function (out, tokens) {
     return out
       .split("\n")
-      .map((line) => String(knownHostRegex.exec(line)))
-      .filter((value, index, self) => value && self.indexOf(value) === index)
-      .map((knownHost) => {
-        if (knownHost != null) {
-          return {
-            name: (tokens[1].endsWith("@") ? tokens[1] : "") + knownHost, // also suggest when user@ is provided
-            description: "SSH host",
-          };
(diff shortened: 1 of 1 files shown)


src/ssh.ts
```diff
@@ -3,15 +3,18 @@ const knownHostRegex = /(?:[a-zA-Z0-9-]+\.)+[a-zA-Z0-9]+/; // will match numeric
...
-          description: "SSH host",
-        };
-      }
-    });
+    return out
+      .split("\n")
+      .map((line) => String(knownHostRegex.exec(line)))
+      .filter((value, index, self) => value && self.indexOf(value) === index)
+      .map((knownHost) => {
+        if (knownHost != null) {
```
QuiiBz, line 11: Thanks for this contribution! I think we can move this last `if` statement in the above `filter`, that would make more sense, and we can then return the object in the `map` directly without the `return` keyword.

zp4rker: Although you didn't directly, you made me realise there was a bug: it was suggesting `null` as an option because I was casting to string before checking if null. Moved the null check into the earlier map to resolve it.

mschrage (approved):

mschrage:
Thanks so much for patching this! ❤️

PR #754, opened 2021-11-29
feat: added support for SWC

bart-krakowski:
This PR adds support for [SWC](https://swc.rs/docs/usage/cli).

### DIFF
--- a/src/swc.ts
+++ b/src/swc.ts
@@ -1,0 +1,127 @@
+const completionSpec: Fig.Spec = {
+  name: "swc",
+  description:
+    "SWC is a super-fast compiler written in rust; producing widely-supported javascript from modern standards and typescript",
+  options: [
+    {
+      name: ["--filename", "-f"],
+      description:
+        "Filename to use when reading from stdin. This will be used in source maps and errors",
+    },
+    {
+      name: "--config-file",
+      description: "Path to a .swcrc file to use",
+      args: {
+        name: "CONFIG_FILE",
+        description: "Path to a .swcrc file to use",
+        template: ["filepaths"],
+      },
+    },
+    {
+      name: "--env-name",
+      description:
+        "The name of the 'env' to use when loading configs and plugins. Defaults to the value of SWC_ENV, or else NODE_ENV, or else development",
+    },
+    {
+      name: "--no-swcrc",
+      description: "Whether or not to look up .swcrc files",
+    },
+    {
+      name: "--ignore",
+      description: "List of glob paths to not compile",
+      args: {
+        name: "PATH",
+        description: "A glob path to not compile",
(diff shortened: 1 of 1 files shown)


src/swc.ts
```diff
@@ -0,0 +1,127 @@
+const completionSpec: Fig.Spec = {
+  name: "swc",
+  description:
+    "SWC is a super-fast compiler written in rust; producing widely-supported javascript from modern standards and typescript",
+  options: [
+    {
+      name: ["--filename", "-f"],
+      description:
+        "Filename to use when reading from stdin. This will be used in source maps and errors",
```
mschrage, line 9: This option seems like it should take an argument.

src/swc.ts
```diff
@@ -0,0 +1,127 @@
...
+        name: "CONFIG_FILE",
+        description: "Path to a .swcrc file to use",
+        template: ["filepaths"],
+      },
+    },
+    {
+      name: "--env-name",
+      description:
+        "The name of the 'env' to use when loading configs and plugins. Defaults to the value of SWC_ENV, or else NODE_ENV, or else development",
+    },
```
mschrage, line 24: `--env-name` also takes an argument!

src/swc.ts
```diff
@@ -0,0 +1,127 @@
...
+      name: ["--watch", "-w"],
+      description: "Watch for changes and recompile",
+    },
+    {
+      name: ["--quiet", "-q"],
+      description: "Suppress compilation output",
+    },
+    {
+      name: ["--source-maps", "-s"],
+      description: "Values: true|false|inline|both",
```
mschrage, line 57: `--source-maps` takes an argument. It can be `true|false|inline|both`, so you could add those to a `suggestions` array on the `arg` object.

mschrage (changes requested):
Just a few small tweaks and we can get this merged in! Thank you so much for contributing @
bart-krakowski

src/swc.ts
```diff
@@ -21,6 +26,11 @@ const completionSpec: Fig.Spec = {
       name: "--env-name",
       description:
         "The name of the 'env' to use when loading configs and plugins. Defaults to the value of SWC_ENV, or else NODE_ENV, or else development",
+      args: {
+        name: "ENV_NAME",
```
fedeci, line 30: Probably we can provide suggestions for common environment names here like 
```suggestion
        name: "ENV_NAME",
        suggestions: ["production", "development", "test"],
```

fedeci (approved):
Apart from that comment it looks good! Thanks!

mschrage (approved):

mschrage:
Thanks for the quick turn around @bart-krakowski!

PR #765, opened 2021-12-02
feat: add `just` spec

SeparateRecords:
Adds the `just` spec (https://github.com/casey/just)
`just` is a task runner, similar in usage to `make` but focused on running tasks instead of being a build system.
You can create a "justfile" and invoke any amount of "recipes". These recipes can also take arguments (see below).

**Additional info:**
Relevant issue: withfig/fig#638
TL;DR: Fig doesn't have a way to model this behaviour natively, so it all has to be implemented in a generator. I've gone to great lengths to make things "just work" 😉

Current to-do:
- [x] Suggest aliases in args generator
- [x] Factor out args generator into functions
- [x] Use said functions in the other generators
- [x] Re-write comments for clarity

Default justfile

Aliases work too

Specified justfile

Set variables

### DIFF
--- a/src/just.ts
+++ b/src/just.ts
@@ -64,7 +64,7 @@ interface Binding {
  * - `--justfile=name`
  */
 function getJustfilePath(tokens: string[]): string | null {
-  const flagRe = /^(-[A-Za-eg-z]*?f=?|--justfile(?:=|\b))/;
+  const flagRe = /^(-[A-Za-eg-z]*f=?|--justfile(?:=|$))/;
   for (const [index, token] of tokens.entries()) {
     const match = token.match(flagRe);
     if (match === null) {
--- a/src/just.ts
+++ b/src/just.ts
@@ -81,15 +81,14 @@ function getRecipeSuggestions(
   for (const [name, recipe] of Object.entries(justfile.recipes)) {
     suggestions.push({
       name,
-      displayName: showRecipeParameters
-        ? getRecipeNameWithParams(recipe)
-        : name,
-      description: recipe.doc ?? "recipe",
+      displayName: showRecipeParameters ? formatPrettyRecipeName(recipe) : name,
+      description: recipe.doc ?? "Recipe",
       icon: "fig://icon?type=command",
     });
   }
 
-  // Now the aliases. Like the git aliases, these don't list their usage
+  // Now the aliases. Like the git aliases, these don't list their usage.
+  // Also like with the git spec, these use the commandkey icon.
   for (const [name, alias] of Object.entries(justfile.aliases)) {
     suggestions.push({
       name,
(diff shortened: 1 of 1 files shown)


src/just.ts
```diff
@@ -0,0 +1,507 @@
...
+  args: {
+    // This would normally say "recipes" but because this is also used
+    // for the recipes' arguments it needs to be more generic.
+    name: "args",
+    isVariadic: true,
+    isOptional: true,
+    optionsCanBreakVariadicArg: false,
+    generators: {
+      // TODO: Check if this is necessary
+      trigger: " ",
```
SeparateRecords, line 439: Is this necessary? I was under the (possibly incorrect) impression that generators are run once per token by default if not specified otherwise

import-brain (approved):
LGTM, I'm just not sure if that whitespace trigger is necessary.

SeparateRecords:
> LGTM, I'm just not sure if that whitespace trigger is necessary.

I didn't think so either but wasn't 100% sure, thanks for the second opinion 😊

import-brain:
> > LGTM, I'm just not sure if that whitespace trigger is necessary.
> 
> I didn't think so either but wasn't 100% sure, thanks for the second opinion 😊

To be honest, I'm not sure either, but if the whitespace is necessary, you can always revert commit fce4a9e and force push here :)

src/just.ts
```diff
@@ -0,0 +1,541 @@
...
+    } else if (parameter.kind === "star") {
+      parts.push(`[${parameter.name}...]`);
+    } else {
+      console.error(`Unreachable: unknown kind '${parameter.kind}'`);
+    }
+  }
+  return parts.join(" ");
+}
+
+interface RecipeArityMapping {
```
QuiiBz, line 152: Any reason why using an `interface` here, but `type` above?

src/just.ts
```diff
@@ -0,0 +1,541 @@
...
+        ],
+      },
+    },
+    {
+      name: ["-c", "--command"],
+      insertValue: "--command '{cursor}'",
+      description:
+        "Run an arbitrary command with the working directory, .env, overrides, and exports",
+      args: {
+        name: "command",
```
QuiiBz, line 240: I think we can add the [`isCommand`](https://fig.io/docs/reference/arg#iscommand) flag here

QuiiBz (changes requested):
Didn't know `just`, that looks really cool.

src/just.ts
SeparateRecords: There was a reason but it's not relevant anymore, I'll make it consistent with `interface` where possible

src/just.ts
SeparateRecords: Totally right! I'll add that.

SeparateRecords:
Realised there's another opportunity to make things simpler. Will implement that right now!

src/just.ts
SeparateRecords: Finally figured out how to word my rationale! It was gnawing at me all day, I just couldn't put it into words haha

I use `interface` when I'm planning on implementing a type, but if it's to model something external (eg data from an API or CLI) then I use `type`.

That distinction doesn't matter at all but that's what I do in my own projects, so it's habit 😁

fedeci (approved):
Thanks!

import-brain (approved):

SeparateRecords:
Just want to check, is there something holding this up from being merged?

mschrage:
Looks good to me! Let's get this merged 🤩

PR #778, opened 2021-12-09
Add spec for Atlassian's @forge/cli

slatlasdev:
Feature

Support for Atlassian's @forge/cli
https://www.npmjs.com/package/@forge/cli
https://developer.atlassian.com/platform/forge/getting-started/

**Additional info:**

I am an Atlassian developer, working on this project for our hackathon. 
I can confirm this is up to date with the latest version of the tool.


src/forge.ts
```diff
@@ -0,0 +1,699 @@
...
+          name: ["-h", "--help"],
+          description: "Display help for command",
+          priority: 49,
+        },
+      ],
+      args: {
+        name: "install|uninstall",
+        suggestions: ["install", "uninstall"],
+      },
+    },
```
fedeci, line 25: Is this the command used to generate the spec? If it is it should be removed.

src/forge.ts
```diff
@@ -0,0 +1,699 @@
+// This spec is automatically generated by a customised version of @withfig/commander
+
+const completionSpec: Fig.Spec = {
+  name: "",
```
fedeci, line 4:
```suggestion
  name: "@forge/cli",
```

fedeci (approved):
Thanks! I am interested in what you have improved in @withfig/commander, it would be cool to eventually add the changes to the main tool too!

mschrage:
Hey @slatlasdev! Thank you so much for contributing. It's super cool that this was your hackathon project! I hope the team likes it.  🥳✨ 

Also, I'm really curious if you can imagine Fig improving the internal development experience at Atlassian? Are there any internal scripts or CLI tools that your team uses regularly?

(My co-founder, @brendanfalk, is from Australia as well and is a huge Atlassian fan!)

src/forge.ts
slatlasdev: This command is used to install the bash/zsh autocompletions. As Fig is also doing this, I have moved this command to lowest priority in my new PR https://github.com/withfig/autocomplete/pull/781

slatlasdev:
Hi @mschrage I'm a big fan of thing so far and I've been trying to spread the word internally.

This is definitely a tool I can see improving our developer experience. I wanted to start off with forge as it is a public facing tool so we could get the best benefit from it, but there are many other internal tools that I would love to add as well.

I just made a couple tweaks as suggested in this PR which I've added to another PR here: https://github.com/withfig/autocomplete/pull/781

There are some improvements I'd like to add to `@withfig/commander`, starting with the simplest https://github.com/withfig/autocomplete-tools/pull/26

Other custom modifications I made for this project were:
- Auto capitalising descriptions, as this is enforced by ESLint
- Converting certain arrays to scalars when there's only one item, as enforced by ESLint (is this strictly necessary? I know it makes it slightly cleaner to look at but adds more type juggling)
- Setting `.hidden` to true, or removing a command entirely (not sure which is better atm) when `cmd._hidden` is true in Commander
- When an argument has multiple options, eg `<install|uninstall>` adding both `install` and `uninstall` as options to the `arg.suggestions`. This use of the pipe might be specific to our use case and not a commander thing so not sure if it would make sense to include generally or not.

fedeci:
Great to hear your thoughts here @slatlasdev!
- I 100% understand it is super annoying to see eslint complaining about every single stupid thing like non-capitalized descriptions or one-item-arrays but we choose to standardize some aspects of the repo so that everything looks more cleaner and organized.
- About commander we chose to set the `.hidden` flag to true when cmd._hidden is true because even if the command should not be listed it is still available. We can add some kind of setting to change this behaviour in the @withfig/commander integration though!
- Concerning the multiple args I personally see those more like subcommands than as arguments, am I right?

slatlasdev:
- Hmm, that's odd, I was running `lint:fix` and those capitalisation/array issues still remained, hence why I had to modify the generation script 🤔 
- I couldn't see @withfig/commander setting the `.hidden` flag [currently](https://github.com/withfig/autocomplete-tools/blob/main/packages/commander/src/index.ts#L139), is this something that's in the pipeline? But yes, I think this makes sense and is as good as removing the command.
- Ahh yes, this is a good point, it seems like they should be subcommands instead. It looks like we need to modify our usage of commander rather than changing the generation script.

slatlasdev:
Created https://github.com/withfig/autocomplete-tools/pull/27/ for the hidden change

fedeci:
- That's strange because we set the fixer for conventional descriptions https://github.com/withfig/autocomplete-tools/blob/main/packages/eslint-plugin-fig-linter/rules/conventional-descriptions.js, I will need to investigate this issue!
- Oops, probably hidden is set in the oclif integration? So many to handle that I can't remember anything😅

PR #787, opened 2021-12-13
Add `git blame` support

a7med-mahmoud:
### **What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
Feature

### **What is the current behavior? (You can also link to an open issue here)**
Fig doesn't have completion for the `git blame` subcommand.

### **What is the new behavior (if this is a feature change)?**
Fig has great support for the `git blame` subcommand.

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -608,6 +608,151 @@ const completionSpec: Fig.Spec = {
   ],
   subcommands: [
     {
+      name: "blame",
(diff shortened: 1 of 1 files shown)


mschrage (approved):
This looks great @a7med-mahmoud! Can't believe we missed the `blame` subcommand.

PR #788, opened 2021-12-14
updated lint

queilawithaQ:
changed ubuntu to macos.


fedeci (commented):
It doesn't change anything for the linting process and ubuntu is:
- More available on gh servers
- Probably faster than macos

Thanks for your proposal but I am gonna close this one!

Issue #791, opened 2021-12-16
Add extension option to filepath template


a7med-mahmoud:
It would be great if I can use the filepath template in the spec and choose a specific file extension to target.

mschrage:
Hey @a7med-mahmoud,

Can you sketch out the API you imagine for this?

The current approach would be to use the `filterTermplateSuggestions` function to check the extension, but it potentially makes sense to add some syntactic sugar over this?

a7med-mahmoud:
It would be cool if I can just do this:
```js
{
   ...
   template: {
     name: "filepath",
     options: {
        extensions: ["png", "jpg", "jpeg"]
     }
   }
}
```
Or maybe
```js
{
   ...
   template: {
     name: "filepath",
     options: {
        filter: /png|jpe?g/i
     }
   }
}
```

PR #794, opened 2021-12-18
refactor: add `filepaths` sugar

fedeci:

**Additional info:**


src/dotnet/dotnet-migrate.ts
```diff
@@ -1,27 +1,16 @@
...
-            return fileName.endsWith(suffix) || fileNames.includes(fileName);
-          }
-
-          return false;
-        });
-      },
-    },
+    generators: filepaths({
+      extensions: ["sln"],
+      equals: ["project.json", "global.json"],
```
fedeci, line 12:
```suggestion
      match: /(project\.json)|(global\.json)/g,
```

fedeci:
I still need to implement logic for extensions containing multiple dots, but the most of the work is done.

⚠️ For reviewers, please give a look to replaced generators and ask when if you find something strange because I may have implemented something in a bad way😄

src/_common/generators.ts
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable */
```
fedeci, line 1: Todo: disable fig plugin just for this file.

src/_common/generators.ts
```diff
@@ -0,0 +1,65 @@
...
+  const extensionsSet = new Set(extensions);
+  const equalsSet = new Set(equals);
+  return {
+    template: "filepaths",
+    filterTemplateSuggestions: (suggestions) => {
+      const filtered = suggestions.filter(({ name, type }) => {
+        if (
+          typeof acceptFolders !== undefined &&
+          acceptFolders === (type === "folder")
+        ) {
```
SeparateRecords, line 42: I think this might have a bug when `acceptFolders` is explicitly set to `false` and `type` is anything other than folder.
| `acceptFolders` | `type` | result |
| --- | --- | --- |
| `undefined` | `"folder"` | ✅ condition isn't met |
| `undefined` | `"file"` | ✅ condition isn't met |
| `true` | `"folder"` | ✅ `return true` |
| `true` | `"file"` | ✅ `true === false`, condition isn't met |
| `false` | `"folder"` | ✅ `false === true`, condition isn't met |
| `false` | `"file"` | ❓ `false === false`, so we `return true` which skips the checks below |

If the goal is to not accept folders unless this option is `true`, maybe this would work?
```suggestion
        if (!acceptFolders && type === "folder") {
```

SeparateRecords (commented):
Great job on this so far! Love the ratio of added/deleted 😁✂️

I haven't had too much of a look at the specs that were changed, but I did skim them and it all looks very sound to me. Will try to have a closer look tonight, but I highly suspect you nailed it!

There was only one thing that wasn't super clear to me, are folders supposed to be hidden by default? I think the answer is yes but just wanted to double check

src/_common/generators.ts
fedeci: In my mind `acceptFolders` equal to:
- `undefined` should skip the check and just evaluate the file based on his name (with match, equals...)
- `false` should return false if the type is a folder
- `true` should return true if the type is a folder
I don't know why I used a XNOR, I'll correct that.
We can't use
```js
if (!acceptFolders && type === "folder") {
```
because that would return true in some cases if acceptFolders is false.

src/_common/generators.ts
```diff
@@ -0,0 +1,65 @@
...
+  acceptFolders?: boolean;
+  priorities?: {
+    folder?: number;
+    files?: number;
+  };
+  icons?: {
+    folder?: string;
+    files?: string;
+  };
+}
```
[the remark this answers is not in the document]
fedeci: Love it, thanks!
I was thinking about allowing to pass and entire `Suggestion` object for files and for folders so we can extract relevant properties to update from there.

src/_common/generators.ts
SeparateRecords: Ahh I understand, thank you for explaining that! 🙂
So if `acceptFolders` is undefined, it's treated just like a file, but if it's true/false then conditionally suggest it. I think that's perfectly reasonable!

---
I'm writing this at 12:48 am so keep that in mind as you read the rest of this 😉

This might be surprising behaviour to people who only skim docs, since at least to me an optional boolean means only two behaviours.
Feel free to shoot this down btw - I think a string union could make this clearer (even a `"string" | boolean` union if that's more your style!)

```typescript
/**
 * Defines how folders are suggested.
 * 
 * - **Default:** `filter` will treat folders like files, filtering based on the name.
 * - `always` will always suggest folders.
 * - `never` will never suggest folders.
 */
suggestFolders?: "filter" | "always" | "never";

// ...

const {
  suggestFolders = "filter",
} = options;

// ...

if (type === "folder" && suggestFolders !== "filter") return suggestFolders === "always";
```

Then maybe `acceptFolders` could be used to conditionally add the `"folders"` template? (i.e., accept the _command_ on folder suggestion)

fedeci: I was planning on renaming the variable to something of more understandable, but that's even better! I'll change it later!

fedeci:
Ready!

a7med-mahmoud:
This is great! I can close #791 when this is merged

SeparateRecords:
Great catch @a7med-mahmoud! I'll link this PR so it'll close your issue once it's merged 😊

import-brain (approved):

src/_common/generators.ts
```diff
@@ -0,0 +1,88 @@
...
+  suggestFolders?: "filter" | "always" | "never";
+  /**
+   * Set properties of suggestions of type "file".
+   */
+  editFileSuggestions?: Omit<Fig.Suggestion, "name" | "type">;
+  /**
+   * Set properties of suggestions of type "folder".
+   */
+  editFolderSuggestions?: Omit<Fig.Suggestion, "name" | "type">;
+}
```
mschrage, line 27: Does it make sense for these to be closures instead? Or is that more flexibility than we want to expose?

src/_common/generators.ts
```diff
@@ -0,0 +1,88 @@
...
+  /** Show suggestions where the name matches this expression */
+  matches?: RegExp;
+  /**
+   * Defines how folders are suggested.
+   *
+   * - **Default:** `filter` will treat folders like files, filtering based on the name.
+   * - `always` will always suggest folders.
+   * - `never` will never suggest folders.
+   */
+  suggestFolders?: "filter" | "always" | "never";
```
mschrage, line 18: I'm not sure if "filter" should be the default behavior for `suggestFolders`. For instance, if a filepath generator should only load files with a `yaml` extension, we should still show users files so that they can navigate to the target file if it is not in the current working directory.

src/dotnet/dotnet-migrate.ts
```diff
@@ -1,27 +1,16 @@
...
-            return fileName.endsWith(suffix) || fileNames.includes(fileName);
-          }
-
-          return false;
-        });
-      },
-    },
+    generators: filepaths({
+      extensions: ["sln"],
+      matches: /(project\.json)|(global\.json)/g,
```
mschrage, line 12: Should this use `equals` rather than `matches` ?

```suggestion
      equals: ["project.json", "global.json"]
```

Also potentially should also have `showFolders: "always"`

mschrage (approved):
This looks great @fedeci, although the more I think about it, we should probably move this to a separate npm package.

It can then be imported like:
`import { filepath } from @fig/generators`

And can be used by people working on specs that are not bundled in withfig/autocomplete.

Only downside is that maybe it makes contributing a little more difficult?

src/_common/generators.ts
fedeci: I agree with you. What made me use 'filter' as a default is the fact that several specs are currently using that behaviour.

src/_common/generators.ts
fedeci: I think that's too much and I was even thinking about not adding the capability of customization at all. This should be just some light sugar, if we need callbacks and so on we can just fallback to the standard way imho.

src/dotnet/dotnet-migrate.ts
fedeci: I used a regex because it was using .includes js method before. Equals requires strict equality of the string while that regex allows characters before and after `project.json` and `global.json`. I'll check if those files can for e.g. have a prefix and update the pr accordingly.

fedeci:
I know that's another big refactoring but we may turn this repo in a monorepo and move also some stuff from autocomplete-tools back here. (Like the types and the linter). Otherwise the solution is to create a new package in autocomplete-tools🤔

src/_common/generators.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+        if (equalsSet.has(name)) return true;
+        if (matches && matches.test(name)) return true;
+        if (includes && name.includes(includes)) return true;
+        // handle extensions
+        const [_, ...extensions] = name.split(".");
+        if (extensions.length >= 1) {
+          let stackedExtensions = "";
+          for (let i = 0; i < extensions.length; i++) {
+            stackedExtensions = extensions[i] + stackedExtensions;
```
SeparateRecords, line 68: Minor: Could this be a `for...of` loop? It's a little cleaner - no more correct though and just as readable, feel free to ignore!
```suggestion
          for (const extension of extensions) {
            stackedExtensions = extension + stackedExtensions;
```

SeparateRecords: Actually I think I'm missing something here - given this file name `example.js.rs.ts.json.yml`, by the end of this would `stackedExtensions` be `"ymljsontsrsjs`?

fedeci: In fact that loop should be reversed😄
My oversight
```suggestion
          for (let i = extensions.length; i >= 0; i--) {
            stackedExtensions = extensions[i] + stackedExtensions;
```

fedeci: Exactly, and I also missed the dots🤦‍♂️

src/dotnet/dotnet-migrate.ts
mschrage: I think that was an error with the initial generator: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-migrate

src/_common/generators.ts
mschrage: My sense is that in many cases this an error/oversight of the previous implementation rather than something intentional.

import-brain (approved):

src/_common/generators.ts
```diff
@@ -0,0 +1,92 @@
...
+          do {
+            if (extensionsSet.has(stackedExtensions)) {
+              return true;
+            }
+            i--;
+            // `i` may become -1 which is not a valid index, but the extensionSet check at the beginning is not run in that case,
+            // so the wrong extension is not evaluated
+            stackedExtensions = [extensions[i], stackedExtensions].join(".");
+          } while (i >= 0);
+        }
```
SeparateRecords, line 77: I tried this out, it did end up accessing `extensions[-1]`, but`[undefined, "example"].join(".")` returns `".example"` so it mostly works.

Here's a shorter version, it's a little simpler 🙂

```suggestion
        const lastDot = name.lastIndexOf(".");

        // This loop will never run if there are no extensions
        for (let i = lastDot; i > -1; i--) {
          if (name[i] !== ".") continue;

          // Slice from after the dot to exclude it from the extension
          if (extensionsSet.has(name.slice(i + 1))) return true;
        }
```

---

I tried doing this by looping `lastIndexOf` but it turns out...

```javascript
// Index of the final dot, if we start searching from -1?
".a.b".lastIndexOf(".", -1) === 0
// Of course it's zero... why not, I guess 🥴
```

fedeci: We really don't care that it returns that string because when it is -1 the next check is not performed (the dowhile stops looping). I know it is not intuitive but thats the simplest possible way that does not involve splice.

fedeci:
Moved the implementation to https://github.com/withfig/autocomplete-tools/pull/29

fedeci:
Ready to go live.

mschrage (approved):

PR #798, opened 2021-12-21
feat: add autocomplete for pre-commit

miketheman:
feature - new spec

no spec exists

new completions!

**Additional info:**
Popular CLI tool: https://pre-commit.com/

I left a couple of notes inline to help improve this in the future.

### DIFF
--- a/src/pre-commit.ts
+++ b/src/pre-commit.ts
@@ -1,0 +1,309 @@
+const hooksInConfig: Fig.Generator = {
+  // TODO: replace with a YAML parser when it's available
+  // See: https://github.com/withfig/fig/issues/733
+  script:
+    'sed -nE "s/[[:space:]]+-[[:space:]]+id:[[:space:]](.*)$/\\1/p" .pre-commit-config.yaml',
+  postProcess: (output) =>
+    output.split("\n").map((hookId) => ({
+      name: `${hookId}`,
+    })),
+};
+
+const help: Fig.Option = {
+  name: ["-h", "--help"],
+  description: "Show help message and exit",
+};
+
+const color: Fig.Option = {
+  name: "--color",
+  description: "Whether to use color in output. Defaults to `auto`",
+  args: {
+    name: "color",
+    suggestions: ["auto", "always", "never"],
+  },
+};
+
+const config: Fig.Option = {
+  name: ["--config", "-c"],
+  description: "Path to alternate config file",
+  args: {
+    name: "CONFIG",
+  },
+};
+
+const globalOptions: Fig.Option[] = [help, color, config];
+
+const hookStages = [
+  "pre-commit",
+  "pre-merge-commit",
+  "pre-push",
+  "prepare-commit-msg",
+  "commit-msg",
+  "post-commit",
+  "post-checkout",
+  "post-merge",
+  "post-rewrite",
+];
+
+const hookType: Fig.Option = {
+  name: ["-t", "--hook-type"],
+  description: "Type of hook to install",
+  args: {
+    name: "HOOK_TYPE",
+    suggestions: hookStages,
+  },
+};
+
+const commonRepoOptions: Fig.Option[] = [
+  {
+    name: ["--verbose", "-v"],
+  },
+  {
+    name: ["--all-files", "-a"],
+    description: "Run all files in the repo",
+  },
+  {
+    name: "--files",
+    description: "Specific filenames to run hooks on",
+    args: {
+      name: "FILES",
+      isVariadic: true,
+      template: ["filepaths", "folders"],
+    },
+  },
+  {
+    name: "--show-diff-on-failure",
+    description: "When hooks fail, run `git diff` directly afterward",
+  },
+  {
+    name: "--hook-stage",
+    description: "The stage during which the hook is fired",
+    args: {
+      suggestions: hookStages,
+    },
+  },
+  {
+    name: "--remote-branch",
+    description: "Remote branch ref used by `git push`",
+    args: {
+      name: "REMOTE_BRANCH",
+      // TODO: am I allowed to import a generator for `git.ts`?
+    },
+  },
+  {
+    name: "--local-branch",
+    description: "Local branch ref used by `git push`",
+    args: {
+      name: "LOCAL_BRANCH",
+    },
+  },
+  {
+    name: ["--from-ref", "--source", "-s"],
+    description:
+      "Represents the original ref in a `from_ref...to_ref` diff expression. For `pre-push` hooks, this represents the branch you are pushing to. For `post-checkout` hooks, this represents the branch that was previously checked out",
+    args: {
+      name: "FROM_REF",
+    },
+  },
+  {
+    name: ["--to-ref", "--origin", "-o"],
+    description:
+      "Represents the destination ref in a `from_ref...to_ref` diff expression. For `pre-push` hooks, this represents the branch being pushed. For `post-checkout` hooks, this represents the branch that is now checked out",
+    args: {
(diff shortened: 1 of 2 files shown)


src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
+const hooksInConfig: Fig.Generator = {
+  // TODO: replace with a YAML parser when it's available
+  // See: https://github.com/withfig/fig/issues/733
+  script:
+    'sed -nE "s/[[:space:]]+-[[:space:]]+id:[[:space:]](.*)$/\\1/p" .pre-commit-config.yaml',
```
miketheman, line 5: Tested on macOS 12.1, which uses BSD `sed`. Since fig targets macOS for now, this seemed safe enough until https://github.com/withfig/fig/issues/733 is implemented

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+    args: {
+      suggestions: hookStages,
+    },
+  },
+  {
+    name: "--remote-branch",
+    description: "Remote branch ref used by `git push`",
+    args: {
+      name: "REMOTE_BRANCH",
+      // TODO: am I allowed to import a generator for `git.ts`?
```
miketheman, line 90: I didn't want to copy a bunch of generator code over needlessly, but figured I'd inquire. Seems like there may be a benefit for other git-related utilities to have a set of shared util functions somewhere?

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+    },
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "pre-commit",
+  description: "Pre-commit",
+  args: {},
+  options: [
+    globalOptions[0], // top level only takes `help`
```
miketheman, line 168: I wasn't sure about this approach, but it works. Happy to take any advice on a better accessor for the `help` without having to split out the option from `glopbalOptions`

src/pre-commit.ts
fedeci: This is not the only command we will need to update in the future, it's great for the moment!

src/pre-commit.ts
fedeci: Sure you can, we are doing something similar for `yarn`, `npm` and `pnpm` specs

src/pre-commit.ts
fedeci: Can't we just do this?
```suggestion
    help,
```

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+            name: "REPO",
+          },
+          isRepeatable: true,
+        },
+      ],
+    },
+    {
+      name: "clean",
+      description: "Clean out pre-commit files",
+      options: [...globalOptions],
```
fedeci, line 204: super nitpicky:
```suggestion
      options: globalOptions,
```

src/pre-commit.ts
mschrage: We are adding something like this now. See https://github.com/withfig/autocomplete/pull/794

And I agree that git generators are good candidates to be shared between specs.

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+    },
+    {
+      name: "gc",
+      description: "Clean unused cached repos",
+      options: [...globalOptions],
```
fedeci, line 209:
```suggestion
      options: globalOptions,
```

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+          name: "--allow-missing-config",
+          description:
+            "Whether to allow a missing `pre-commit` configuration file or exit with a failure code",
+        },
+      ],
+    },
+    {
+      name: "migrate-config",
+      description: "Migrate list configuration to new map configuration",
+      options: [...globalOptions],
```
fedeci, line 249:
```suggestion
      options: globalOptions,
```

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+        name: "hook",
+        description: "A single hook-id to run",
+        isOptional: true,
+        generators: hooksInConfig,
+      },
+    },
+    {
+      name: "sample-config",
+      description: "Produce a sample .pre-commit-config.yaml file",
+      options: [...globalOptions],
```
fedeci, line 265:
```suggestion
      options: globalOptions,
```

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+    },
+    {
+      name: "uninstall",
+      description: "Uninstall the pre-commit script",
+      options: [...globalOptions, hookType],
+    },
+    {
+      name: "help",
+      description: "Show help for a specific command",
+      options: [...globalOptions],
```
fedeci, line 304:
```suggestion
      options: globalOptions,
```

src/pre-commit.ts
```diff
@@ -0,0 +1,309 @@
...
+    suggestions: ["auto", "always", "never"],
+  },
+};
+
+const config: Fig.Option = {
+  name: ["--config", "-c"],
+  description: "Path to alternate config file",
+  args: {
+    name: "CONFIG",
+  },
```
mschrage, line 31:
```suggestion
  args: {
    name: "CONFIG",
    template: "filepaths"
  },
```

src/pre-commit.ts
miketheman: Awesome! Will look at those now.

src/pre-commit.ts
mschrage: It seems like `globalOptions` are _already_ split out. I would just use the existing `help` value to make this more explicit.

miketheman: Oh, that's sneaky. Will update!

src/pre-commit.ts
fedeci, line 304: This accepts an arg with the name of the command, right? We can probably extract the keys from `Object.keys(completionSpec.subcommands) and remove "help" from there` 

@mschrage would it be nice to have some other sugar to get this behaviour for free adding a new template? Commander has this kind of help too if I am not wrong and many specs may benefit of it.

mschrage (changes requested):
This looks good to me! Also really agree with your instincts around shared generators.

Just a few nits:

fedeci (changes requested):
I added a few suggestions, but apart from those nits this PR looks great. Thank you!

fedeci:
Reviews overlapping, nice😆

src/pre-commit.ts
miketheman: I have a refactor locally, but it depends on #800 being merged first, so will hold off pushing just yet.

src/pre-commit.ts
mschrage: I think Cobra does as well. Definitely makes sense to me to add this in as `template: "subcommands"` or `template: "help"`.

fedeci: I will add it then!

src/pre-commit.ts
miketheman: Pushed 570460225550ffddafe85b3e6a384c8952139162

miketheman:
I think this is ready for review/merge!

mschrage:
Hey @miketheman! Thanks so much for this contribution 🤩

### Next steps

🐥 **If you have a Twitter**: tweet that `precommit` is now supported and we'll retweet from the [@fig](https://twitter.com/fig) account :)

🛠 **Are you interested in being a codeowner?** You'll be pinged to review any future PRs for the `precommit` completion spec.

PR #801, opened 2021-12-23
feat: dynamically complete `yarn create` CLIs

a7med-mahmoud:

Feature

Having `create-*` CLIs hardcoded in an array that Fig searches in.

Search for `create-*` term using npm API and load CLI spec dynamically if exists.


fedeci (approved):
Thanks!

import-brain (approved):

PR #804, opened 2021-12-26
Export all specs from index.ts

mschrage:
Exports all specs as modules. (Autogenerated in index.js)
It is challenging to use `@withfig/autocomplete` when installed as an `npm` package.
`import { git } from '@withfig/autocomplete'` or `import * as specs from '@withfig/autocomplete'`

**Additional info:**

The default export (a `string[]` of all completion files in `src/**.*.ts`) remains the same to ensure backwards compatibility


mschrage:
@fedeci @QuiiBz @sullivan-sean this PR still doesn't produce a fully idiomatic TypeScript npm package since the `.d.ts` types aren't generated. Any thoughts on how to resolve this?

Screen Shot 2021-12-26 at 6 21 20 PM

mschrage:
It might be as easy as generating a `.d.ts` file like this and referencing it in the `types` field of `package.json`:

```ts
declare module '@withfig/autocomplete' {
    export declare const git: Fig.Spec
    export declare const aws: Fig.Spec
    export declare const heroku: Fig.Spec
    ...
    const completions: string[]
    export default completions
}
```

generate-index.ts
```diff
@@ -25,10 +25,26 @@ const files = glob
...
+const commandTypeDeclarations = commands
+  .map((x) => `export declare const ${normalize(x)}: Fig.Spec`)
+  .join("\n");
+
+fs.writeFileSync(
+  "build/index.d.ts",
+  `${commandTypeDeclarations}
+
+const completions: string[]
+export default completions
```
fedeci, line 48:
```suggestion
declare const completions: string[]
export { completions as default }
```

generate-index.ts
```diff
@@ -25,10 +25,26 @@ const files = glob
...
+  `var e=[${files.join(",")}];export{e as default};${commandExports}`
+);
+
+const commandTypeDeclarations = commands
+  .map((x) => `export declare const ${normalize(x)}: Fig.Spec`)
+  .join("\n");
+
+fs.writeFileSync(
+  "build/index.d.ts",
+  `${commandTypeDeclarations}
```
fedeci, line 45:
```suggestion
  `import "@withfig/autocomplete-types"
  
${commandTypeDeclarations}
```

fedeci:
We are exporting ESModules from the packages, so we also need to specify `type: "module"` in the `package.json`.

fedeci (approved):

fedeci:
I don't think we need autocomplete-types as a dep, then we should be ready to merge it.

miketheman:
It appears the [build&release step is still broken](https://github.com/withfig/autocomplete/runs/4751487793?check_suite_focus=true#step:4:11) as a result of this change.
I corrected the lint error in #821 , and poked at how to resolve the ts-node issue, but got lost in the mess of what `ts-node` is even trying to do, or updating versions, etc.

PR #809, opened 2021-12-29
Feat/eb basics autocomplete

nmassardot:
Feature

It doesn't support eb (elasticbeanstalk from AWS) command in CLI

Added the suggestion and init autocomplete with it options

**Additional info:**

- [X] Add me as codeowner of new files (you will be added as a codeowner of the files added in this PR)

### DIFF
--- a/src/eb.ts
+++ b/src/eb.ts
@@ -27,7 +27,7 @@ const completionSpec: Fig.Spec = {
           description: "List of child directories to initialize",
         },
         {
-          name: ["-p", "--plataform"],
+          name: ["-p", "--platform"],
           description: "The platform version to use",
           args: {
             name: "plataform-version",
--- a/src/eb.ts
+++ b/src/eb.ts
@@ -1,6 +1,16 @@
 const completionSpec: Fig.Spec = {
   name: "eb",
   description: "AWS Elastic Beanstalk",
+  subcommands: [
+    {
+      name: "init",
+      description: "Initialize a new AWS Elastic Beanstalk application",
+      args: {
+        name: "name",
+        description: "The name of the application",
+      },
+    },
+  ],
 };
 
 export default completionSpec;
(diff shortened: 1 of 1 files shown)


src/eb.ts
```diff
@@ -0,0 +1,48 @@
...
+          args: {
+            name: "keyname",
+          },
+        },
+        {
+          name: "--modules",
+          description: "List of child directories to initialize",
+        },
+        {
+          name: ["-p", "--plataform"],
```
fedeci, line 30:
```suggestion
          name: ["-p", "--platform"],
```

fedeci (commented):
A small typo but it looks good, thanks!

src/eb.ts
nmassardot: Corrected!

mschrage (approved):
Quick question: is the standalone elastic beanstalk CLI different from the one bundled with the aws CLI? 

We have an autogenerated completion spec foe that here: https://github.com/withfig/autocomplete/blob/master/src/aws/ebs.ts

nmassardot:
> Quick question: is the standalone elastic beanstalk CLI different from the one bundled with the aws CLI?
> 
> We have an autogenerated completion spec for that here: https://github.com/withfig/autocomplete/blob/master/src/aws/ebs.ts

The one I added comes from [here](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb3-cmd-commands.html), they are different, am I right?

fedeci:
@mschrage I also had the same doubt, but seems they have different subcommands. Maybe `eb` is an older deprecated version of `aws eb`?

nmassardot:
@fedeci I don't think it is, I have found it in many guides for deploying directly to eb. Anyways, let me know, because i have been using it the last days to make deploys in my work

fedeci:
No problem in merging this, even if it was deprecated which would mean that still a lot of users are using this!
Thanks for your contribution🎉

PR #810, opened 2021-12-29
feat: add `pass` completion spec

whobutsb:
New Feature: Added [pass](https://www.passwordstore.org/) auto completion.

**Additional info:**
Man page for pass - https://git.zx2c4.com/password-store/about/

- [ ] Add me as codeowner of new files (you will be added as a codeowner of the files added in this PR)

### DIFF
--- a/src/pass.ts
+++ b/src/pass.ts
@@ -108,10 +108,7 @@ const completionSpec: Fig.Spec = {
     {
       name: "git",
       description: "Password store git functions",
-      args: {
-        name: "git-command-arguments",
-        description: "Git command arguments",
-      },
+      loadSpec: true,
     },
     {
       name: "version",
@@ -284,21 +281,12 @@ const completionSpec: Fig.Spec = {
       name: "grep",
       description:
         "Searches inside each decrypted password file for search-string. Grep options can be used",
+      loadSpec: true,
       args: {
         name: "pass-name",
         description: "The password name you want to grep for",
         isOptional: false,
       },
-      options: [
-        {
-          name: ["-e", "--regexp"],
-          description:
-            "Use PATTERN as the pattern. This can be used to specify multiple search patterns, or to protect a pattern beginning with a hyphen (-). (-e is specified by POSIX.)",
-          args: {
-            name: "pattern",
-          },
-        },
-      ],
     },
   ],
 };
--- a/src/pass.ts
+++ b/src/pass.ts
@@ -35,6 +35,26 @@ const completionSpec: Fig.Spec = {
   name: "pass",
   description:
     "Pass - stores, retrieves, generates, and synchronizes passwords securely",
+  args: {
+    name: "pass-name",
+    description: "The password you want to show",
+    generators: listPasswords,
+    isOptional: false,
+  },
+  options: [
+    {
+      name: ["--clip", "-c"],
+      description: "Copy the password to the clipboard",
+    },
+    {
+      name: ["--qrcode", "-q"],
+      description: "Display a QRcode of the password",
+    },
+    {
+      name: "--help",
+      description: "Show help for pass",
+    },
+  ],
   subcommands: [
     {
       name: "init",
@@ -277,12 +297,6 @@ const completionSpec: Fig.Spec = {
           },
         },
       ],
-    },
-  ],
-  options: [
-    {
-      name: ["--help", "-h"],
-      description: "Show help for pass",
     },
   ],
 };
--- a/src/pass.ts
+++ b/src/pass.ts
@@ -146,7 +146,8 @@ const completionSpec: Fig.Spec = {
       args: {
         name: "pass-name",
         description: "The password name.",
-        isOptional: false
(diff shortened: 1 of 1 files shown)


whobutsb:
👍

fedeci (dismissed):
Can you fix the listing errors, please? `npm run lint:fix` should do the work!

whobutsb:
@fedeci Linted!  Thank you for the consideration!

whobutsb:
Question for the fig team.  Within the `pass` command you can use `pass git` which allows you to interact with the local `~/.password-store` git repo and run normal git commands.  Is there a way to reference the fig `git` completion spec when someone types in `pass git`?  I doesn't make sense to rewrite the entire git spec.  Currently I have the following for the arguments: 

```
    {
      name: "git",
      description: "Password store git functions",
      args: {
        name: "git-command-arguments",
        description: "Git command arguments",
      },
    },
```

The `pass grep` command is the same as well where you can grep inside the files to find a particular line.  It would be great to show the git and grep specs as completions when you enter the sub commands.

Thanks!

whobutsb:
> Question for the fig team. Within the `pass` command you can use `pass git` which allows you to interact with the local `~/.password-store` git repo and run normal git commands. Is there a way to reference the fig `git` completion spec when someone types in `pass git`? I doesn't make sense to rewrite the entire git spec. Currently I have the following for the arguments:
> 
> ```
>     {
>       name: "git",
>       description: "Password store git functions",
>       args: {
>         name: "git-command-arguments",
>         description: "Git command arguments",
>       },
>     },
> ```
> 
> The `pass grep` command is the same as well where you can grep inside the files to find a particular line. It would be great to show the git and grep specs as completions when you enter the sub commands.
> 
> Thanks!

Figured this out!  Added `loadSpec: true` to the `pass git` and `pass grep` commands, to get the full completion.  Super useful!

mschrage:
Hey @whobutsb! `loadSpec` is the right property, but I'm surprised it passes typechecking as a boolean.

I think it should be `loadSpec: "git"` and `loadSpec: "grep"` instead

whobutsb:
@mschrage Updated!  Thanks for the hint.  Yeah it was working fine for both commands with `true`, interesting stuff!

whobutsb (commented):
Ran `npm run lint:fix`.

src/pass.ts
```diff
@@ -0,0 +1,293 @@
...
+  postProcess: (output) => {
+    return output.split("\n").map((dir) => ({
+      name: dir,
+    }));
+  },
+};
+
+const listGpgIds: Fig.Generator = {
+  script: function () {
+    return `gpg --list-keys --with-colons --with-fingerprint | awk -F: '/^uid:/ { print $10 }`;
```
mschrage, line 25: Can you try to move the parsing logic to Javascript? Sometimes people have different versions of `awk` and `sed` installed, so it is best to keep as much as possible in the JS runtime.

src/pass.ts
```diff
@@ -0,0 +1,293 @@
+const listPasswords: Fig.Generator = {
+  script: function () {
+    return `grep -r -l '' $HOME/.password-store --exclude-dir=.git | sed "s|$HOME/.password-store/||g"`;
```
mschrage, line 3: Can we remove `sed` and put this logic in JS?

src/pass.ts
```diff
@@ -0,0 +1,293 @@
...
+  postProcess: (output) => {
+    return output.split("\n").map((password) => ({
+      name: password.replace(".gpg", ""),
+    }));
+  },
+};
+
+const listDirectories: Fig.Generator = {
+  script: function () {
+    return `ls -dR1a $HOME/.password-store/*/ | sed "s|$HOME/.password-store/||g"`;
```
mschrage, line 14: Same as above?

mschrage (changes requested):
Thanks for contributing @whobutsb! A few small tweaks but this looks great overall :)

src/pass.ts
whobutsb: @mschrage - I'm going to remove this function altogether.  I wasn't 100% sure that the output that I was receiving was the correct argument for the `pass init` command.

src/pass.ts
whobutsb: Updated now using: 
```
const listPasswords: Fig.Generator = {
  script: function () {
    return `grep -r -l '' $HOME/.password-store --exclude-dir=.git`;
  },
  postProcess: (output) => {
    return output.split("\n").map((password) => ({
      name: password.split(".password-store/").pop().replace(".gpg", ""),
    }));
  },
};
```
This will remove everything before the `.password-store/` directory and output just the pass-names and directories.

src/pass.ts
whobutsb: Same as previous.

fedeci (approved):
Thanks!

import-brain (approved):

mschrage (approved):

PR #812, opened 2022-01-02
Add Exercism support

hendrikmitk:
Feature: Add completion spec for Exercism CLI

Currently, there are no completion options for the `exercism` command.

There will be completion options available for the `exercism` command.

**Additional info:**
https://exercism.org/docs/using/solving-exercises/working-locally

### DIFF
--- a/src/exercism.ts
+++ b/src/exercism.ts
@@ -1,0 +1,173 @@
+const helpOption: Fig.Option[] = [
+  {
+    name: ["-h", "--help"],
+    description: "More information about command",
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "exercism",
+  description: "Solve coding exercises in your chosen programming languages",
+  subcommands: [
+    {
+      name: ["configure", "c"],
+      description: "Configure the command-line client",
+      options: [
+        ...helpOption,
+        {
+          name: ["-a", "--api"],
+          description: "API base url",
+          args: {
+            name: "url",
+            isOptional: false,
+          },
+        },
+        {
+          name: "--no-verify",
+          description: "Skip online token authorization check",
+        },
+        {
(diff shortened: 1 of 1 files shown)


src/exercism.ts
```diff
@@ -0,0 +1,173 @@
+const helpOption: Fig.Option[] = [
+  {
+    name: ["-h", "--help"],
+    description: "More information about command",
+  },
+];
```
mschrage, line 6: I saw that this `helpOption` is included everywhere -- it's a nice touch of polish, but you can get the same behavior with less copy and pasting by setting [`isPersistent`](https://fig.io/docs/reference/option#ispersistent) to true!

Just make sure `helpOption` added to the root subcommand and the `--help` flag will appear for all child subcommands as well. 

```suggestion
const helpOption: Fig.Option[] = [
  {
    name: ["-h", "--help"],
    description: "More information about command",
    isPersistent: true,
  },
];
```

mschrage (changes requested):
This looks great! Thanks so much for contributing @hendrikmitk! Just one small tweak and we can get this merged in:

src/exercism.ts
hendrikmitk: Thanks a lot for the feedback @mschrage! Using `isPersistent` is indeed a nice touch that makes the whole thing a bit clearer. I removed the `helpOption` because the same option had already been available at the top level.

This also made me notice that the global flags were not yet available for all subcommands as intended.

mschrage (approved):

mschrage:
Thanks so much for contributing @hendrikmitk! 🥳

hendrikmitk:
Thanks to you and your team for building such a great tool @mschrage! 🙂

PR #813, opened 2022-01-02
Feat/eb deploy

nmassardot:
Feature

It doesn't support all the commands of `eb` commands in CLI, si now `deploy` and `abort` was added

Added the autocomplete for `deploy` and `abort`, with it suggestions

**Additional info:**
My objective is to add all the available commands for `eb`


mschrage (approved):
Thanks for adding the missing `deploy` subcommand! Looks good to me :)

Really appreciate the contribution! 🥳

PR #814, opened 2022-01-02
Add `swift` spec

dlbuckley:
> **What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

### Add spec for `swift` tools.

> **What is the current behavior? (You can also link to an open issue here)**

### No current support.

> **What is the new behavior (if this is a feature change)?**

### All current `swift` commands have ben added.

> **Additional info:**

### There is some possible followup work with some generators, but this covers the basic command interface for now.

### DIFF
--- a/src/swift.ts
+++ b/src/swift.ts
@@ -1,0 +1,929 @@
+const commonOptions: Fig.Option[] = [
+  {
+    name: ["--help", "-h", "-help"],
+    description: "Show help information",
+  },
+  {
+    name: "--version",
+    description: "Show the version",
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "swift",
+  description: "Swift compiler",
+  options: [
+    {
+      name: ["--help", "-h", "-help"],
+      description: "Show help information",
+    },
+  ],
+  subcommands: [
+    // swift build
+    {
+      name: "build",
+      description: "Build sources into binary products",
+      options: [
+        ...commonOptions,
+        {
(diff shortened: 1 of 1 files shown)


dlbuckley:
I've noticed that I've missed a couple of options. I will add these at some point over the next day or so.

dlbuckley:
I've now added the missing options and this PR should be ready for review.

import-brain (approved):

mschrage (changes requested):
This looks great! The only thing missing is a [`parserDirective`](https://fig.io/docs/reference/subcommand#parserdirectives) at the root level to inform the parse that `flagsArePosixNoncompliant` since the options start with a single dash.

dlbuckley:
I missed that flag when going though the docs, thank for pointing it out. The change has now been made.

mschrage (approved):

dlbuckley:
@mschrage It looks like the CI job needs to be kicked off again as it looks like it failed due to an action download issue.

fedeci:
Thanks!

kabiroberai:
I'm a bit late to the party but fwiw it should be possible to automate the generation of specs for all but the root swift[c] subcommand, using the experimental SPM spec generator at https://github.com/withfig/fig-swift-argument-parser. Is it possible to extract subcommands into their own spec files?

fedeci:
@kabiroberai do you mean something like aws subcommands?

PR #816, opened 2022-01-04
feat: `vsce` support

a7med-mahmoud:

Feature

No support for `vsce`: the vscode extension manager.

Added support for the `vsce` with all the commands and options.

### DIFF
--- a/src/npx.ts
+++ b/src/npx.ts
@@ -83,6 +83,9 @@ const suggestions: Fig.Suggestion[] = [
     name: "ignite-cli",
     icon: "🔥",
   },
+  {
+    name: "vsce",
+  },
 ];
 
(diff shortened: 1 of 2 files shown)


a7med-mahmoud:
I can add a publishers generator for better completion but that will come with a cost.
The problem is I have to run `vsce ls-publishers` which requires access to keychain info of `vscode-vsce` which distracts the completion and needs to "Allow always" the access of it to work well.

a7med-mahmoud:
Please share your opinions on whether I should add the publishers completion with the risk described above.

mschrage:
I think you shouldn’t add the generator

mschrage (approved):
👍

import-brain (approved):

Issue #817, opened 2022-01-05
[git] Add support for gitmoji inside git commit message


yaronuliel:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [x] I would like to work on this.

### What CLI tool does this relate to?

git

### Which statement makes the most sense?

Something else

### Issue Details

When getting autocomplete for `git commit -m ` it would be great to get a list of available gitmojies
Gitmoji ( [https://gitmoji.dev/](https://gitmoji.dev/) ) is de-facto standard for  tagging commit messages visually, and is supported by all major git providers (github, bitbucket, gitlab) - having autocompletion for it is super useful, since it is hard to remember the whole list, with its nuance 

I didn't dig deep enough to check if fig already supports autocompletion for textual option values ("open answers"), but this would probably also be a good idea for possible integrations with tools like grammerly, etc...

brendanfalk:
@yaronuliel Cleaning things up and closing this as we decided gitmoji is best served as a mixin as opposed to in the spec.

PR #818, opened 2022-01-06
Update git spec to support gitmoji suggestions on commit message

yaronuliel:
close withfig/autocomplete#817

Feature

autocomplete for `git commit -m ` shows suggestions for standard gitmoji

https://user-images.githubusercontent.com/5229872/148439032-f846a981-bc4a-431f-95d5-ae4ab8c0c878.mov

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -500,6 +500,361 @@ const headSuggestions = [
   },
 ];
 
+const gitMojies = [
+  {
+    code: ":art:",
+    emoji: "🎨",
+    description: "Improve structure / format of the code",
+  },
+  {
+    code: ":zap:",
+    emoji: "⚡️",
+    description: "Improve performance",
+  },
+  {
+    code: ":fire:",
+    emoji: "🔥",
+    description: "Remove code or files",
+  },
+  {
+    code: ":bug:",
+    emoji: "🐛",
+    description: "Fix a bug",
+  },
+  {
+    code: ":ambulance:",
+    emoji: "🚑️",
+    description: "Critical hotfix",
+  },
+  {
+    code: ":sparkles:",
+    emoji: "✨",
+    description: "Introduce new features",
+  },
+  {
+    code: ":memo:",
+    emoji: "📝",
+    description: "Add or update documentation",
+  },
+  {
+    code: ":rocket:",
+    emoji: "🚀",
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
I don't think we should provide these suggestions for everyone. Gitmoji is a separate feature from git and I would rather prefer a separate spec for the gitmoji cli or some kind of mechanism to detect if user has gitmoji cli installed and only in that case provide suggestions for gitmoji in git spec

mschrage (changes requested):
Hey @yaronuliel! I think this is awesome but I'm not sure if this should be the default behavior in the `git` completion spec. 

Is there anyway to detect if someone is using gitmoji by running a CLI command and then we could show these suggestions? 

Perhaps by running `gitmoji --version` to see if the CLI tool is installed or by running `git log --pretty=format:%s` and checking if the last n commits start with an emoji?

yaronuliel:
@mschrage gitmoji doesn't require any external tool, and works with all major git providers without any additional software, or tools. In my opinion it is not different from suggesting `Bugfix: {cursor}` which is also just a convention. Searching for previous gitmoji in git history is indeed an option, but it means that it won't work for new git projects (maybe we can leave the :tada: (`:tada:` which represents new project git commit, and will usually be added to first commit in a repo) available for new projects, and show the rest only if it was used

@fedeci The [gitmoji cli](https://github.com/carloscuesta/gitmoji-cli) in a utility that helps using gitmojies (as I mentioned in the issue #817 it is quite hard to remember all the possible notations, and this is what gitmoji-cli is for). Having a completion for `$ gitmoji-cli ...` will off course require a separate spec.

Anyway, I do agree that adding a config option that will allow users to switch the feature on/off is required (the big question is about the default)

fedeci:
I think you can use [personal mixins](https://fig.io/docs/guides/personal-mixins) and we should probably work on some way to allow mixins to be shared.
WDYT @mschrage?

mschrage:
@yaronuliel I agree with @fedeci here. This makes more sense as a [mixin](https://fig.io/docs/guides/personal-mixins) than part of the git completion spec.

I'm going to close this PR for now. But I love the idea and definitely want to find a way for you to share this with the community.

PR #820, opened 2022-01-07
[fastly] Update `fastly` spec to 1.4.0

kailan:
https://github.com/fastly/cli/releases/tag/v1.4.0

This release includes a new `--service-name` flag which can be used in place of `--service-id`, hence the updates to most command definitions.

also: happy new year, team! ✨


fedeci (approved):
Thanks!

PR #831, opened 2022-01-12
fix: build script

fedeci:
Instead of becoming crazy trying to fix some problems with modules just use vanilla node along with an old good js file😄


mschrage:
Wasn't this resolved by #828 ?

mschrage (approved):

PR #832, opened 2022-01-12
feat(eb): add `clone` subcommand

nmassardot:
Feature

It doesn't support all the `eb` commands of AWS

Now fig supports `clone` subcommands and all it's options. Also, `open` subcommand was added

**Additional info:**
I will be working to add all the commands of `eb`


fedeci (approved):
Thanks!

PR #840, opened 2022-01-14
fix(build): target `ES6`

fedeci:
This should fix issues concerning JS features that are not yet supported by safari or that are not supported by old safari versions.


mschrage (approved):
This looks good to me. Perhaps @sullivan-sean can weigh in as well.

fedeci:
I'm merging this as it may solve issues with older os.

PR #844, opened 2022-01-14
fix(grep): add `isRepeatable` to option

### LINKED ISSUE #842: Fig fails to autosuggest CLI options more than once, even if proper usage
### Description:

Particularly for the standard `grep` utility (and I expect others), Fig does not autosuggest an option more than once. For example, `grep`'s `--exclude-dir` parameter (documentation [here](https://linux.die.net/man/1/grep)), which is used to exclude a single directory from search, can be repeated multiple times in order to exclude multiple directories, such as: `grep -r foo --exclude-dir foo/bar --exclude-dir foo/baz`

However, if `--exclude-dir` is entered once (autosuggesting correctly the first time), Fig fails to autosuggest the same option afterwards.

### Details:
|macOS|Fig|Shell|
|-|-|-|
|12.0.1|Version 1.0.53 (B339)|/bin/zsh|
<details><summary><code>fig diagnostic</code></summary>
<p>
<pre>Version 1.0.53 (B339) [U.S.] 
UserShell: /bin/zsh
Bundle path: /Applications/Fig.app
Autocomplete: true
Settings.json: true
CLI installed: true
CLI tool path: /Users/lexnasser/.fig/bin/fig
Accessibility: true
Number of specs: 0
SSH Integration: false
Tmux Integration: false
Keybindings path: <none>
iTerm Integration: false 
Hyper Integration: false
VSCode Integration: false
Docker Integration: false
Symlinked dotfiles: false
Only insert on tab: false
Installation Script: true
PseudoTerminal Path: 
SecureKeyboardInput: false
SecureKeyboardProcess: <none>
Current active process: ??? (???) - ???
Current working directory: ???
Current window identifier: ???</pre>
</p>
</details>

fedeci:
Fixes #842

### DIFF
--- a/src/grep.ts
+++ b/src/grep.ts
@@ -257,6 +257,7 @@ const completionSpec: Fig.Spec = {
       name: "--exclude-dir",
       description:
         "If -R is specified, only directories matching the given filename pattern are searched.  Note that --exclude-dir patterns take priority over --include-dir patterns",
+      isRepeatable: true,
       args: {
         name: "dir",
         template: "folders",


lexnasser:
Thanks for fixing this @fedeci !

PR #848, opened 2022-01-18
feat(aws/apprunner): add spec

nmassardot:
Feat

It doesn't support any of `aws apprunner` commands

Now `aws apprunner create-connection` and all of its options are supported

**Additional info:**

### DIFF
--- a/src/aws/apprunner.ts
+++ b/src/aws/apprunner.ts
@@ -5,6 +5,16 @@ const completionSpec: Fig.Spec = {
     {
       name: "create-connection",
       description: "Create an app runner connection resource",
+      options: [
+        {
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
We should add this subcommand to https://github.com/withfig/autocomplete/blob/master/src/aws.ts
using a `loadSpec` directive to make it available when running `aws apprunner`.

brendanfalk:
Hi all - closing this as we are now part of AWS, we are likely going to do a much tighter integration with the AWS CLI. Stay tuned.

This PR currently doesn't have the correct `loadSpec` functionality and is missing a lot of subcommands and options

PR #851, opened 2022-01-18
feat/Add options to git diff

### LINKED ISSUE #827: [git] `git diff` missing --stat flag
### Sanity checks

git

### Issue Details

This would've been a time when Fig could've shined, as I don't use this option often enough to remember it offhand.

For `git diff` there is no completion for the `--stat` flag. `git diff --stat` shows line counts for changed files. Note this can optionally take parameters, according to the help output: `--stat[=<width>[,<name-width>[,<count>]]]`.  Relatedly, there are also `--numstat` and `--shortstat` options.

In fact, it seems that the only two completions offered for `git diff` options are `--cached` and `--staged`. `git diff --help` lists many many more. I'm not sure if you prefer broad "this completion is incomplete" issues or specific issues for missing completions based on demand. I'll go with the latter, and stick to reporting just the missing `--stat` option, as that's what I was reaching for right now.

khaledosama999:
Resolves #827

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -2593,6 +2593,29 @@ const completionSpec: Fig.Spec = {
           name: "--cached",
           description: "Show difference between staged changes and last commit",
         },
+        {
+          name: "--help",
+          description: "Shows different options",
+        },
+        {
+          name: "--numstat",
+          description:
+            "Shows number of added and deleted lines in decimal notation",
+        },
+        {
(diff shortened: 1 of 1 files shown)


mschrage (approved):
Thanks so much for this contribution @khaledosama999 ❤️

fedeci (approved):
Thanks!

PR #852, opened 2022-01-19
feat: update mkdir completion

qazwini:
Adds folder completion to the `mkdir` command.

Currently, if I’m at ~ and I’m trying to make a new folder in ~/Desktop/work/, I have to manually type in `mkdir Desktop/work/<folder name>`. There is no autocomplete for “Desktop/work/”.

If I am at ~ and I’m trying to make a new folder in ~/Desktop/work/, I will see my list of folders in the cwd as soon as I type `mkdir `. Then I will be able to autocomplete Desktop/ and same for work/.

**Additional info:**

### DIFF
--- a/src/mkdir.ts
+++ b/src/mkdir.ts
@@ -3,6 +3,7 @@ const completionSpec: Fig.Spec = {
   description: "Make directories",
   args: {
     name: "directory name",
+    template: "folders",
   },
   options: [
     {
--- a/src/mkdir.ts
+++ b/src/mkdir.ts
@@ -4,6 +4,7 @@ const completionSpec: Fig.Spec = {
   args: {
     name: "directory name",
     template: "folders",
+    suggestCurrentToken: true,
   },
   options: [
     {


mschrage:
@qazwini we definitely should provide some suggestions here, however I think just using the `folder` template could create some UX issues.

For instance, if a user is trying to create a new directory called "App" in a folder that already contains another folder called "Application":

Screen Shot 2022-01-18 at 11 15 30 PM

The user would need to hide the autocomplete window before being able to press enter.

What we really want is this behavior. 
Screen Shot 2022-01-18 at 11 17 05 PM

My naive attempt was to add another generator that just outputs the current token. 
```ts
 generators: [{
        template: [ "folders" ],
      },
      // ensure that current token is always a valid suggestion
      {
        custom:  async (tokens) => {
          return [ { name: tokens[tokens.length - 1] } ]
        }
      }
    ]
```

However, this results in strange behavior like this:
Screen Shot 2022-01-18 at 11 21 13 PM
Screen Shot 2022-01-18 at 11 21 25 PM

Any thoughts @fedeci or @SeparateRecords ?

fedeci:
IMO the only solution without touching the AE is to create a new custom generator that simulates the behaviour of the folders template but is able to handle that edge case @mschrage.

mschrage:
@fedeci this feels like something that should be handled internally in the autocomplete-engine. I think we should add a new field on `Arg` that triggers this behavior?

fedeci:
Will work on that in these days!

qazwini:
I also just noticed the same issue exists for `touch`. For, `dir/dir/newfile`, `dir/dir/` will not autocomplete.

fedeci:
@qazwini @a7med-mahmoud See https://github.com/withfig/autocomplete-tools/pull/36

src/mkdir.ts
```diff
@@ -3,6 +3,7 @@ const completionSpec: Fig.Spec = {
   description: "Make directories",
   args: {
     name: "directory name",
+    template: "folders",
```
fedeci, line 6:
```suggestion
    template: "folders",
    suggestCurrentToken: true,
```

fedeci (changes requested):

PR #856, opened 2022-01-20
Add `clang` support

a7med-mahmoud:
**What kind of change does this PR introduce? (Bugfix, feature, docs update, ...)**

Feature

No support for `clang`, the C/C++ compiler.

Added support for `clang` and its commands/options.

**Additional info:**

TBH I didn't write the whole spec, most of it is generated by a parser I wrote to parse the `clang --help` command and produce a spec out of it, then I fixed some issues in the spec and improved it a bit :D if there're any issues make sure to let me know!

### DIFF
--- a/src/clang.ts
+++ b/src/clang.ts
@@ -1,0 +1,11 @@
+const completionSpec: Fig.Spec = {
+  name: "clang",
+  description: "Clang LLVM compiler",
+  args: {
+    name: "file",
+    isVariadic: true,
+    template: "filepaths",
+  },
+};
+
+export default completionSpec;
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
There are so many non posix compliant options here, we need to specify it in the `parserOptions` ([see](https://fig.io/docs/reference/subcommand#parserdirectives))

a7med-mahmoud:
Thanks for the note @fedeci! I've fixed that now.

fedeci (approved):
Thanks! 🚀

PR #857, opened 2022-01-20
feat: add traceroute completion spec

mashanz:
Feature: add support for the `traceroute` command line tool.

There is no current behavior.

New behavior is support for the `traceroute` command line tool.

**Additional info:**
Please let me know if there are any changes desired by the maintainers.

### DIFF
--- a/src/traceroute.ts
+++ b/src/traceroute.ts
@@ -1,0 +1,130 @@
+const completionSpec: Fig.Spec = {
+  name: "traceroute",
+  description: "Print the route packets take to network host",
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show help for traceroute",
+    },
+    {
+      name: "-a",
+      description: "Turn on AS# lookups for each hop encountered",
+    },
+    {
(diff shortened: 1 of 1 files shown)


src/traceroute.ts
```diff
@@ -0,0 +1,130 @@
...
+    {
+      name: ["--help", "-h"],
+      description: "Show help for traceroute",
+    },
+    {
+      name: "-a",
+      description: "Turn on AS# lookups for each hop encountered",
+    },
+    {
+      name: ["-A", "as_server"],
```
fedeci, line 14: I think `as_server` actually is the argument of the option and not another way of calling it.
```suggestion
      name: "-A",
      args: { name: "as_server" },
```
It's the same for other options of this spec, could you fix those?

fedeci (changes requested):

src/traceroute.ts
mashanz: Hi @fedeci thanks for the review, just updated the arguments.

fedeci (approved):
Thanks!

PR #859, opened 2022-01-21
Set `valet link` argument as optional

SeparateRecords:
In response to an email.

The `valet link` hostname argument is optional: https://laravel.com/docs/8.x/valet#the-link-command


fedeci (approved):
👍🏻

PR #861, opened 2022-01-21
[fastly] Update `fastly` spec to 1.6.0

kailan:
Since the last spec update, we've added a `--json` flag to many commands that allows the CLI to print JSON output rather than human-readable data.


fedeci (approved):
Thanks!

PR #867, opened 2022-01-22
`sqlite3` support

a7med-mahmoud:

feature

no support for sqlite3

added support for sqlite3

### DIFF
--- a/src/sqlite3.ts
+++ b/src/sqlite3.ts
@@ -55,10 +55,7 @@ const completionSpec: Fig.Spec = {
       description: "Read/process named file",
       args: {
         name: "FILENAME",
-        generators: [
-          dbFilepath,
-          { custom: async (tokens) => [{ name: tokens[tokens.length - 1] }] },
-        ],
+        generators: dbFilepath,
       },
     },
     {
--- a/src/sqlite3.ts
+++ b/src/sqlite3.ts
@@ -1,0 +1,181 @@
+import { filepaths } from "@fig/autocomplete-generators";
+
+const dbFilepath = filepaths({
+  extensions: ["sql", "sqlite", "sqlite3", "db"],
+});
+
+const completionSpec: Fig.Spec = {
+  name: "sqlite3",
+  description: "A command line interface for SQLite version 3",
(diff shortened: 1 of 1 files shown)


import-brain (approved):

src/sqlite3.ts
```diff
@@ -0,0 +1,181 @@
...
+
+const dbFilepath = filepaths({
+  extensions: ["sql", "sqlite", "sqlite3", "db"],
+});
+
+const completionSpec: Fig.Spec = {
+  name: "sqlite3",
+  description: "A command line interface for SQLite version 3",
+  parserDirectives: {
+    flagsArePosixNoncompliant: true,
```
mschrage, line 11: 🙏

src/sqlite3.ts
```diff
@@ -0,0 +1,181 @@
...
+      description: "Print commands before execution",
+    },
+    {
+      name: "-init",
+      description: "Read/process named file",
+      args: {
+        name: "FILENAME",
+        generators: [
+          dbFilepath,
+          { custom: async (tokens) => [{ name: tokens[tokens.length - 1] }] },
```
mschrage, line 60: Is this signaling that any value the user types should be immediately executable (eg. show the red arrow)?

If so, we should potentially provide some flag for this at the `arg` level. cc @sullivan-sean @fedeci 

The new field could be something like `promptToExecuteAnyArgument: true` or `suggestExecutionForAnyArgument: true`? This would also be useful for the arguments to commands like `mkdir` or `touch`. See #852

a7med-mahmoud: Yes, I actually used the snippet you provided on that issue, although providing a flag for it would be very useful :)

a7med-mahmoud: I did some more research and found out that this command doesn't really "init" a new db file although the name suggests that, so I removed the "any arg" trick from the spec.

a7med-mahmoud:
Is there anything else needed for this PR to be merged?

fedeci (approved):
It looks good to me, thanks!

PR #884, opened 2022-01-27
Readme: Fig is a brew cask formula

abumalick:
We must use `--cask` to install fig

**Additional info:**


fedeci (approved):
Thanks!

PR #886, opened 2022-01-27
feat(omz): add reload option

kxxoling:
**What kind of change does this PR introduce? (Feature)**

image

Add a new `reload` option right after `pr`


fedeci (approved):
Thanks!

Issue #858, opened 2022-01-20
[vim] Add support for vim command line options


pedropombeiro:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

vim

### Which statement makes the most sense?

This completion spec does not exist

### Issue Details

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Nov 13 2021 05:04:57)

Usage: vim [arguments] [file ..]       edit specified file(s)
   or: vim [arguments] -               read text from stdin
   or: vim [arguments] -t tag          edit file where tag is defined
   or: vim [arguments] -q [errorfile]  edit file with first error

Arguments:
   --			Only file names after this
   -v			Vi mode (like "vi")
   -e			Ex mode (like "ex")
   -E			Improved Ex mode
   -s			Silent (batch) mode (only for "ex")
   -d			Diff mode (like "vimdiff")
   -y			Easy mode (like "evim", modeless)
   -R			Readonly mode (like "view")
   -Z			Restricted mode (like "rvim")
   -m			Modifications (writing files) not allowed
   -M			Modifications in text not allowed
   -b			Binary mode
   -l			Lisp mode
   -C			Compatible with Vi: 'compatible'
   -N			Not fully Vi compatible: 'nocompatible'
   -V[N][fname]		Be verbose [level N] [log messages to fname]
   -D			Debugging mode
   -n			No swap file, use memory only
   -r			List swap files and exit
   -r (with file name)	Recover crashed session
   -L			Same as -r
   -T <terminal>	Set terminal type to <terminal>
   --not-a-term		Skip warning for input/output not being a terminal
   --ttyfail		Exit if input or output is not a terminal
   -u <vimrc>		Use <vimrc> instead of any .vimrc
   --noplugin		Don't load plugin scripts
   -p[N]		Open N tab pages (default: one for each file)
   -o[N]		Open N windows (default: one for each file)
   -O[N]		Like -o but split vertically
   +			Start at end of file
   +<lnum>		Start at line <lnum>
[...]

PR #888, opened 2022-01-28
Add support for vim command line options, for #858

Yohannfra:
Add vim command line options 
No command line options
Command line options

### DIFF
--- a/src/vim.ts
+++ b/src/vim.ts
@@ -5,6 +5,177 @@ const completionSpec: Fig.Spec = {
   args: {
     template: "filepaths",
   },
+  options: [
+      {
+          name: "-v",
+          description: "Vi mode (like 'vi')",
+      },
+      {
+          name: "-e",
+          description: "Ex mode (like 'ex')",
+      },
+      {
+          name: "-E",
+          description: "Improved Ex mode",
+      },
+      {
+          name: "-s",
+          description: "Silent (batch) mode (only for 'ex')",
+          dependsOn: ["-e"],
+      },
+      {
+          name: "-d",
+          description: "Diff mode (like 'vimdiff')",
+      },
+      {
+          name: "-y",
+          description: "Easy mode (like 'evim', modeless)",
+      },
+      {
+          name: "-R",
+          description: "Readonly mode (like 'view')",
+      },
+      {
+          name: "-Z",
+          description: "Restricted mode (like 'rvim')",
+      },
+      {
+          name: "-m",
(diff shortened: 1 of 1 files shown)


src/vim.ts
```diff
@@ -5,6 +5,260 @@ const completionSpec: Fig.Spec = {
...
+    },
+    {
+      name: "-r (with file)",
+      description: "Recover crashed session",
+      args: {
+        name: "filename",
+        isOptional: false,
+        template: "filepaths",
+      },
+    },
```
mschrage, line 101: The `name` field must only contain the flag name and nothing else. Please combine the two `-r` flags.

src/vim.ts
```diff
@@ -5,6 +5,260 @@ const completionSpec: Fig.Spec = {
...
+        name: "command",
+        isOptional: false,
+      },
+    },
+    {
+      name: "-c",
+      description: "Execute <command> after loading the first file",
+      args: {
+        name: "command",
+        isOptional: false,
```
mschrage, line 189: Add `isCommand: true` here so that Fig will continue to provide suggestions :)

src/vim.ts
```diff
@@ -5,6 +5,260 @@ const completionSpec: Fig.Spec = {
...
+        template: "filepaths",
+      },
+    },
+    {
+      name: "-i",
+      description: "Use <viminfo> instead of .viminfo",
+      args: {
+        name: "viminfo",
+        isOptional: false,
```
mschrage, line 246: Should this have `template: "filepaths"`?

mschrage (changes requested):
Thanks so much for fleshing out the `vim` spec @Yohannfra. There are a few minor changes before we can get this merged in.

src/vim.ts
Yohannfra: yes thanks

src/vim.ts
```diff
@@ -5,6 +5,259 @@ const completionSpec: Fig.Spec = {
...
+      name: "-n",
+      description: "No swap file, use memory only",
+    },
+    {
+      name: "-r",
+      description:
+        "Recover crashed session if filename is specified, otherwise list swap files and exit",
+      args: {
+        name: "filename",
+        isOptional: false,
```
mschrage, line 95: `isOptional` should be `true` here, right?

Yohannfra: yes oops! thanks you

mschrage (approved):

PR #899, opened 2022-02-02
feat(okteto): add generators for `ns` and `ctx`

RinkiyaKeDad:
Signed-off-by: Arsh Sharma <arshsharma461@gmail.com>

Adds generators for `okteto ns use/delete` and `okteto ctx use/delete`

https://github.com/okteto/okteto/issues/1998

https://user-images.githubusercontent.com/56963264/152126700-1bd8d582-044f-4fe4-b5c7-4b84a72f1373.mov

**Additional info:**

A couple of questions:

- where do we upload the Okteto logo so that we can use `fig://icon?type=okteto`?
- as seen in the demo, there is a bit of delay showing the namespaces when doing `okteto ns use/delete` - is there a way to do some sort of caching to speed this up?
- is it possible to show the generated results in fig at the top of the list, that is, above the flag suggestions?

Thanks in advance!

cc @rberrelleza

### DIFF
--- a/src/okteto.ts
+++ b/src/okteto.ts
@@ -1,8 +1,7 @@
 const contexts: Fig.Generator = {
   script: "okteto context list",
   postProcess: (output) => {
-    return output.split("\n").map((context, ind) => {
-      if (ind == 0) return;
+    return output.split("\n").slice(1).map((context, ind) => {
       context = context.split(" ")[0];
       return {
         name: context.replace("*", "").trim(),
--- a/src/okteto.ts
+++ b/src/okteto.ts
@@ -1,5 +1,8 @@
 const contexts: Fig.Generator = {
   script: "okteto context list",
+  cache: {
+    ttl: 1000 * 60 * 30, // 30 minutes
+  },
   postProcess: (output) => {
     return output.split("\n").slice(1).map((context, ind) => {
       context = context.split(" ")[0];
@@ -14,6 +17,9 @@ const contexts: Fig.Generator = {
 
 const namespaces: Fig.Generator = {
   script: "okteto namespace list",
+  cache: {
+    ttl: 1000 * 60 * 30, // 30 minutes
+  },
   postProcess: (output) => {
     return output.split("\n").slice(1).map((namespace, ind) => {
       namespace = namespace.split(" ")[0];
--- a/src/okteto.ts
+++ b/src/okteto.ts
@@ -15,8 +15,7 @@ const contexts: Fig.Generator = {
 const namespaces: Fig.Generator = {
   script: "okteto namespace list",
   postProcess: (output) => {
-    return output.split("\n").map((namespace, ind) => {
-      if (ind == 0) return;
+    return output.split("\n").slice(1).map((namespace, ind) => {
       namespace = namespace.split(" ")[0];
       return {
         name: namespace.replace("*", "").trim(),
--- a/src/okteto.ts
+++ b/src/okteto.ts
@@ -4,14 +4,17 @@ const contexts: Fig.Generator = {
     ttl: 1000 * 60 * 30, // 30 minutes
   },
   postProcess: (output) => {
-    return output.split("\n").slice(1).map((context, ind) => {
-      context = context.split(" ")[0];
-      return {
-        name: context.replace("*", "").trim(),
-        description: "Context",
-        icon: "fig://icon?type=okteto",
-      };
-    });
+    return output
+      .split("\n")
+      .slice(1)
+      .map((context, ind) => {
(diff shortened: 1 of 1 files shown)


src/okteto.ts
```diff
@@ -1,3 +1,33 @@
+const contexts: Fig.Generator = {
+  script: "okteto context list",
```
fedeci, line 2: This is how we can add cache to a generator
```suggestion
  script: "okteto context list",
  cache: {
    ttl: 1000 * 60 * 60 * 24 * 2 // 2 days
  },
```

fedeci (changes requested):
/cc @mschrage there is an icon request https://www.okteto.com/favicon-32x32.png?v=f033e5af6700c9cb7abf797278bfe155

src/okteto.ts
RinkiyaKeDad: Thanks @fedeci! If this `ttl` is 2 days, does this mean that even if new `contexts` are available, they won't be shown until the `ttl` expires, and we request for the `contexts` again?

fedeci: That's what we do!

RinkiyaKeDad: @rberrelleza do we want this? This would mean new `contexts` don't show up immediately. We could reduce the `ttl` to an hour to make it more convenient, but I could still see this being problematic in cases where a newly created `context` won't load up in the suggestions immediately.

rberrelleza: @RinkiyaKeDad I think that we should have a cache. 30 minutes sounds reasonable as a starting point. Once cache invalidation is available, we can rethink this approach.

src/okteto.ts
```diff
@@ -1,3 +1,37 @@
...
+  },
+  postProcess: (output) => {
+    return output.split("\n").slice(1).map((context, ind) => {
+      context = context.split(" ")[0];
+      return {
+        name: context.replace("*", "").trim(),
+        description: "Context",
+        icon: "fig://icon?type=okteto",
+      };
+    });
```
fedeci, line 14:
```suggestion
    return output
      .split("\n")
      .slice(1)
      .map((context, ind) => {
        context = context.split(" ")[0];
        return {
          name: context.replace("*", "").trim(),
          description: "Context",
          icon: "fig://icon?type=okteto",
        };
      });
```

src/okteto.ts
```diff
@@ -1,3 +1,37 @@
...
+  },
+  postProcess: (output) => {
+    return output.split("\n").slice(1).map((namespace, ind) => {
+      namespace = namespace.split(" ")[0];
+      return {
+        name: namespace.replace("*", "").trim(),
+        description: "Namespace",
+        icon: "fig://icon?type=okteto",
+      };
+    });
```
fedeci, line 31:
```suggestion
    return output
      .split("\n")
      .slice(1)
      .map((namespace, ind) => {
        namespace = namespace.split(" ")[0];
        return {
          name: namespace.replace("*", "").trim(),
          description: "Namespace",
          icon: "fig://icon?type=okteto",
        };
      });
```

fedeci (approved):
Thanks!

RinkiyaKeDad:
Thanks a lot for making the changes and getting the PR merged @fedeci! I also wanted to ask if there is a way we can automate sending PRs to this repo whenever we release a new version of our CLI?

fedeci:
That's definitely something I have to start working on! @ryanccn implemented something and I will probably take some inspiration from his work.

RinkiyaKeDad:
Awesome, thanks! 

Do you know how long will it take for the changes in this PR to reflect? I updated fig to the latest version but running `okteto ns delete` still doesn't load the autocomplete suggestions based on generators for me. And has the Okteto logo been uploaded?

fedeci:
Does it provide suggestions for okteto subcommands? Changes should be live as soon as the PR gets merged.
I will add the icon soon.

RinkiyaKeDad:
Hey @fedeci! Apologies for the late reply; the suggestions are now working. Could you please add the icon soon? We need to record a demo for a blog post we're writing 🙏 

Screenshot 2022-03-14 at 5 13 24 PM

fedeci:
@RinkiyaKeDad see https://github.com/withfig/autocomplete/commit/9ecb1538ffd0bc226bdd2883b97337da7abbc371. This won't be available immediately but it will in the next fig release.

RinkiyaKeDad:
Thanks @fedeci! Do you know when's the next release planned?

fedeci:
@mschrage knows that better than me!

mschrage:
Later today @RinkiyaKeDad

PR #902, opened 2022-02-03
Add attribution for ffmpeg spec

ryanccn:

docs update

### DIFF
--- a/src/ffmpeg.ts
+++ b/src/ffmpeg.ts
@@ -1,5 +1,3 @@
-/* eslint-disable @withfig/fig-linter/no-useless-arrays */
-
 /*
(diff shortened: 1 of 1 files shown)


src/ffmpeg.ts
```diff
@@ -1,5 +1,10 @@
 /* eslint-disable @withfig/fig-linter/no-useless-arrays */
```
fedeci, line 1: I haven't seen this one :) You can safely remove it because `npm run lint:fix` should fix the errors automatically so it can be run in the CI.

fedeci (approved):
Thanks!

src/ffmpeg.ts
ryanccn: Fixed

PR #904, opened 2022-02-03
feat(velociraptor): add completion spec

ryanccn:

feature

added a completion spec for [Velociraptor](https://velociraptor.run), the npm-style script runner for Deno

**Additional info:**

### DIFF
--- a/src/vr.ts
+++ b/src/vr.ts
@@ -1,44 +1,33 @@
-// const SCRIPT_KEYWORD = "    • ";
-// const scriptGenerator: Fig.Generator = {
-//   script: "vr",
-//   postProcess: (out) => {
-//     let suggestions: Fig.Suggestion[] = [];
+const SCRIPT_KEYWORD = "    • ";
 
-//     try {
-//       const lines = out
-//         .split("\n")
-//         .filter(Boolean)
-//         .map((k) => {
-//           return k.replace(/\u001\[[0-9;]*m/g, "");
-//         });
+const scriptGenerator: Fig.Generator = {
+  script: "NO_COLOR=1 vr",
+  postProcess: (out) => {
+    let suggestions: Fig.Suggestion[] = [];
 
-//       for (const idx_ in lines) {
-//         const idx = (idx_ as unknown) as number;
+    try {
+      const lines = out.split("\n").filter(Boolean);
 
-//         if (idx >= lines.length - 2) continue;
+      for (const line of lines) {
+        if (!line.startsWith(SCRIPT_KEYWORD)) continue;
 
-//         const line = lines[idx];
+        const idx = lines.findIndex((k) => k === line);
 
-//         if (line.search(SCRIPT_KEYWORD) === -1) continue;
+        suggestions = [
+          ...suggestions,
+          {
+            name: line.replace(SCRIPT_KEYWORD, ""),
+            description: lines[idx + 1] ? lines[idx + 1].trim() : null,
+          },
+        ];
+      }
+    } catch (error) {
+      console.error(error);
+    }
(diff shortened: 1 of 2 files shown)


src/vr.ts
```diff
@@ -0,0 +1,58 @@
...
+        isVariadic: true,
+      },
+      options: [{ name: ["-h", "--help"], description: "Show help" }],
+    },
+    {
+      name: "upgrade",
+      description:
+        "Upgrade Velociraptor to the latest version or to a specific one",
+      args: {
+        name: "version",
```
fedeci, line 28: If you run `vr upgrade` without args does it install the latest version? In that case we should mark this arg as optional

src/vr.ts
```diff
@@ -0,0 +1,58 @@
...
+  name: "vr",
+  description: "The npm-style script runner for Deno",
+
+  subcommands: [
+    {
+      name: "run",
+      description: "Run a script",
+      args: {
+        name: "script",
+      },
```
fedeci, line 11: Can we create a generator for this one?

ryanccn:
It turns out that parsing `vr`'s output is incredibly hard because I can't force it to not use colors, and it doesn't have any program-friendly output (such as JSON output)...﻿

src/vr.ts
ryanccn: This is done

fedeci:
@ryanccn do you think the `velociraptor` maintainers are willing to add a --json option to the command so that we can get a nicer and more parsable output?

ryanccn:
That's great! I guess I'll wait for the upstream issue to be resolved 😅

fedeci:
We can merge it without the generators if you prefer! We can add those in a second time.

ryanccn:
OK, sure! I'm not sure why the linter is failing on `cat` though 🤔

fedeci:
Rebasing on top of main should fix!

SeparateRecords:
BTW, velociraptor respects NO_COLOR - try running any generators with `NO_COLOR=1 vr [...]` 🙂

ryanccn:
Voilà!

ryanccn:
(It looks for the bullet points, lol)

src/vr.ts
```diff
@@ -0,0 +1,94 @@
...
+          },
+        },
+      ],
+    },
+  ],
+
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show help for Velociraptor",
```
fedeci, line 80: We can add this prop to `--help` and remove his repetitions in the subcommands
```suggestion
      description: "Show help for Velociraptor",
      isPersistent: true,
```

src/vr.ts
```diff
@@ -0,0 +1,94 @@
...
+        const idx = lines.findIndex((k) => k === line);
+
+        suggestions = [
+          ...suggestions,
+          {
+            name: line.replace(SCRIPT_KEYWORD, ""),
+            description: lines[idx + 1] ? lines[idx + 1].trim() : null,
+          },
+        ];
+      }
```
fedeci, line 23: I prefer a C style loop instead of the `.findIndex`.
```suggestion
      for (let i = 0; i < lines.length; i++) {
        const line = lines[i];
        if (!line.startsWith(SCRIPT_KEYWORD)) continue;

        suggestions.push({
          name: line.replace(SCRIPT_KEYWORD, ""),
          ...(lines[i + 1] && { description: lines[i + 1].trim() }),
        }),
      }
```

fedeci (changes requested):

fedeci (approved):
Thanks!

PR #906, opened 2022-02-03
feat(cargo): Add all remaining subcommands

grant0417:
This adds all the remaining subcommands to cargo, thankfully cargo is based on clap 3 which I just recently improved.

What is missing still from cargo spec:
* All the current generators could be improved, most are slow and not get functionally
* Still missing tons of plugins (cargo-outdated, cargo-edit, etc.)


fedeci (approved):
Thanks!

PR #909, opened 2022-02-04
feat(fastlane): add fastlane spec

hangai247:
feature

[Fastlane](https://fastlane.tools/)  Automation tool for beta deployments and releases for iOS and Android apps.
**Additional info:**


hangai247:
> Hello @hangai247, thank you very much for creating a Pull Request! Here is a small checklist to get this PR merged as quickly as possible:
> 
> * [x ]  Do all subcommands / options which take arguments include the `args` property (`args: {}`)?
> * [ ]  Are all options modular? E.g. `-a` `-u` `-x` instead of `-aux`
> * [ ]  Have all other checks passed?
> 
> Please add a 👍 as a reaction to this comment to show that you read this.

src/fastlane.ts
```diff
@@ -0,0 +1,1188 @@
+const actionGenerators: Fig.Generator = {
+  custom: async (context) => {
+    const searchTerm = context[context.length - 1];
+    return [
```
fedeci, line 4: Here we want each action to have the subcommands below, right?
```ts
  interface ScriptsRow {
    name: string
    deprecated: boolean,
    description: string
  }
  ...
  script: "fastlane actions",
  postProcess: async (out) => {
    // we need to perform some kind of parsing to get the rows and columns of the table.
   rows.map((r: ScriptsRow[]) => {
     return {
       name: r.name,
       ...( r.deprecated && { deprecated: r.deprecated }),
       description: r.description,
       subcommands: [] // The subcommands listed below
     }
   })
  }
```

fedeci (changes requested):
Thank you, I left a couple of comments, let me know if you need help!

src/fastlane.ts
hangai247: @fedeci Thank you for your review. I will fix all of the above implicit configs. :)

About the action generator. Actually, I was trying to do the same. But unfortunately, fastlane actions and plugins command response is returning some weird markdown txt response, which I couldn't figure out easily.

fedeci: Yeah, that's a super weird output. We may try to propose the fastlane team to add a --json flag to the command to get a nicer output WDYT?

hangai247: Sounds very good. Thank you very much. I will remove this part for now and I hope we can incrementally improve it in the future. 😄

hangai247 (commented):
Approved

fedeci (approved):
Thanks!

Issue #901, opened 2022-02-03
[git]


SeparateRecords:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

git

### Which statement makes the most sense?

This completion spec isn't complete

### Issue Details

`git diff` allows you to compare more than just one commit or file.

Quick example of wrong behaviour: `git diff HEAD --` will trigger suggestions. This is a valid command. As is `git diff HEAD HEAD~3`, but again no suggestions.

bolshoytoster:
@SeparateRecords I've opened a [PR](https://github.com/withfig/autocomplete/pull/911) that should fix this.

PR #911, opened 2022-02-05
feat: improve git diff

bolshoytoster:
Bug fix
#901
- Autocomplete ignores '--' in `git diff`.

- Autocomplete only suggests one commit when you can provide as many as 
you want.

- Autocomplete recognises '--' and provides `filepath` suggestions.

- `git diff` now uses `isVariadic` to suggest many arguments.


fedeci (approved):
Thanks!

fedeci:
Could you fix the lint errors?

PR #916, opened 2022-02-07
feat(black): add completion spec

msonnberger:
feature

No autocomplete for Black

Add autocomplete spec for [Black](https://github.com/psf/black) code formatter

### DIFF
--- a/src/black.ts
+++ b/src/black.ts
@@ -1,5 +1,9 @@
+// https://github.com/psf/black
 const blackVersions: Fig.Generator = {
   script: "gh release list --repo psf/black",
+  cache: {
+    ttl: 1000 * 60 * 60 * 24 * 2, // 2 days
+  },
   postProcess: (output) => {
     if (!output.includes("Latest")) {
       return [];
--- a/src/black.ts
+++ b/src/black.ts
@@ -20,7 +20,7 @@ const completionSpec: Fig.Spec = {
   description: "Python code formatter",
   args: {
     name: "file or directory",
-    template: ["filepaths", "folders"],
+    template: "filepaths",
     isVariadic: true,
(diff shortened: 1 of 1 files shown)


src/black.ts
```diff
@@ -0,0 +1,209 @@
+const blackVersions: Fig.Generator = {
```
fedeci, line 1:
```suggestion
// https://github.com/psf/black
const blackVersions: Fig.Generator = {
```

src/black.ts
```diff
@@ -0,0 +1,209 @@
...
+    });
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "black",
+  description: "Python code formatter",
+  args: {
+    name: "file or directory",
+    template: ["filepaths", "folders"],
```
fedeci, line 19: nit: folder is always added as part of `filepaths`.
```suggestion
    template: ["filepaths"],
```

src/black.ts
```diff
@@ -0,0 +1,209 @@
+const blackVersions: Fig.Generator = {
+  script: "gh release list --repo psf/black",
```
fedeci, line 2: We can add a couple days of cache here, WDYT?
```suggestion
  script: "gh release list --repo psf/black",
  cache: {
    ttl: 1000 * 60 * 60 * 24 * 2, // 2 days
  },
```

fedeci (changes requested):
This looks really good, I left a couple comments to resolve.

src/black.ts
msonnberger: sure, that makes sense! I've added your suggestions.

fedeci (approved):
Thanks!

PR #917, opened 2022-02-07
feat: add twilio CLI

SeparateRecords:
**NOTE:** I'm intentionally leaving out the API subcommand for the initial PR because it's _huge_.

https://user-images.githubusercontent.com/52195359/152804755-d8d91192-dd6b-4252-87ba-e1f351d6d054.mp4

## Explanation for reviewers

Two necessary pieces of context:
1. The Twilio CLI has "subcommands" but they're all a single colon-separated token. I chose to model this in the spec as nested subcommands.
2. The goal is to build the subcommand token, which means we need to use a generator, but subcommands inherently can't have a "generator".

This spec relies on the behaviour that subcommands have precedence over optional arguments, and arguments can have generators.

The argument generator helps build the subcommand, and once the subcommand is complete, Fig provides the suggestions for args and options.

### DIFF
--- a/src/twilio.ts
+++ b/src/twilio.ts
@@ -23,10 +23,10 @@ function flattenSubcommands(
 
   // This subcommand has no children - join the name of all of its parent
   // commands with the separator and set that as the new name.
-  const name =
-    parents.map(({ name }) => name).join(separator) +
-    separator +
-    subcommand.name;
+  const name = parents
+    .map(({ name }) => name)
+    .concat(subcommand.name)
+    .join(separator);
 
   return [{ ...subcommand, name }];
 }
--- a/src/twilio.ts
+++ b/src/twilio.ts
@@ -1,0 +1,1470 @@
+/* eslint-disable @withfig/fig-linter/no-useless-insertvalue */
+
+/** A version of Fig.Subcommand with a predictable name prop */
+interface TwilioSubcommand extends Fig.Subcommand {
+  name: string;
+  // Bad documentation is a bug. Descriptions are required.
+  description: string;
+  subcommands?: TwilioSubcommand[];
+}
+
+/** Recursively flatten subcommands by joining them with a separator */
+function flattenSubcommands(
+  separator: string,
+  parents: TwilioSubcommand[],
+  subcommand: TwilioSubcommand
+): TwilioSubcommand[] {
+  // If this subcommand has children, flatten it further.
+  if (subcommand.subcommands && subcommand.subcommands.length > 0) {
+    return subcommand.subcommands.flatMap((child) =>
+      flattenSubcommands(separator, parents.concat(subcommand), child)
+    );
+  }
+
+  // This subcommand has no children - join the name of all of its parent
+  // commands with the separator and set that as the new name.
+  const name =
+    parents.map(({ name }) => name).join(separator) +
+    separator +
+    subcommand.name;
+
+  return [{ ...subcommand, name }];
+}
+
+/**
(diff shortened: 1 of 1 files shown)


SeparateRecords:
@hangai247 I'd love your review as well - is there anything glaringly obvious missing?

src/twilio.ts
```diff
@@ -0,0 +1,1470 @@
...
+};
+
+const autocompleteSubcommand: TwilioSubcommand = {
+  name: "autocomplete",
+  description: "Display autocomplete installation instructions",
+  args: {
+    name: "shell",
+    description: "Shell type",
+    suggestions: [
+      { name: "bash", icon: "fig://icon?type=string" },
```
fedeci, line 1409: These suggestions are not working :(

src/twilio.ts
```diff
@@ -0,0 +1,1470 @@
...
+const updateSubcommand: TwilioSubcommand = {
+  name: "update",
+  description: "Update the twilio CLI",
+  args: {
+    name: "channel",
+    isOptional: true,
+  },
+  options: [
+    {
+      name: "--from-local",
```
fedeci, line 1396: These neither :(

fedeci (commented):
I'm giving a look at some specific parts of the spec to find potential problems

fedeci:
Schermata 2022-02-07 alle 17 10 57
I'm seeing these are a result of the call to `flattenSubcommands` in `createSeparatedSubcommandTree`

src/twilio.ts
fedeci: I think that's because we treat colon-chained commands as arguments and the parser becomes crazy because it expects other chained args while we are adding a "normal" subcommand arg.
At the same time I don't know how we could fix this problem. Maybe `generateSpec` behaves better?

src/twilio.ts
```diff
@@ -0,0 +1,1470 @@
...
+};
+
+const silentOption: Fig.Option = {
+  name: "--silent",
+  description:
+    'Suppress output and logs. This is a shorthand for "-l none -o none"',
+};
+
+// These options are on subcommands, not on the top level twilio CLI
+const persistentOptions = [logOption, outputFormatOption, silentOption];
```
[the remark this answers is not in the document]
SeparateRecords: I thought so too but they're not on each su command and not on the root command

SeparateRecords:
Absolutely not! I know what's causing that, and fixing that will also fix the problem with the other suggestions not showing

mschrage:
This is wild @SeparateRecords! I'm super impressed that this is possible without making any modifications to the autocomplete engine.

fedeci:
Yeah this is mind blowing🤯

SeparateRecords:
Thank you both! It's wild that the autocomplete engine is robust enough to handle this... well frankly this abuse of it 😆

When I have the time, I'll try my hand at making something like a `subcommandBuilder` property to simplify this, since Twilio definitely isn't the only CLI that would benefit. Once I've got a concrete idea I'll get write up a proposal.

@fedeci the issue with suggestions not showing is fixed now if you'd like to take another look! 🙂

fedeci (approved):
It's perfect now🎉 Great job @SeparateRecords!
@mschrage this needs to be twitted IMHO

hangai247:
@SeparateRecords Excuse me for the late response here. The change and the algorithm you've written here to achieve Twilio CLI topic and sub commands is awesome. Thank you very much. It will help my work a lot. ;)

PR #922, opened 2022-02-09
Add `gh release` subcommands

ryanccn:

feature

Add `gh release` subcommands

**Additional info:**

### DIFF
--- a/src/gh.ts
+++ b/src/gh.ts
@@ -77,10 +77,28 @@ const ghGenerators: Record<string, Fig.Generator> = {
     },
   },
 
+  listRepo: {
+    script: "gh repo list --json nameWithOwner,description",
+    postProcess: (out) => {
+      const data: { nameWithOwner: string; description: string }[] = JSON.parse(
+        out
+      );
+
+      return data.map((k) => ({
+        name: k.nameWithOwner,
+        description: k.description,
+      }));
+    },
+  },
+
   remoteBranches: {
     script:
       "git --no-optional-locks branch -r --no-color --sort=-committerdate",
     postProcess: postProcessRemoteBranches,
+  },
+
+  remoteTags: {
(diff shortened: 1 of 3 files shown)


ryanccn:
Uh oh... how do I remove the commits from the other PR

src/gh.ts
```diff
@@ -77,11 +77,29 @@ const ghGenerators: Record<string, Fig.Generator> = {
...
+    postProcess: (out) => {
+      const data: { nameWithOwner: string; description: string }[] = JSON.parse(
+        out
+      );
+
+      return data.map((k) => ({
+        name: k.nameWithOwner,
+        description: k.description,
+      }));
+    },
```
fedeci, line 91:
```suggestion
    postProcess: (out) => {
      if (!out) return [];
    	try {
    	  const data: { nameWithOwner: string; description: string }[] = JSON.parse(
          out.trim()
        );

      	return data.map((k) => ({
        	name: k.nameWithOwner,
        	description: k.description,
      	}));
    	} catch {
    	  return [];
    	}
    },
```

fedeci (changes requested):

ryanccn:
If we don't use `try...catch`, will Fig crash?

fedeci:
If JSON.parse throws an error Fig may crash. I am not sure if we are handling errors from `postProcess` in the engine, but even in that case I prefer to see error handling here.

SeparateRecords:
@ryanccn I'd love to see this get merged! Are you still working on it? No worries if you need more time 🙂

ryanccn:
> @ryanccn I'd love to see this get merged! Are you still working on it? No worries if you need more time 🙂

I think the only problem right now is the commits from another PR being included in this one as well... I don't know how to fix that

PR #929, opened 2022-02-14
feat: update checkov completion spec

eurogig:
Checkov has new command-line options so those were added to the completion spec

There were a few missing elements for existing options which was fixed (eg. added filepath template to config-file)

### DIFF
--- a/src/checkov.ts
+++ b/src/checkov.ts
@@ -61,9 +61,9 @@ const completionSpec: Fig.Spec = {
       name: ["--output", "-o"],
       description:
         "Report output format. Add multiple outputs by using the flag multiple times (-o sarif -o cli)",
+      isRepeatable: true,
       args: {
         name: "FORMAT",
-        isVariadic: true,
         suggestions: [
           "cli",
           "cyclonedx",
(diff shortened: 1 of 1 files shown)


src/checkov.ts
```diff
@@ -28,6 +57,71 @@ const completionSpec: Fig.Spec = {
         template: "folders",
       },
     },
+    {
+      name: ["--output", "-o"],
+      description:
+        "Report output format. Add multiple outputs by using the flag multiple times (-o sarif -o cli)",
+      args: {
+        name: "FORMAT",
+        isVariadic: true,
```
fedeci, line 66: `isVariadic` allows for `--output cli json [...]`, not `--output json --output cli`. Should be fixed or the first example works correctly? In case it doesn't work we should set the `repeatable: true` property of the output option and remove the variadic one here.

src/checkov.ts
```diff
@@ -238,6 +341,23 @@ const completionSpec: Fig.Spec = {
...
+        suggestions: ["critical", "high", "medium", "low", "none"],
+      },
+    },
+    {
+      name: "--skip-cve-package",
+      description:
+        "Ignore specific open source package when SCA scanning for CVEs in package dependencies. Can be used multiple times to skip multiple packages",
+      args: {
+        name: "SKIP_CVE_PACKAGE",
+        isVariadic: true,
```
fedeci, line 358: Here too

fedeci (commented):
Thanks for the PR, I left a couple comments here and there!

src/checkov.ts
eurogig: Got it... misunderstood.  Thanks will fix.

src/checkov.ts
eurogig: and done

PR #931, opened 2022-02-16
feat: add list repositories generator

ShawnCockburn:

feature

no generation of repo inputs

generation of repo inputs

**Additional info:**

the npm test failed for the `mkdir.ts` file but I made no changes to this file, only gh.ts


SeparateRecords:
This looks great! Tested it out, love being able to see my repos right where I'm typing, feels like magic 🤩

Would it be possible to add searching for repos by typing `<username>/`? I think you could do this with `custom`. Not sure how much extra work this would be or if you already looked into this - we can do that in a follow up, if you'd prefer.

CleanShot 2022-02-17 at 11 21 26@2x

Only feedback I have on your code is the inline interfaces - this is purely a personal preference, I just like to keep generator functions shorter by moving interface / type declarations to the global scope. It's totally fine to keep them where they are though!

ShawnCockburn:
> This looks great! Tested it out, love being able to see my repos right where I'm typing, feels like magic 🤩
> 
> Would it be possible to add searching for repos by typing `<username>/`? I think you could do this with `custom`. Not sure how much extra work this would be or if you already looked into this - we can do that in a follow up, if you'd prefer.
> 
> CleanShot 2022-02-17 at 11 21 26@2x
> 
> Only feedback I have on your code is the inline interfaces - this is purely a personal preference, I just like to keep generator functions shorter by moving interface / type declarations to the global scope. It's totally fine to keep them where they are though!

Thanks mate!

I did try doing that but it would require multiple API calls _(unless there is a better way?)_ and I thought I would add it at a later point :) <sub>(soon)</sub>.

I'm more than happy to move the types to wherever but I thought they would be more semantic in the function itself as it's the only thing that uses them in the code; easier for future contributors.

mschrage:
> I did try doing that but it would require multiple API calls (unless there is a better way?) and I thought I would add it at a later point :) (soon).

@SeparateRecords @ShawnCockburn we can split this into multiple PRs. I think we can get this merged in if everything is working as is (even if it doesn't do everything).

mschrage (approved):

ShawnCockburn:
> > I did try doing that but it would require multiple API calls (unless there is a better way?) and I thought I would add it at a later point :) (soon).
> 
> @SeparateRecords @ShawnCockburn we can split this into multiple PRs. I think we can get this merged in if everything is working as is (even if it doesn't do everything).

yes, that sounds good. ill have a play this weekend.

@mschrage is there a way to run conditional scripts then process them?

fedeci:
`custom` generator is what you are looking for!

ShawnCockburn:
> `custom` generator is what you are looking for!

Thank you! ♥️

PR #947, opened 2022-02-18
feat: add `killall` spec

SeparateRecords:
Includes generators for process names and users. Both are not linux-compatible. Listing users uses dscl, process names checks for ".app"

CleanShot 2022-02-19 at 02 09 15@2x
CleanShot 2022-02-19 at 02 09 28@2x

Next up is updating deno to 1.19

### DIFF
--- a/src/killall.ts
+++ b/src/killall.ts
@@ -1,3 +1,5 @@
+// Linux incompatible
+
 const signals = [
   "hup",
   "int",
@@ -35,7 +37,7 @@ const signals = [
 
 const completionSpec: Fig.Spec = {
   name: "killall",
-  description: "",
+  description: "Kill processes by name",
   args: {
     name: "process_name",
     isVariadic: true,
@@ -55,7 +57,10 @@ const completionSpec: Fig.Spec = {
             return {
               name,
               description: path,
-              priority: !badChars.some((char) => nameChars.has(char)) ? 51 : 40,
+              priority:
+                !badChars.some((char) => nameChars.has(char)) && isApp
+                  ? 51
+                  : 40,
               icon: isApp
                 ? "fig://" + path.slice(0, appExtIndex + 4)
                 : "fig://icon?type=gear",
--- a/src/killall.ts
+++ b/src/killall.ts
@@ -1,0 +1,148 @@
+const signals = [
+  "hup",
+  "int",
+  "quit",
+  "ill",
+  "trap",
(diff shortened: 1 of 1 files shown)


src/killall.ts
```diff
@@ -0,0 +1,148 @@
...
+  "prof",
+  "winch",
+  "info",
+  "usr1",
+  "usr2",
+];
+
+const completionSpec: Fig.Spec = {
+  name: "killall",
+  description: "",
```
fedeci, line 38:
```suggestion
  description: "Used for killing any running process on the system based on a given name",
```

src/killall.ts
```diff
@@ -0,0 +1,148 @@
+const signals = [
```
fedeci, line 1: We should start adding comments at the top of the spec if they are or not Linux compatible: 
```suggestion
// Linux incompatible
const signals = [
```

src/killall.ts
```diff
@@ -0,0 +1,148 @@
...
+      name: "-d",
+      description: "Be verbose (dry run) and display number of user processes",
+    },
+    {
+      name: "-e",
+      description:
+        "Use the effective user ID instead of the real user ID for matching processes with -u",
+    },
+    {
+      name: "-help",
```
fedeci, line 77: This requires `parserOptions.flagsArePosixNonCompliant`

fedeci (approved):
🚀

src/killall.ts
SeparateRecords: 🤦

src/killall.ts
SeparateRecords: Oh weird, all seemed to be working for me without it. Some options can be combined (eg `-Iu` and some can't be (`-help`), this prevents combining options right?

No too big of a deal anyway! Happy to add this 😃 (when does anyone even use flags on `killall` anyway?)

SeparateRecords: Current behaviour is correct as far I can tell, is this actually a bug?

https://user-images.githubusercontent.com/52195359/154792100-546a11cb-3a0a-47aa-a72c-4c6cb0dd7cf4.mp4

fedeci: This should only affect `-help` or `-SIGx` but probably because they do not have any argument they are unaffected. Let's leave this as is for the moment.

fedeci (approved):

PR #949, opened 2022-02-19
feat: add `rush` spec

F3n67u:

feature

Add spec for rushjs https://rushjs.io/

**Additional info:**


src/rush.ts
```diff
@@ -0,0 +1,840 @@
...
+            "The position in the word to be completed. The default value is 0",
+        },
+      ],
+    },
+  ],
+  options: [
+    {
+      name: ["-h", "--help"],
+      description: "Show this help message and exit",
+    },
```
fedeci, line 832: Adding `isPersistent` will add this option to subcommands too. We can remove those.
```suggestion
    {
      name: ["-h", "--help"],
      description: "Show this help message and exit",
      isPersistent: true,
    },
```

fedeci (changes requested):
Just one small change and we are ready to merge! The spec looks awesome!

src/rush.ts
F3n67u: Thanks for the review. I also removed the redundant help option declaration for subcommands.

fedeci (approved):
Thanks!

PR #950, opened 2022-02-19
feat: add rmdir completion

qazwini:
Feature.

`rmdir` has no completion.

Adds completion to `rmdir`.

**Additional info:**
I initially created a new generator for this that filters out all of the non-empty folders, but using the folders template makes more sense. Using the initial method, we would not get completion for `path1` in `rmdir path1/path2/`.


fedeci (approved):
Thanks!

PR #955, opened 2022-02-19
chore(deps): bump `typescript` to `4.5.0`

F3n67u:
Changes:

1. upgrade typescript to 4.5.0
2. Use tilde instread of caret when specify typescript dependency, because upgrade typescript minor version will introduce breaking change

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "pretty-quick": "^3.1.0",
     "progress": "^2.0.3",
     "ts-node": "^9.1.1",
-    "typescript": "~4.5.0"
+    "typescript": "^4.5.5"
   },
   "dependencies": {
     "@types/glob": "^7.2.0",
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "pretty-quick": "^3.1.0",
     "progress": "^2.0.3",
     "ts-node": "^9.1.1",
-    "typescript": "^4.2.3"
+    "typescript": "~4.5.0"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


fedeci:
According to semver minor versions do not introduce breaking changes when upgrading, but when downgrading. 
I mean: the output of `1.1.x` cannot be used with `1.0.x`, but it can with `1.2.x`. 
For what concerns typescript we can safely use caret because new minor version will only introduce new syntaxes which do not break older code.

package.json
```diff
@@ -91,7 +91,7 @@
     "pretty-quick": "^3.1.0",
     "progress": "^2.0.3",
     "ts-node": "^9.1.1",
-    "typescript": "^4.2.3"
+    "typescript": "~4.5.0"
```
fedeci, line 94:
```suggestion
    "typescript": "^4.5.5"
```

F3n67u:
> 

every typescript minor version could possibly introduce breaking change because typescript does not follows semver. 

see https://github.com/microsoft/TypeScript/issues/14116 for discussion.

fedeci:
That's true, but referring to the simplicity of this codebase in particular I don't expect any breaking changesa. IMHO we should keep the caret.

F3n67u:
> That's true, but referring to the simplicity of this codebase in particular I don't expect any breaking changes. IMHO we should keep the caret.

ok. let's keep the caret

fedeci:
Thanks mate! We will lock to some version if something breaks with new typescript versions.

PR #956, opened 2022-02-19
feat(ansible-playbook): update generators to use `.yml` files instead of hostname pattern

RobertJndw:
Bug fix

Old spec incorrectly expects a hostname pattern 

**What is the new behavior?**
New spec autocompletes `.yml` file(s) according to the [official Ansible Docs for the CLI](https://docs.ansible.com/ansible/latest/cli/ansible-playbook.html)


fedeci (approved):
Thanks!

PR #957, opened 2022-02-20
feat(kill): add process names, fix `-s` generator

SeparateRecords:
CleanShot 2022-02-20 at 17 51 47@2x

CleanShot 2022-02-20 at 17 51 53@2x


fedeci (approved):
Thanks!

PR #958, opened 2022-02-20
feat(man): add generator and options

SeparateRecords:
### Suggestions from other sections
CleanShot 2022-02-21 at 01 08 31@2x

### Section 1 & 8 by default
CleanShot 2022-02-21 at 01 08 43@2x

### Options
CleanShot 2022-02-21 at 01 08 52@2x


fedeci (approved):
Thanks!

PR #963, opened 2022-02-22
feat(rush): add project suggestion

F3n67u:
add project suggestion support for rush spec

image

### DIFF
--- a/src/rush.ts
+++ b/src/rush.ts
@@ -14,12 +14,12 @@ const projectGenerator: Fig.Generator = {
     const suggestions = [];
 
     try {
-      if (out.trim() == "") {
+      if (!out) {
         return suggestions;
       }
 
       const rushConfigurationJson: IRushConfigurationJson = JSON.parse(
-        stripJsonComments(out)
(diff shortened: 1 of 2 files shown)


src/rush.ts
```diff
@@ -1,40 +1,75 @@
...
+    const suggestions = [];
+
+    try {
+      if (out.trim() == "") {
+        return suggestions;
+      }
+
+      const rushConfigurationJson: IRushConfigurationJson = JSON.parse(
+        stripJsonComments(out)
+      );
```
fedeci, line 23: `out.trim()` does not mutate the original string:
```suggestion
      if (!out) {
        return suggestions;
      }

      const rushConfigurationJson: IRushConfigurationJson = JSON.parse(
        stripJsonComments(out.trim())
      );
```

fedeci (changes requested):
A small nitpick but generally it looks good!

src/rush.ts
F3n67u: fixed! thanks for review

fedeci (approved):
Thanks!

PR #970, opened 2022-02-23
feat(tar): improve spec

### LINKED ISSUE #962: [tar] missing options
### Sanity checks

tar

### Issue Details

There's a lot missing from this spec, check out `man tar`

bolshoytoster:
Spec update

The spec is incomplete: Closes #962

**Additional Info**
There aren't currently any generators or any 'clusters' (e.g. tar **cfv** a.tar /etc) but I could ad them if wanted.

### DIFF
--- a/src/tar.ts
+++ b/src/tar.ts
@@ -1,44 +1,1418 @@
+const fileOptions: Fig.Option[] = [
+  {
+    name: ["f", "-f", "--file"],
+    description: "Use archive file or device ARCHIVE",
+    isRequired: true,
+    args: { name: "ARCHIVE" },
+  },
+  {
+    name: "--force-local",
+    description: "Archive file is local even if it has a colon",
+    dependsOn: ["f", "-f", "--file"],
+  },
+  {
+    name: ["F", "-F", "--info-script", "--new-volume-script"],
+    description: "Run  COMMAND  at the end of each tape",
+    args: { name: "COMMAND" },
+  },
+  {
+    name: ["L", "-L", "--tape-length"],
+    description: "Change tape after writing Nx1024 bytes",
+    args: { name: "N", suggestions: sizeSuffixes },
+  },
+  {
+    name: ["M", "-M", "--multi-volume"],
+    description: "Create/list/extract multi-volume archive",
+  },
+  {
+    name: "--rmt-command",
+    description: "Use  COMMAND instead of rmt when accessing remote archives",
+    args: { name: "COMMAND" },
+  },
+  {
+    name: "--rsh-command",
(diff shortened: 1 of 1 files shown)


fedeci (commented):
Thanks! My only concern is that this PR refers to GNU tar, but also BSD tar is available, we should differentiate the specs.
Any ideas @SeparateRecords @mschrage? The eventual solution may also be the same for different CLI tools with the same name.

bolshoytoster:
@fedeci there are also three different versions of BSD tar ([freeBSD tar](https://www.freebsd.org/cgi/man.cgi?query=tar), [netBSD tar](https://man.netbsd.org/tar.1) and [openBSD tar](https://man.netbsd.org/tar.1) (There's also a [busybox tar](https://man.netbsd.org/tar.1))).

fedeci:
Love it! Simplicity is always the key😆

mschrage:
@fedeci my sense is long term, we will use the same approach we do for [versioning CLI tools](https://fig.io/docs/handbook/versioning-completion-specs) (perhaps with some syntactic sugar on top).

A CLI spec will export all of the variants it supports.

When a spec is loaded by the autocomplete-engine, we will check if it has variants. If it does, run `getVariant` to determine which one to load.

We would add a list of variants to autocomplete-types, the `getVariant` function and also we'd modify the `completionSpec` function type definition. 

**autocomplete-types**
```ts
type Variant = "GNU" | "openBSD" | "freeBSD" | "netBSD" | "busybox" ...
```

**tar.ts**
```ts
export variants: Variant[] =  ["GNU", "busybox" ]

export const getVariant = async (executeShellCommand): Variant  => {   
  // execute shell commands on device to determine which variant to use
  const help = await executeShellCommand("tar --help")
  return help.includes("GNU") ? "GNU" : "busybox"
}

const completionSpec: Fig.Spec = (version: string?, variant: Variant?): Fig.Spec => {
   switch (variant) {
     case "GNU":
         return { loadSpec: "gnu/tar"}
     case "busybox"
         return { loadSpec: "busybox/tar"}
   }
};
```

mschrage (changes requested):
I am not sure if the autocomplete engine supports flags that are not prefixed with a hyphen? Other than that, everything looks great!

Any thoughts? @fedeci @sullivan-sean

fedeci:
I think `flagsArePosixNonCompliant` will be enough to handle this case.

bolshoytoster:
@fedeci sorry, I didn't see that in the docs. I can add that now.

SeparateRecords:
Is there something blocking this from being merged? Still stuff that needs to be done?

PR #971, opened 2022-02-23
feat(gh): add custom repo completion

ShawnCockburn:

feature && update

`gh` repo completion for users own repos but not others

autocomplete any public repos you want!

and an update to `gh auth`

**Additional info:**

literal magic..  🤯.

@mschrage suggested that custom did not need to be wrapped in `try..catch` and the [docs example](
https://github.com/withfig/autocomplete-tools/blob/aad28dc389b49155e8523bf589a17370a66f1f63/packages/autocomplete-types/index.d.ts#L1014-L1022)  does not have `try catch`.

@fedeci can you confirm if this is the case?


ShawnCockburn:
@mschrage thanks for explaining `custom()` to me it makes so much sense now!

mschrage (changes requested):
I think we will need to add a `trigger: "/"` to cause the custom generator to be reevaluated.

ShawnCockburn:
> I think we will need to add a `trigger: "/"` to cause the custom generator to be reevaluated.

done!

loloboin (approved):

mschrage (approved):

fedeci (approved):
Thanks!

PR #976, opened 2022-02-26
feat: add basic `date` completion spec

qubist:

New completion spec

No completion for unix `date` command

Completion for unix `date` command

**Additional info:**

This basically works. The `date` command has some weird structure so I had to create an argument with two potential options for what it does, and document them both in the same argument object in the spec (see https://github.com/qubist/autocomplete/blob/198754849ded82b92df96a9465710de510556a25/src/date.ts#L72-L78). If anyone knows how to do this differently, I'd be happy to learn and implement it: Ideally, `fig` would decide (like `date` does) which interpretation of its one argument to use based on whether it starts with a `+` character or not.

### DIFF
--- a/src/date.ts
+++ b/src/date.ts
@@ -1,0 +1,80 @@
+const completionSpec: Fig.Spec = {
+  name: "date",
+  description: "Display or set date and time",
+  options: [
+    {
+      name: "-d",
+      description: "Set the kernel's value for daylight saving time",
+      args: {
+        name: "dst",
+      },
+    },
+    {
+      name: "-f",
+      description:
+        "Use specified format for input instead of the default [[[mm]dd]HH]MM[[cc]yy][.ss] format",
+      args: [
+        {
+          name: "input_fmt",
+          description: "The format with which to parse the new date value",
+        },
+        {
+          name: "new_date",
+          description: "The new date to set",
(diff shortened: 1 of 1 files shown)


fedeci (approved):
Thanks!

src/date.ts
```diff
@@ -0,0 +1,80 @@
...
+      name: "-R",
+      description: "Use RFC 2822 date and time output format",
+    },
+    {
+      name: "-r",
+      description:
+        "Print the date and time represented by the specified number of seconds since the Epoch",
+      args: {
+        name: "seconds",
+        descrption:
```
fedeci, line 46:
```suggestion
        description:
```

src/date.ts
```diff
@@ -0,0 +1,80 @@
...
+      description:
+        "Display or set the date in UTC (Coordinated Universal) time",
+    },
+    {
+      name: "-v",
+      description:
+        "Adjust and print (but don't set) the second, minute, hour, month day, week day, month, or year according to val",
+      args: {
+        name: "val",
+        descrption: "[+|-]val[ymwdHMS]",
```
fedeci, line 68:
```suggestion
        description: "[+|-]val[ymwdHMS]",
```

Issue #758, opened 2021-11-30
[npm] Support `--workspace` flags for NPM@v7+


jrnail23:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

npm

### Which statement makes the most sense?

This completion spec isn't complete

### Issue Details

When using npm v7+, `--workspace` and `--workspaces` flags should be available on certain commands.

I know this can be a bit tricky, since workspace support has been added for additional commands in various minor version increments.

https://docs.npmjs.com/cli/v7/using-npm/workspaces

harimm:
What seems to be workspace flags are already present, but broken

<img width="554" alt="workspace flags" src="https://user-images.githubusercontent.com/2451152/155839530-296efd61-da1a-46b3-a7db-ab1f5ca7a6fc.png">

The flags should be `-w` or `--workspace` and `-ws` or `--workspaces`

<img width="589" alt="workspace flags help" src="https://user-images.githubusercontent.com/2451152/155839717-0fcc8be9-b968-469e-9c9c-a7109e0aea1a.png">

In the command help, it is written as `-w|--workspace` and `-ws|--workspaces`. The `|` has been misinterpreted as the `l` character when the spec was created

PR #977, opened 2022-02-26
fix(npm): workspace and workspaces flag names

harimm:
Bug Fix: #758

With this fix, the `-w` or `--workspace` and `-ws` or `--workspaces` flags are displayed correctly

**Additional info:**


fedeci (approved):
Thanks!

PR #984, opened 2022-02-28
feat: implement bat spec

threadflow:

Implement bat command spect

bat command could auto-complete now

The feature before is just a copy of cat that introduce the command briefly. 

Now, the whole auto-complete is worked.

**Additional info:**

### DIFF
--- a/src/bat.ts
+++ b/src/bat.ts
@@ -1,9 +1,312 @@
+import { Script } from "vm";
+
 const completionSpec: Fig.Spec = {
   name: "bat",
-  description: "Concatenate and print files",
+  description: "A cat(1) clone with syntax highlighting and Git integration",
+  options: [
+    {
+      name: ["-A", "--show-all"],
+      description: "Show non-printable characters",
+    },
+    {
+      name: ["-p", "--plain"],
+      description: "Show plain style, no decorations",
+    },
+    {
+      name: ["-l", "--language"],
+      description: "Explicitly set the language for syntax highlighting",
+      args: {
+        name: "<language>",
+        generators: {
+          script: "bat --list-languages",
+          postProcess: function (out) {
+            return [].concat(
+              ...out.split("\n").map((outLine) => {
+                const langLine = outLine.split(":");
+                const descriptionString = langLine[0].trim();
+                return outLine
+                  .replace(/\:/g, ",")
+                  .split(",")
+                  .map((item) => {
+                    return {
+                      name: item.trim(),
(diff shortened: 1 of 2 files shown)


threadflow:
👍

src/bat.ts
```diff
@@ -1,9 +1,312 @@
+import { Script } from "vm";
```
fedeci, line 1: Do we need this import?

src/bat.ts
```diff
@@ -1,9 +1,312 @@
...
+                return outLine
+                  .replace(/\:/g, ",")
+                  .split(",")
+                  .map((item) => {
+                    return {
+                      name: item.trim(),
+                      description: descriptionString + " language",
+                    };
+                  });
+              })
```
fedeci, line 36:
```suggestion
              out.split("\n").map((outLine) => {
                const langLine = outLine.split(":");
                const descriptionString = langLine[0].trim();
                return outLine
                  .replace(/\:/g, ",")
                  .split(",")
                  .map((item) => {
                    return {
                      name: item.trim(),
                      description: descriptionString + " language",
                    };
                  });
              }
```

fedeci (changes requested):

src/bat.ts
threadflow: Sorry, this line copied from template and forget to removed before commit. It can be removed.

src/bat.ts
fedeci: Yeah, I missed a return 😁

fedeci: Yeah, I missed a return 😁

fedeci:
```suggestion
              return out.split("\n").map((outLine) => {
                const langLine = outLine.split(":");
                const descriptionString = langLine[0].trim();
                return outLine
                  .replace(/\:/g, ",")
                  .split(",")
                  .map((item) => {
                    return {
                      name: item.trim(),
                      description: descriptionString + " language",
                    };
                  });
              }
```

threadflow: I do test the previous suggested code with adding the 'return'.  It not work.

The code here is used to unpack a 2 dimension array, so the committed code is needed.

fedeci: I see, the I still prefer `Array.flat` method than `.concat` which happens to have some perf issues on some browsers.

threadflow: I refactored this option with `flat()` and tested. Please check again

mschrage (approved):

PR #986, opened 2022-03-01
feat: add svn completion spec

sh-sabbir:
Feature

Currently not available

Partial support for SVN auto-completion

**Additional info:**
Added support for a few subcommands. But I'll continue improving the spec by adding more properties to the spec

### DIFF
--- a/src/svn.ts
+++ b/src/svn.ts
@@ -1,0 +1,293 @@
+const subcommandsList: Fig.Suggestion[] = [
+  {
+    name: "add",
+    type: "subcommand",
+  },
+  {
+    name: "auth",
+    type: "subcommand",
+  },
+  {
+    name: ["blame", "praise", "annotate", "ann"],
+    type: "subcommand",
+  },
+  {
+    name: "cat",
+    type: "subcommand",
+  },
+  {
+    name: ["changelist", "cl"],
+    type: "subcommand",
+  },
+  {
+    name: ["checkout", "co"],
+    type: "subcommand",
+  },
+  {
+    name: "cleanup",
+    type: "subcommand",
+  },
+  {
+    name: ["commit", "ci"],
+    type: "subcommand",
+  },
+  {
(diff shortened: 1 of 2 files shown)


src/svn.ts
```diff
@@ -0,0 +1,293 @@
...
+  description: "The first content tracker",
+  subcommands: [
+    {
+      name: ["help", "h"],
+      description: "Show help for svn",
+      args: {
+        name: "subcommand",
+        description: "Help about specific subcommand",
+        isOptional: true,
+        suggestions: subcommandsList,
```
fedeci, line 238:
```suggestion
        template: ["help"],
```

fedeci, line 238:
```suggestion
        template: ["help"],
```

src/svn.ts
```diff
@@ -0,0 +1,293 @@
...
+          priority: 100,
+        },
+        ...globalOptions,
+      ],
+    },
+  ],
+  options: [
+    {
+      name: "--version",
+      isPersistent: false,
```
fedeci, line 278:
```suggestion
```

src/svn.ts
```diff
@@ -0,0 +1,293 @@
...
+      name: "info",
+      description: "Show information about a local or remote item",
+    },
+    {
+      name: ["checkout", "co"],
+      description: "Check out a working copy from a repository",
+      args: {
+        name: "repository",
+        description: "The repository you want to checkout",
+        isOptional: false,
```
fedeci, line 255:
```suggestion
```

fedeci (changes requested):
A couple nits and it is good to be shipped!

src/svn.ts
fedeci: It returns all the sibling subcommands but appears to be broken rn. I'm fixing it though so you can safely add it!

mschrage (approved):

PR #995, opened 2022-03-03
feat: add `id` spec

anthonycastaneda:
**What kind of change does this PR introduce? New completion spec for id

**Additional info:**
This is my first spec, and first pull request.  I'm a beginner, so the code may not be correct.  It did pass the npm test and lint:fix.

### DIFF
--- a/src/id.ts
+++ b/src/id.ts
@@ -1,0 +1,77 @@
+const exclusiveOptions = ["-A", "-F", "-G", "-M", "-P", "-g", "-p", "-u"];
+
+const completionSpec: Fig.Spec = {
+  name: "id",
+  description:
+    "The id utility displays the user and group names and numeric IDs, of the calling process, to the standard output.  If the real and effective IDs are different, both are displayed, otherwise only the real ID is displayed. If a user (login name or user ID) is specified, the user and group IDs of that user are displayed.  In this case, the real and effective IDs are assumed to be the same",
+  options: [
+    {
+      name: "-A",
+      description:
+        "Display the process audit user ID and other process audit properties, which requires privilege",
+      exclusiveOn: exclusiveOptions,
+    },
+    {
+      name: "-F",
+      description: "Display the full name of the user",
+      exclusiveOn: exclusiveOptions,
+    },
+    {
+      name: "-G",
+      description:
+        "Display the different group IDs (effective, real and supplementary) as white-space separated numbers, in no particular order",
+      exclusiveOn: exclusiveOptions,
+      args: {
+        name: "-n",
+        description:
+          "Display the name of the user or group ID for the -G, -g and -u options instead of the number.  If any of the ID numbers cannot be mapped into names the number will be displayed as usual",
+      },
+    },
+    {
+      name: "-M",
+      description: "Display the MAC label of the current process",
+      exclusiveOn: exclusiveOptions,
+    },
+    {
+      name: "-P",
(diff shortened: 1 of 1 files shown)


src/id.ts
```diff
@@ -0,0 +1,77 @@
...
+    },
+    {
+      name: "-p",
+      description: "Make the output human-readable",
+      exclusiveOn: exclusiveOptions,
+    },
+    {
+      name: "-u",
+      description: "Display the effective user ID as a number",
+      exclusiveOn: exclusiveOptions,
```
SeparateRecords, line 63: `-n` can only be specified if one of `-G`, `-g`, or `-u` are given

```suggestion
      dependsOn: ["-G", "-g", "-u"],
      exclusiveOn: exclusiveOptions
```

src/id.ts
```diff
@@ -0,0 +1,77 @@
...
+        name: "-nr",
+        description:
+          "Display the name of the user or group ID for the -G, -g and -u options instead of the number.  If any of the ID numbers cannot be mapped into names the number will be displayed as usual",
+      },
+    },
+  ],
+  // Only uncomment if id takes an argument
+  args: {
+    name: "user",
+    isOptional: true,
```
SeparateRecords, line 74: This generator _should_ show all the users on the computer! It's lifted [straight out of `killall.ts`](https://github.com/withfig/autocomplete/blob/126dbafce73301ae36af84ee7e163b9923cbf438/src/killall.ts#L116-L124)

```suggestion
    isOptional: true,
    generators: {
      script: "dscl . -list /Users | grep -v '^_'",
      postProcess: (out) =>
        out
          .trim()
          .split("\n")
          .map((username) => ({
            name: username,
            icon: "fig://template?badge=👤",
          })),
    }
```

SeparateRecords (dismissed):
Looking really good!! Left a couple of comments, all super easy to fix up.

There's also one option missing (`-r`) that should be added.

Quick note, don't just use "commit suggestion" since that will do individual commits and make me co-author of them haha

@fedeci I'll defer to you on whether we should keep the `exclusiveOn` or `dependsOn` - it's really hard for me to visualize the relationship between these damn options...

src/id.ts
anthonycastaneda: The code block shows `-u` but your comment mentions `-n`.  I've made this change on the `-n` option.

src/id.ts
```diff
@@ -0,0 +1,74 @@
...
+  name: "id",
+  description:
+    "The id utility displays the user and group names and numeric IDs, of the calling process, to the standard output.  If the real and effective IDs are different, both are displayed, otherwise only the real ID is displayed. If a user (login name or user ID) is specified, the user and group IDs of that user are displayed.  In this case, the real and effective IDs are assumed to be the same",
+  options: [
+    {
+      name: "-A",
+      description:
+        "Display the process audit user ID and other process audit properties, which requires privilege",
+      exclusiveOn: exclusiveOptions,
+    },
```
mschrage, line 13: I'm not sure if you can have "-A" in the list of exclusive options for itself? @fedeci any idea if this is okay or not?

mschrage (changes requested):

src/id.ts
fedeci: I'm checking!

fedeci: It doesn't appear to be a problem!

fedeci (approved):
Thanks!

PR #1002, opened 2022-03-06
feat: add `softwareupdate` spec

harimm:
Feature: New completion spec

CLI request: closes https://github.com/withfig/autocomplete/issues/942
Spec does not exist for `softwareupdate` command 

Adds completion spec for `softwareupdate` command

**Additional info:**

### DIFF
--- a/src/softwareupdate.ts
+++ b/src/softwareupdate.ts
@@ -24,18 +24,15 @@ const completionSpec: Fig.Spec = {
         {
           name: ["-a", "--all"],
           description: "All appropriate updates",
-          dependsOn: ["-i", "--install"],
         },
         {
           name: ["-R", "--restart"],
           description:
             "Automatically restart (or shut down) if required to complete installation",
-          dependsOn: ["-i", "--install"],
         },
         {
           name: ["-r", "--recommended"],
           description: "Only recommended updates",
-          dependsOn: ["-i", "--install"],
         },
       ],
     },
--- a/src/softwareupdate.ts
+++ b/src/softwareupdate.ts
@@ -74,10 +74,13 @@ const completionSpec: Fig.Spec = {
       name: "--history",
       description:
         "Show the install history.  By default, only displays updates installed by softwareupdate",
-    },
-    {
-      name: "--all",
-      description: "Include all processes in history (including App installs)",
+      options: [
+        {
+          name: "--all",
+          description:
+            "Include all processes in history (including App installs)",
+        },
+      ],
     },
   ],
   options: [
--- a/src/softwareupdate.ts
+++ b/src/softwareupdate.ts
@@ -1,0 +1,128 @@
+const completionSpec: Fig.Spec = {
+  name: "softwareupdate",
+  description:
+    "Software Update checks for new and updated versions of your software based on information about your computer and current software",
+  subcommands: [
+    /** Manage updates: */
+    {
+      name: ["-l", "--list"],
+      description: "List all appropriate update labels",
+    },
+    {
+      name: ["-d", "--download"],
+      description: "Download Only",
+    },
+    {
+      name: ["-i", "--install"],
+      description: "Install",
+      args: {
+        name: "label",
+        isVariadic: true,
+        isOptional: true,
+      },
+      options: [
+        {
+          name: ["-a", "--all"],
+          description: "All appropriate updates",
+          dependsOn: ["-i", "--install"],
+        },
+        {
+          name: ["-R", "--restart"],
+          description:
+            "Automatically restart (or shut down) if required to complete installation",
+          dependsOn: ["-i", "--install"],
+        },
+        {
+          name: ["-r", "--recommended"],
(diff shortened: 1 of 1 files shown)


src/softwareupdate.ts
```diff
@@ -0,0 +1,130 @@
+const completionSpec: Fig.Spec = {
+  name: "softwareupdate",
+  description:
+    "Software Update checks for new and updated versions of your software based on information about your computer and current software",
+  subcommands: [
```
fedeci, line 5:
```suggestion
  options: [
```

src/softwareupdate.ts
```diff
@@ -0,0 +1,130 @@
...
+      description: "List all appropriate update labels",
+    },
+    {
+      name: ["-d", "--download"],
+      description: "Download Only",
+    },
+    {
+      name: ["-i", "--install"],
+      description: "Install",
+      args: {
```
fedeci, line 18: Can we add a generator for this arg? It should be fairly simple using the `softwareupdate --list` results

src/softwareupdate.ts
```diff
@@ -0,0 +1,130 @@
...
+          name: ["-R", "--restart"],
+          description:
+            "Automatically restart (or shut down) if required to complete installation",
+        },
+        {
+          name: ["-r", "--recommended"],
+          description: "Only recommended updates",
+          exclusiveOn: ["-a", "--all"],
+        },
+      ],
```
fedeci, line 39: Since we cannot have options of options, we need to `dependsOn: ["--install"]` to these.

src/softwareupdate.ts
```diff
@@ -0,0 +1,130 @@
...
+      name: "--history",
+      description:
+        "Show the install history.  By default, only displays updates installed by softwareupdate",
+      options: [
+        {
+          name: "--all",
+          description:
+            "Include all processes in history (including App installs)",
+        },
+      ],
```
fedeci, line 83: Same here

src/softwareupdate.ts
```diff
@@ -0,0 +1,130 @@
...
+    {
+      name: "--verbose",
+      description: "Enable verbose output",
+      isPersistent: true,
+    },
+    {
+      name: ["--help", "-h"],
+      description: "Show help for softwareupdate",
+    },
+  ],
```
fedeci, line 128: `isPersistent` really doesn't make sense if there is no subcommand, so we can remove it safely!

fedeci (changes requested):
I left some comments to improve the spec, but thanks for writing this down!
p.s. when you link an issue you can use a closing keyword to make the issue close automatically when this PR gets merged

src/softwareupdate.ts
harimm: I initally wrote these stuff as options but the man page mentions several of these as `commands` and not `options`. Also when this is what `softwareupdate -h` gives you

```
softwareupdate -h
usage: softwareupdate <cmd> [<args> ...]

** Manage Updates:
	-l | --list		List all appropriate update labels (options:  --no-scan, --product-types)
	-d | --download		Download Only
	-i | --install		Install
		<label> ...	specific updates
		-a | --all		All appropriate updates
		-R | --restart		Automatically restart (or shut down) if required to complete installation.
		-r | --recommended	Only recommended updates
	--list-full-installers		List the available macOS Installers
	--fetch-full-installer		Install the latest recommended macOS Installer
		--full-installer-version	The version of macOS to install. Ex: --full-installer-version 10.15
	--install-rosetta	Install Rosetta 2
	--background		Trigger a background scan and update operation

** Other Tools:
	--dump-state		Log the internal state of the SU daemon to /var/log/install.log
	--evaluate-products	Evaluate a list of product keys specified by the --products option
	--history		Show the install history.  By default, only displays updates installed by softwareupdate.
	--all 			Include all processes in history (including App installs)

** Options:
	--no-scan		Do not scan when listing or installing updates (use available updates previously scanned)
	--product-types <type>		Limit a scan to a particular product type only - ignoring all others
		Ex:  --product-types macOS  || --product-types macOS,Safari
	--products		A comma-separated (no spaces) list of product keys to operate on.
	--force			Force an operation to complete.  Use with --background to trigger a background scan regardless of "Automatically check" pref
	--agree-to-license		Agree to the software license agreement without user interaction.

	--verbose		Enable verbose output
	--help			Print this help
```
Only the ones mentioned under `** Options I have added as option

fedeci: Let's keep it as is then!

fedeci (approved):
Thanks!

fedeci:
I forgot about this [comment](https://github.com/withfig/autocomplete/pull/1002#discussion_r820212761), @harimm would you mind opening a new PR adding that generator?

harimm:
@fedeci Created draft PR https://github.com/withfig/autocomplete/pull/1008/

Keeping it as draft as I haven't been able to test it out yet

PR #1008, opened 2022-03-07
feat(softwareupdate): add generator for available updates

harimm:
Adds a generator for listing available updates for `softwareupdate`

**Additional info:**
Based on comment on https://github.com/withfig/autocomplete/pull/1002


src/softwareupdate.ts
```diff
@@ -1,3 +1,16 @@
...
+  script: "softwareupdate -l",
+  postProcess: (out) => {
+    return out
+      .split("\n")
+      .filter((line) => line.startsWith("* Label: "))
+      .map((line) => ({
+        name: '"' + line.substring(9) + '"',
+        description: "Available update",
+      }));
+  },
```
fedeci, line 11: This should work but we need to specify a timeout and a cache because the process is really slow to return the results.

src/softwareupdate.ts
```diff
@@ -1,3 +1,16 @@
+const updatesGenerator: Fig.Generator = {
+  script: "softwareupdate -l",
```
SeparateRecords, line 2: Can we use the longer `--list` option? The verbosity make it a little nicer to read 🙂

harimm: Updated. Thanks.

src/softwareupdate.ts
harimm: I updated the timeout to 20 seconds and cache ttl to 5 minutes. Does that seem reasonable or do you have any suggestions? I don't see the scriptTimeout used anywhere so far

Also, I have a question about the generator. When does the generator script execute? Does it execute when the base command itself it typed or does it only run once the associated subcommand/arg/option is typed?

src/softwareupdate.ts
```diff
@@ -1,3 +1,20 @@
...
+      .split("\n")
+      .filter((line) => line.startsWith("* Label: "))
+      .map((line) => ({
+        name: '"' + line.substring(9) + '"',
+        description: "Available update",
+      }));
+  },
+  scriptTimeout: 20000,
+  cache: {
+    ttl: 300000,
```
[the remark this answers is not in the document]
fedeci: Cache revalidation added

src/softwareupdate.ts
```diff
@@ -1,3 +1,20 @@
...
+    return out
+      .split("\n")
+      .filter((line) => line.startsWith("* Label: "))
+      .map((line) => ({
+        name: '"' + line.substring(9) + '"',
+        description: "Available update",
+      }));
+  },
+  scriptTimeout: 20000,
+  cache: {
```
fedeci, line 13:
```suggestion
  cache: {
    strategy: "stale-while-revalidate"
```

fedeci:
@harimm Is this ready to be merged after the SWR cache strategy?

src/softwareupdate.ts
harimm: Updated

harimm:
@fedeci I have made one small change to the generator to insert with enclosing quotes. Should be good to go. Out of curiosity, how does the SWR strategy work w.r.to the configured ttl?

fedeci:
We fetch the data and give that the ttl lifetime. Each time that data is requested again if data is still valid (respects ttl) we just return that cached data, otherwise we return that cached data and we send a new fetch.

fedeci (approved):
👍

fedeci:
Thanks @harimm!

PR #1009, opened 2022-03-07
feat: add `meteor` spec

couturecraigj:

Adding Meteor to the autocomplete suggestions but specifically the `npm subcommand`

It does not suggest it because it is not in there.

It now suggests at least Meteor npm suggestions

**Additional info:**

### DIFF
--- a/src/meteor.ts
+++ b/src/meteor.ts
@@ -1,0 +1,16 @@
+import npm from "./npm";
+
+const completionSpec: Fig.Spec = {
+  name: "meteor",
+  description: "",
+  subcommands: [npm],
+  options: [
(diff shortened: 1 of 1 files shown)


src/meteor.ts
```diff
@@ -0,0 +1,16 @@
+import npm from "./npm";
+
+const completionSpec: Fig.Spec = {
+  name: "meteor",
+  description: "`meteor` command-line tool",
+  subcommands: [npm],
```
fedeci, line 6:
```suggestion
  subcommands: [{
    name: "npm",
    loadSpec: "npm"
  }],
```

PR #1010, opened 2022-03-07
feat: added support for Playwright

wladiston:
Feature

Does not support Playwright commands

Added support for few commands used by playwright

**Additional info:**

### DIFF
--- a/src/npx.ts
+++ b/src/npx.ts
@@ -80,6 +80,10 @@ const suggestions: Fig.Suggestion[] = [
     icon: "https://remix.run/favicon-light.1.png",
   },
   {
+    name: "playwright",
+    icon: "https://playwright.dev/img/playwright-logo.svg",
+  },
+  {
     name: "ignite-cli",
     icon: "🔥",
   },
--- a/src/playwright.ts
+++ b/src/playwright.ts
@@ -5,7 +5,7 @@ const testsGenerator: Fig.Generator = {
   },
 };
 
-const browserSuggestions: Fig.Arg["suggestions"] = [
+const browserSuggestions: Fig.Suggestion[] = [
   {
     name: "chromium",
     displayName: "Chromium",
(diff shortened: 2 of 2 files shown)


src/npx.ts
```diff
@@ -79,6 +79,10 @@ const suggestions: Fig.Suggestion[] = [
     name: "remix",
     icon: "https://remix.run/favicon-light.1.png",
   },
+  {
+    name: "playwright",
+    icon: "https://playwright.dev/img/playwright-logo.svg",
+  },
```
fedeci, line 85: This needs to be added to the yarn spec too

src/playwright.ts
```diff
@@ -0,0 +1,106 @@
+const testsGenerator: Fig.Generator = {
+  custom: async (tokens, executeShellCommand) => {
+    // TODO: load the list of test files specified in testDir on playwright.config file
+    return [] as Fig.Suggestion[];
+  },
+};
+
+const browserSuggestions: Fig.Arg["suggestions"] = [
```
fedeci, line 8:
```suggestion
const browserSuggestions: Fig.Suggestion[] = [
```

src/playwright.ts
```diff
@@ -0,0 +1,106 @@
...
+  {
+    name: "webkit",
+    displayName: "WebKit",
+  },
+];
+
+const helpOption: Fig.Option = {
+  name: ["--help", "-h"],
+  description: "Display help for command",
+  priority: -1,
```
fedeci, line 46: -1 is not a valid priority, to always have this at the bottom of the list use a priority : 0 <= p <= 25
You can also add `isPersistent: true` to the --help option so that each subcommand of the spec will have it.
```suggestion
  priority: 1,
```

src/playwright.ts
```diff
@@ -0,0 +1,106 @@
...
+        isVariadic: true,
+        template: ["filepaths", "folders"],
+      },
+      options: [
+        {
+          name: "-g",
+          description: "Run the test with the title",
+          args: {
+            name: "title",
+            isOptional: false,
```
fedeci, line 69: `isOptional: false` is implicit
```suggestion
```

fedeci:
I left some suggestions, let me know what you think!

PR #1015, opened 2022-03-08
fix: add `{cursor}` to options that insert an `=`

SeparateRecords:
Related: https://github.com/withfig/fig/issues/1031

This was a quick regex find-replace, please let me know if I missed anything.

### DIFF
--- a/src/wp.ts
+++ b/src/wp.ts
@@ -76,7 +76,7 @@ const global_parameter_skip_plugins2: Fig.Option = {
   name: "--skip-plugins",
   insertValue: "--skip-plugins={cursor}",
   requiresEquals: true,
-  displayName: "--skip-plugins={cursor}",
+  displayName: "--skip-plugins=",
   description:
     "Skip loading all plugins, or a comma-separated list of plugins. Note: mu-plugins are still loaded",
   args: {
@@ -145,7 +145,7 @@ const global_parameter_debug1: Fig.Option = {
 
 const global_parameter_debug2: Fig.Option = {
   name: "--debug",
-  displayName: "--debug={cursor}",
+  displayName: "--debug=",
   requiresEquals: true,
   insertValue: "--debug={cursor}",
   description:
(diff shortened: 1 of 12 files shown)


src/wp.ts
```diff
@@ -69,8 +69,8 @@ const global_parameter_skip_plugins1: Fig.Option = {
 
 const global_parameter_skip_plugins2: Fig.Option = {
   name: "--skip-plugins",
-  insertValue: "--skip-plugins=",
-  displayName: "--skip-plugins=",
+  insertValue: "--skip-plugins={cursor}",
+  displayName: "--skip-plugins={cursor}",
```
mschrage, line 73: `{cursor}` shouldn't be in display name.

src/wp.ts
```diff
@@ -85,8 +85,8 @@ const global_parameter_skip_themes1: Fig.Option = {
 
 const global_parameter_skip_themes2: Fig.Option = {
   name: "--skip-themes",
-  insertValue: "--skip-themes=",
-  displayName: "--skip-themes=",
+  insertValue: "--skip-themes={cursor}",
+  displayName: "--skip-themes={cursor}",
```
mschrage, line 89: `{cursor}` shouldn't be in display name.

mschrage (commented):
Do we want to specify the `insertValue` like this? Or should we just set `requiresEquals: true`

SeparateRecords:
`requiresEquals` should be set on these for sure. Will do that. It currently doesn't insert the equals or put the cursor in the right spot though, so this still has to be done

SeparateRecords:
`requiresEquals: true` is now on options that insert `...={cursor}`

mschrage:
Should `requiresEquals` automatically set the insertion text to `name={cursor}`?

SeparateRecords:
I think so! There's an [old convo in the fig discord about exactly this actually](https://discord.com/channels/837809111248535583/839904572993568820/908897442878078977)

fedeci:
@SeparateRecords do we have an issue in GH to track that?

fedeci (approved):

PR #1028, opened 2022-03-16
chore(fig): Update spec to B405

SeparateRecords:
Fixes `internal` subcommand to be properly hidden and aliased

### DIFF
--- a/src/fig/fig@latest.ts
+++ b/src/fig/fig@latest.ts
@@ -1287,7 +1287,6 @@ const completion: Fig.Spec = {
                   description: "Fish shell",
                 },
               ],
-              isRequired: true,
             },
             {
               name: "when",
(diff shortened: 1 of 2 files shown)


fedeci (commented):
I am starting to change those props today in clap. We should also try to check if the push-to-autocomplete-action correctly works for this spec

SeparateRecords:
> I am starting to change those props today in clap.

🙌 You're amazing!

> We should also try to check if the push-to-autocomplete-action correctly works for this spec

For sure, would be best to dogfood it with our own. Does that block us from merging this first? Should we test on this branch?

fedeci:
I think we will have new feature branches to test the action, let's merge this one first!

fedeci (approved):
Thanks!

PR #1032, opened 2022-03-16
feat: add `publish` spec

SeparateRecords:
CLI for https://github.com/JohnSundell/Publish

### DIFF
--- a/src/publish.ts
+++ b/src/publish.ts
@@ -1,0 +1,45 @@
+const completionSpec: Fig.Spec = {
+  name: "publish",
+  description: "",
+  subcommands: [
+    {
+      name: "new",
+      description: "Set up a new website in the current folder",
+    },
+    {
+      name: "run",
+      description:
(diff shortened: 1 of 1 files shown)


fedeci (approved):

src/publish.ts
```diff
@@ -0,0 +1,45 @@
...
+    {
+      name: "new",
+      description: "Set up a new website in the current folder",
+    },
+    {
+      name: "run",
+      description:
+        "Generate and run a localhost server on default port 8000 for the website in the current folder",
+      parserDirectives: {
+        flagsArePosixNoncompliant: true,
```
fedeci, line 14: @SeparateRecords is this required?

SeparateRecords: I think so, `-p` won't work with `-p8000`

fedeci: Yes it wouldn't, that's correct then!

PR #1034, opened 2022-03-18
chore: replace deprecated `String.prototype.substr()`

CommanderRoot:

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

No user facing change

**Additional info:**

### DIFF
--- a/src/yo.ts
+++ b/src/yo.ts
@@ -3,7 +3,7 @@ function toTitleCase(str: string): string {
     .trim()
     .replace(
       /\w\S*/g,
-      (txt) => txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase()
+      (txt) => txt.charAt(0).toUpperCase() + txt.slice(1).toLowerCase()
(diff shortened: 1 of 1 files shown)


fedeci:
Is it the only occurrence of `.substr`?

CommanderRoot:
My IDE only found the changed one and https://github.com/withfig/autocomplete/blob/b75b676416c9b05ba342efda42303f538d2203b8/src/cargo.ts#L13
But that's not JavaScript so it's not affected by the deprecation

fedeci (approved):
Thanks!

PR #1035, opened 2022-03-19
feat(deno): update to 1.20, improve doc generator

SeparateRecords:
Adds `deno task` and `deno bench`, improves (and simplifies) the logic for the `deno doc` suggestion generator.

The deno config JSON schema also allows for additional keys. This spec uses the `"fig"` property to customize the:
* Priority
* Display name
* Icon
* Description
* Hidden status

This PR also adds a deno directory with supporting files:
1. Vendored type definitions for deno doc (please let me know if I need to do this differently)
2. Automatically generated types for the config JSON schema

## Images

Fig 1. Task suggestions
Task suggestions

Fig 2. Example config for the above
Configuration file containing fig property

Fig 3. `deno doc` suggesting something it couldn't before
Documentation suggestions

### DIFF
--- a/src/deno.ts
+++ b/src/deno.ts
@@ -699,7 +699,7 @@ function getUniqueValues<T, U>(input: readonly T[], fn: (item: T) => U): T[] {
   return out;
 }
 
-function getDocNodeChildren(node: Node): Node[] {
+function getNodeChildren(node: Node): Node[] {
   if (!("kind" in node)) {
     return [];
   }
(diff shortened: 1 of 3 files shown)


src/deno.ts
```diff
@@ -1251,6 +1251,127 @@ const denoCache: Fig.Subcommand = {
...
+          const configPath = getConfigPath(tokens);
+          let configString: string;
+          if (configPath) {
+            configString = await executeShellCommand(`\\cat '${configPath}'`);
+          } else {
+            configString = await executeShellCommand(
+              "until [[ ( -f deno.json || -f deno.jsonc || $PWD = '/' ) ]]; do cd ..; done; \\cat deno.json 2>/dev/null || \\cat deno.jsonc 2>/dev/null"
+            );
+          }
+          const config = JSON.parse(configString) as DenoConfig;
```
SeparateRecords, line 1354: This will fail when parsing jsonc — needs to be fixed.

SeparateRecords:
Reverted to draft because this doesn't handle doc nodes having the same name, and the scope has increased

fedeci (approved):
This really leverages the deno spec to another level!

PR #1038, opened 2022-03-20
feat(fig/fig@latest): update spec

mschrage:

### DIFF
--- a/src/fig/fig@latest.ts
+++ b/src/fig/fig@latest.ts
@@ -1,0 +1,1395 @@
+import {
+  edit,
+  settingsSpecGenerator,
+  subsystemsGenerator,
+  themesGenerator,
+} from "./shared";
+const completion: Fig.Spec = {
+  name: "fig",
+  description: "The CLI for Fig",
+  subcommands: [
+    {
+      name: "app",
+      description: "Interact with the desktop app",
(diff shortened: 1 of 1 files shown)


src/fig/fig@latest.ts
```diff
@@ -0,0 +1,1395 @@
...
+  debug: (debug) =>
+    edit(debug.subcommands, {
+      logs: (logs) =>
+        edit(logs.args, {
+          files: {
+            isVariadic: true,
+          },
+        }),
+    }),
+});
```
fedeci, line 1395: /cc @SeparateRecords this can be removed, right?

SeparateRecords: Looks like the edits need to stay for now - the `internal` subcommand is aliased to `_`, and the `fig debug logs [files...]` arg is variadic, but it doesn't look like that's in the spec. I'd imagine that's easy enough to fix on the rust side though!

fedeci: The variadic prop is added when `Arg::multiple_values(true)`, aliases are still unhandled.
/cc @grant0417

fedeci: Opened a PR for subcommand aliases.

PR #1044, opened 2022-03-23
feat(fig/fig@latest): update spec

mschrage:


fedeci:
@grant0417 @SeparateRecords we should be able to remove the bottom section of the spec where we assign the aliases and the repeatable options now!

fedeci:
@grant0417 @SeparateRecords we should be able to remove the bottom section of the spec where we assign the aliases and the repeatable options now!

grant0417:
> @grant0417 @SeparateRecords we should be able to remove the bottom section of the spec where we assign the aliases and the repeatable options now!

Awesome, thanks for all your work with the action and clap generator! 👍

PR #1045, opened 2022-03-23
feat: update node spec

ac-f:
image
Adding spec of node: `--preserve-symlinks`


mschrage (approved):
Thanks so much for contributing!

PR #1049, opened 2022-03-24
refactor(deno): move generators to another file

SeparateRecords:
This is in preparation for switching to an automatically generated spec 🎉 

Must be merged after #1048 (this also adds swr to deno tasks generator)


fedeci (approved):
🚀

PR #1054, opened 2022-03-24
feat: add drush spec

andeersg:
Feature

Missing support for drush autocomplete.

Autocomplete support for drush.

**Additional info:**

- [x] Basic autocomplete for commands

### DIFF
--- a/src/drush.ts
+++ b/src/drush.ts
@@ -11,7 +11,7 @@ interface DrushOption {
   shortcut: string;
   accept_value: boolean;
   is_value_required: boolean;
-  is_multiple: boolean; // not supported by fig
+  is_multiple: boolean; // isRepeatable in fig
   description: string;
   default: null | boolean; // not supported by fig
 }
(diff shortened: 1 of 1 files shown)


src/drush.ts
```diff
@@ -0,0 +1,96 @@
...
+  description: string;
+  default: null | string | Array<string>;
+}
+
+interface DrushOption {
+  name: string;
+  shortcut: string;
+  accept_value: boolean;
+  is_value_required: boolean;
+  is_multiple: boolean; // not supported by fig
```
fedeci, line 14: Is this what you are looking for? https://fig.io/docs/reference/option#isrepeatable

andeersg: It's a property from drush output, but I guess there is no need to have it defined here since it's not used. Will remove them 👍

andeersg: That's true, will fix it 👍

fedeci (approved):
Thanks!

PR #1061, opened 2022-03-25
fix(open): update completion spec description

raviriley:

`open` completion spec description update.

The description only describes opening files, even though [(as seen on @fig's Twitter)](https://twitter.com/fig/status/1507028223879614466) the `open` command can also open applications and URLs.


fedeci (approved):
Thanks!

PR #1063, opened 2022-03-26
feat: add `fisher` completion spec

Tim4497:

Missing support for [Fisher](https://github.com/jorgebucaran/fisher) autocomplete.

Added basic autocomplete, there is no complete list of all available fisher plugins so this will be a work in progress.


src/fisher.ts
```diff
@@ -0,0 +1,145 @@
...
+    {
+      name: "list",
+      description: "List plugins",
+      args: [
+        {},
+        {
+          name: "RegEx",
+          description: "Search in list with regular expression",
+          isOptional: true,
+        },
```
mschrage, line 130: Is it intentional to have an empty argument here?

Is this how the command would be invoked?
```
fisher list xxx '.*?'
```

Tim4497: My intention was to indicate that you just can press enter to list all plugins, (like a empty regex)

fedeci: @Tim4497 then you can safely remove that empty arg!
With this structure:
```ts
args: [
        {
          name: "RegEx",
          description: "Search in list with regular expression",
          isOptional: true,
        },
]
```

- `fisher list`: returns a list of all plugins
- `fisher list [regex]` will return all the plugins that match the regex

Tim4497:
worked on another completion spec and merged it into master but now its reverted again

fedeci (approved):
Thanks!

PR #1076, opened 2022-03-29
feat: add `tccutil` spec

SeparateRecords:
Not that this matters much since if you need to use `tccutil` Fig probably isn't working, but it's a fun and small spec!

image

Getting a list of bundle IDs is a surprisingly difficult task... I'll reuse this generator for `open -b`
image

### DIFF
--- a/src/tccutil.ts
+++ b/src/tccutil.ts
@@ -36,10 +36,12 @@ const services: Fig.Suggestion[] = [
   },
   {
     name: "FileProviderPresence",
-    description:
-      "Do you want to allow client to see when you are using files managed by it? It will see which applications are used to access files and whether you are actively using them. It will not see when files that are not managed by it are accessed",
+    description: "Allow the app to see when you are using files managed by it?",
   },
-  { name: "Location", description: "Use your current location" },
+  {
+    name: "Location",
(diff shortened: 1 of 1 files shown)


src/tccutil.ts
```diff
@@ -0,0 +1,183 @@
...
+  {
+    name: "DeveloperTool",
+    description:
+      "Allows app to run software locally that do not meet the system’s security polic",
+  },
+].map((item) => Object.assign(item, { icon: "⚙️" }));
+
+const generateBundleIds = (path: string): Fig.Generator => ({
+  scriptTimeout: 15000,
+  cache: { ttl: 1000 * 60 * 60 * 24 },
```
fedeci, line 143: We can use swr here. It will guarantee us to always get some data (stale or not).

SeparateRecords: Done!

fedeci: We can also keep the ttl to avoid rerunning the script in background each time :)

mschrage (approved):

mschrage:
@SeparateRecords I'll merge this in but can you fix lint error @SeparateRecords (I know it wasn't from your PR)

PR #1078, opened 2022-03-29
feat(mix): add options for `mix run`

Darth-Knoppix:
Improve option suggestions for the `mix run` command based off [Mix run documentation](https://hexdocs.pm/mix/1.12/Mix.Tasks.Run.html#module-command-line-options).

Fig does not suggest any options for the `mix run` command.

**What is the new behavior?**

Fig will suggest options like:
- `mix run --no-halt`
- `mix run --no-deps-check` 

**Additional info:**
- [Documentation with the options the run subcommand can take](https://hexdocs.pm/mix/1.12/Mix.Tasks.Run.html#module-command-line-options).

### DIFF
--- a/src/mix.ts
+++ b/src/mix.ts
@@ -83,34 +83,42 @@ const completionSpec: Fig.Spec = {
         {
           name: "--preload-modules",
           description: "Preloads all modules defined in applications",
+          args: {},
         },
         {
           name: "--no-compile",
           description: "Does not compile even if files require compilation",
+          args: {},
         },
         {
           name: "--no-deps-check",
           description: "Does not check dependencies",
+          args: {},
         },
         {
           name: "--no-archives-check",
           description: "Does not check archives",
+          args: {},
         },
         {
           name: "--no-halt",
           description: "Does not halt the system after running the command",
+          args: {},
         },
         {
           name: "--no-mix-exs",
           description: "Allows the command to run even if there is no mix.exs",
+          args: {},
         },
         {
           name: "--no-start",
           description: "Does not start applications after compilation",
+          args: {},
         },
         {
           name: "--no-elixir-version-check",
           description: "Does not check the Elixir version from mix.exs",
+          args: {},
         },
       ],
     },
--- a/src/mix.ts
+++ b/src/mix.ts
@@ -1,7 +1,14 @@
+import { filepaths } from "@fig/autocomplete-generators";
+
 const constants = {
   elixir_logo_url:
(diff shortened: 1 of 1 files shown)


src/mix.ts
```diff
@@ -46,6 +46,82 @@ const completionSpec: Fig.Spec = {
...
+            name: "code",
+            description: "String contianing code to execute",
+          },
+        },
+        {
+          name: ["-r", "--require"],
+          description: "Executes the given pattern/file",
+          args: {
+            name: "file|pattern",
+            description: "The file|pattern to execute",
```
fedeci, line 73:
```suggestion
            description: "The file|pattern to execute",
            template: "filepaths",
```

src/mix.ts
```diff
@@ -46,6 +46,82 @@ const completionSpec: Fig.Spec = {
...
     },
+    {
+      name: "run",
+      description: "Starts the current application and runs code",
+      options: [
+        {
+          name: "--config",
+          description: "Loads the given configuration files",
+          args: {
+            name: "file",
```
fedeci, line 57:
```suggestion
            name: "file",
            template: "filepaths",
```

src/mix.ts
```diff
@@ -46,6 +46,82 @@ const completionSpec: Fig.Spec = {
...
+          args: {},
+        },
+        {
+          name: "--no-start",
+          description: "Does not start applications after compilation",
+          args: {},
+        },
+        {
+          name: "--no-elixir-version-check",
+          description: "Does not check the Elixir version from mix.exs",
```
fedeci, line 120: Are you sure these accept an argument?

src/mix.ts
```diff
@@ -46,6 +46,82 @@ const completionSpec: Fig.Spec = {
...
+          args: {
+            name: "file",
+          },
+        },
+        {
+          name: ["-e", "--eval"],
+          description: "Evaluates the given code",
+          args: {
+            name: "code",
+            description: "String contianing code to execute",
```
fedeci, line 65:
```suggestion
            description: "String containing code to execute",
            isScript: true,
```

src/mix.ts
```diff
@@ -46,6 +46,82 @@ const completionSpec: Fig.Spec = {
...
+            name: "file|pattern",
+            description: "The file|pattern to execute",
+          },
+        },
+        {
+          name: ["-p", "--parallel"],
+          description: "Executes the given pattern/file",
+          args: {
+            name: "file|pattern",
+          },
```
fedeci, line 81:
```suggestion
          description: "Makes all requires parallel",
```

fedeci (changes requested):

src/mix.ts
Darth-Knoppix: Thanks, I missed this in the docs 👌

src/mix.ts
Darth-Knoppix: Ah, this was a misunderstanding. Fixing now 👍

Darth-Knoppix:
Thanks for the tips and great feedback @fedeci!

I used the `filepaths` template but added a custom `filterTemplateSuggestions` to bump up the priority for script files (.exs) as they are the most likely to be run too.

src/mix.ts
```diff
@@ -99,6 +176,16 @@ const completionSpec: Fig.Spec = {
...
+function generatePreferredFilepaths(): Fig.Generator {
+  return {
+    template: "filepaths",
+    filterTemplateSuggestions: (paths) =>
+      paths.map((path) =>
+        path.name.endsWith(".exs") ? { ...path, priority: 75 } : path
+      ),
+  };
+}
+
```
mschrage, line 188: This doesn't need to be a function - it can just be a constant.
```suggestion
const preferredFilepaths: Fig.Generator = {
    template: "filepaths",
    filterTemplateSuggestions: (paths) =>
      paths.map((path) =>
        path.name.endsWith(".exs") ? { ...path, priority: 75 } : path
      ),
}

```

fedeci: It can be also simplified with the `@fig/autocomplete-generators` package!

Darth-Knoppix: Thanks, I didn't see this either 🤦. I've simplified it based on what's in the elixir autocomplete using `@fig/autocomplete-generators` as suggested 🙏

fedeci (approved):
Thanks!

PR #1079, opened 2022-03-30
feat(open): improve generators

SeparateRecords:
`open -a` now lists all apps on your system - this is surprisingly fast. `open` can open all of these, not just apps in /Applications or ~/Applications

image

Adds completions for `open -b`. This is surprisingly slow. Moved generator from tccutil

image


fedeci (approved):
Thanks!

PR #1086, opened 2022-03-31
feat(robot): add generators

davmarek:
Feature - new generator for `robot spec`

Currently there is no generator for suggesting test cases

Now there is a generator for suggesting test cases from `.robot` files

### DIFF
--- a/src/robot.ts
+++ b/src/robot.ts
@@ -1,10 +1,11 @@
 import { filepaths } from "@fig/autocomplete-generators";
 
 const tagsGenerator: Fig.Generator = {
-  custom: async (tokens, executeShellCommand) => {
-    const out = await executeShellCommand(
-      'for i in $(find -E . -regex ".*.robot" -type f); do cat -s $i ; done'
-    );
+  script:
+    'for i in $(find -E . -regex ".*.robot" -type f); do cat -s $i ; done',
+  postProcess: (out) => {
+    // find all lines with tags
+    // regex: line that starts with 2+ spaces, than '[Tags]  ' and words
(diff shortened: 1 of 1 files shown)


src/robot.ts
```diff
@@ -46,6 +46,38 @@ const variablesGenerator: Fig.Generator = {
   },
 };
 
+const testCasesGenerator: Fig.Generator = {
+  custom: async (tokens, executeShellCommand) => {
+    const out = await executeShellCommand(
+      'for i in $(find -E . -regex ".*.robot" -type f); do cat -s $i ; done'
+    );
```
mschrage, line 53: This doesn't need to be a custom generator. You can put this script in the `script` field and then use the `postProcess` function to handle the parsing logic.

davmarek: Will do

src/robot.ts
```diff
@@ -46,6 +46,38 @@ const variablesGenerator: Fig.Generator = {
...
+        suggestions.push({
+          name: testCase,
+          description: "Test case",
+        });
+      }
+    }
+
+    return suggestions;
+  },
+};
```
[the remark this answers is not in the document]
davmarek: On it

src/robot.ts
davmarek: I changed both `tagsGenerator` and `testCasesGenerator` to use the `script` field and `postProcess` function

src/robot.ts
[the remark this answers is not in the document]
davmarek: I added comments to make it more understandable

fedeci (approved):
Thanks!

PR #1092, opened 2022-04-01
feat: add lp completion spec

teziovsky:
Feature

`lp` command is not currently supported

Autocompletion support for `lp`

**Additional info:**
lp docs found [here](https://ss64.com/osx/lp.html)


src/lp.ts
```diff
@@ -0,0 +1,121 @@
...
+        name: "name=value",
+      },
+    },
+    {
+      name: "--help",
+      description: "Show help for lp",
+    },
+  ],
+  args: {
+    template: "filepaths",
```
fedeci, line 117: If the file doesn't exist `filepaths` won't be sufficient:
```suggestion
    template: "filepaths",
    suggestCurrentToken: true,
```

fedeci (changes requested):
Thanks, this looks great! There is a small nit to fix and we are ready to merge it!

fedeci (approved):
Thanks!

PR #1094, opened 2022-04-02
feat: add rsync spec

### LINKED ISSUE #510: [rsync] completion spec
### Sanity checks

rsync

### Issue Details

Lots of arguments with the potential to re-use generators from SSH

Guergeiro:
Feature.

`rsync` is not available. Closes #510 .

Support for `rsync`.

### DIFF
--- a/src/rsync.ts
+++ b/src/rsync.ts
@@ -1,0 +1,839 @@
+const infoArgs: Fig.SingleOrArray<Fig.Arg> = [
+  { name: "BACKUP", description: "Mention files backed up" },
+  {
+    name: "COPY",
+    description: "Mention files copied locally on the receiving side",
+  },
+  { name: "DEL", description: "Mention deletions on the receiving side" },
+  {
+    name: "FLIST",
+    description: "Mention file-list receiving/sending (levels 1-2)",
+  },
+  {
+    name: "MISC",
+    description: "Mention miscellaneous information (levels 1-2)",
+  },
+  {
+    name: "MOUNT",
+    description: "Mention mounts that were found or skipped",
+  },
+  {
+    name: "NAME",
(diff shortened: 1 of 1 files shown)


src/rsync.ts
```diff
@@ -0,0 +1,831 @@
...
+      description: "Also compare destination files relative to DIR",
+      requiresEquals: true,
+      args: {
+        name: "DIR",
+        template: "folders",
+      },
+    },
+    {
+      name: "--copy-dest",
+      description: "... and include copies of unchanged files",
```
fedeci, line 534: This description seems to be missing something

src/rsync.ts
```diff
@@ -0,0 +1,831 @@
...
+    { name: "--existing", description: "Skip creating new files on receiver" },
+    {
+      name: "--ignore-existing",
+      description: "Skip updating files that already exist on receiver",
+    },
+    {
+      name: "--remove-source-files",
+      description: "Sender removes synchronized files (non-dirs)",
+    },
+    { name: "--del", description: "An alias for --delete-during" },
```
fedeci, line 340: Can you add this one as an array to the --delete-during option?

fedeci (changes requested):
Thanks!

src/rsync.ts
Guergeiro: Will do!

src/rsync.ts
Guergeiro: Yup. It's the same as `--compare-dest`. Will add it!

Guergeiro: Fixed in c45653a

src/rsync.ts
Guergeiro: Fixed in c45653a

fedeci (approved):
Thanks!

PR #1097, opened 2022-04-02
feat: add Clojure spec

KyleErhabor:
This pull request adds a spec for [Clojure CLI](https://clojure.org/guides/deps_and_cli).

There is currently no spec for Clojure CLI.

The spec mirrors the options from the help command (`clojure --help`) and [reference](https://clojure.org/reference/deps_and_cli). The CLI makes no use of subcommands and instead overloads options, so they've been added individually. For example, `-Stree` to list dependencies in a Clojure deps project and `-Sverbose` to log debug information. The CLI also makes use of aliases, which are overloaded on some options (e.g. `-X` and `-M`) and delimited by colons. For example, `-M:db:log` represents the `-M` option with the `db` and `log` aliases. Technically a user doesn't need the initial colon (i.e. they can do `-Mdb:log`), but that form is less common, with most users preferring the former.

**Additional info:**
I chose not to implement some "options" like `path` and `-` since I don't know their syntax nor when they're relevant. Additionally, I didn't implement suggestions like `-X:deps tree` since when it's appropriate to surface is a bit complex. Regardless, I believe the current spec will be useful to Clojure projects today.


fedeci:
Hey @KyleErhabor, we are working on some feature to make this spec better, I'll get back to you in a couple days! Thanks for this spec!

KyleErhabor:
@fedeci have the features been implemented?

SeparateRecords:
(not @fedeci but) Yes indeed! You can now use the `optionArgSeparators` parser directive to set the valid option argument separators, eg. `optionArgSeparators: ":"`, then on each of the options you can set `requiresSeparator: true`.

SeparateRecords:
BTW, you might need to rebase or merge master then `npm i` to get the changes in `autocomplete-types`.

fedeci:
> (not @fedeci but) Yes indeed! You can now use the `optionArgSeparators` parser directive to set the valid option argument separators, eg. `optionArgSeparators: ":"`, then on each of the options you can set `requiresSeparator: true`.

I think `requiresSeparator: ":"` along with `optionArgSeparators: ":"` must be used. `requiresSeparator: true` defaults to the equal if I am not wrong. I don't remember how I implemented it😄

KyleErhabor:
Thanks for the updates, I'll try to get something working tomorrow.

mschrage:
Hey @KyleErhabor are you still working on this?

KyleErhabor:
@mschrage Yes! I'm currently working on simplifying the codebase for options in general. From the additions mentioned and what's available in my environment locally, I've interpreted that:
- `optionArgSeparators` is a parser directive and not available on individual options.
- `requiresSeparator` is available on individual options, but not applicable when arguments are optional (as the name states).

Sadly, for my use case, the two options are likely insufficient for defining the signature of some options. For example, the `-X` option has a signature of `-X[aliases] a/fn? [kpath v]* kv-map?`, where `[aliases]` is a colon-separated list and the rest spaces. While I could use the two options to define `aliases`, it would not capture the meaning of the last three arguments. This could potentially be solved by grouping arguments together (so aliases are in one group and the last three are in another), but that is currently not a feature, and nor do I think it would be worth the effort to pursue.

Instead, I believe defining `-X` with `aliases` specified via the `displayName` and `insertValue` option values, as was done before, with the three trailing arguments treated as regular space-delimited arguments is the more practical solution. While `-X` is just an example, it's applicable to a number of other options, such as `-M`, and would help with other problems, such as whether or not to include a leading `:` for `insertValue`.

To wrap up,
- While `optionArgSeparators` and `requiresSeparator` are nice, they're probably not valid for my situation.
- Relying on `displayName` and `insertValue` would be better.
- Dealing with the leading `:` is a bit annoying since it results in two separate options (one for when the user wants to add aliases and one for when the user doesn't). Would leaving the colon out and expecting the user to insert it be a better compromise?

This is how I've interpreted my situation. Is there something I misunderstood?

fedeci:
Are `kv-path` and `kv-map` key-value arguments?

KyleErhabor:
Generators would probably not work since `kv-path` (accepting key-value arguments) is dynamic and meant to set (create new or override existing) config attributes defined in edn (Clojure's version of JSON). Though, I will try to get the rest of the options working. 🙂

fedeci:
Thanks a lot for the effort @KyleErhabor!

fedeci (approved):
Definitely better! Thanks!

PR #1107, opened 2022-04-04
feat(brew): add `reinstall` subcommand

amorfinv:
I added ```brew reinstall``` because I use it pretty often and it was missing. All it does is install and reinstall. Similar to ```brew upgrade``` but simpler.

See here for more info

https://docs.brew.sh/Manpage#reinstall-options-formulacask-

### DIFF
--- a/src/brew.ts
+++ b/src/brew.ts
@@ -1155,11 +1155,13 @@ const completionSpec: Fig.Spec = {
           name: "--binaries",
           description:
             "Disable/enable linking of helper executables (default: enabled)",
+          exclusiveOn: ["--no-binaries"],
         },
         {
           name: "--no-binaries",
           description:
             "Disable/enable linking of helper executables (default: enabled)",
+          exclusiveOn: ["--binaries"],
         },
         {
           name: "--require-sha",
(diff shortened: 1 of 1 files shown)


src/brew.ts
```diff
@@ -1100,6 +1100,92 @@ const completionSpec: Fig.Spec = {
...
+        },
+        {
+          name: "--quarantine",
+          description:
+            "Disable/enable quarantining of downloads (default: enabled)",
+        },
+        {
+          name: "--no-quarantine",
+          description:
+            "Disable/enable quarantining of downloads (default: enabled)",
```
fedeci, line 1176: These options are mutually [exclusiveOn](https://fig.io/docs/reference/option#exclusiveon).

src/brew.ts
```diff
@@ -1100,6 +1100,92 @@ const completionSpec: Fig.Spec = {
...
+        {
+          name: "--binaries",
+          description:
+            "Disable/enable linking of helper executables (default: enabled)",
+        },
+        {
+          name: "--no-binaries",
+          description:
+            "Disable/enable linking of helper executables (default: enabled)",
+        },
```
fedeci, line 1163: These too

fedeci (changes requested):
Thanks for this first contribution!

PR #1112, opened 2022-04-06
feat(git): add config key suggestions

SeparateRecords:
Scraped from the manual page. Descriptions need to be cleaned up still.

Because of Fig's UI and {cursor} support, this is actually **better than fish completions** for git-config (which are often incorrect and slow down the whole shell)


fedeci (approved):
This is difficult to review😄, I trust the CI!

PR #1114, opened 2022-04-07
feat(git): support `git-xyz` commands on $PATH

SeparateRecords:
Shows commands found on your $PATH  in the list of subcommands. These commands can also be supported explicitly in the `optionalCommands` array, to add options and arguments. Currently this only supports `git-open`, `git-recent`, and `git-flow` but it's easy to add more.


fedeci (approved):
👍

PR #1116, opened 2022-04-07
fix(robot): test cases generator regex

davmarek:
Bug fix

Generator for test cases names doesn't work in some situations

Generator for test cases names works


fedeci (approved):
Thanks!

PR #1119, opened 2022-04-08
feat(fig): add generators for `fig plugins`

SeparateRecords:
Screen Shot 2022-04-08 at 1 53 36 pm
Screen Shot 2022-04-08 at 1 55 15 pm


mschrage:
😍 Love this @SeparateRecords

PR #1121, opened 2022-04-08
feat(git): add user config generator

SeparateRecords:
This makes 3 changes:
1. `git config` suggests the current token, since all config keys are valid.
2. Keys from your actual git config (eg. aliases, remotes) are shown
3. `--get-regexp` takes an argument now
Screen Shot 2022-04-08 at 9 27 03 pm


mschrage:
@SeparateRecords you are a completion spec wizard at this point!

mschrage (approved):

PR #1128, opened 2022-04-08
feat(aws/eks): update spec about kubernetes version

yysu:
update aws eks kubernetes version according to [aws eks doc](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html)

Some kubernetes versions are deprecated. Update them to the current support one.

**Additional info:**


fedeci (approved):
Thanks! It would be great if these version suggestions could be generated using `Fig.Generator` so they stay updated

PR #1130, opened 2022-04-09
Improve target description

Clement-Jean:

Instead of always having `Make target` as description, we can now have some documentation for our rules.

Shows `Make target` as description for all suggestions.

- target defined like this: `all: [rule]+ ## This builds everything`, will have `This builds everything` as description
- if the `##` comment is not define it will show `Make target` as description

### DIFF
--- a/src/make.ts
+++ b/src/make.ts
@@ -1,12 +1,15 @@
 const listTargets: Fig.Generator = {
-  script: `make -qp | awk -F':' '/^[a-zA-Z0-9][^$#\\/\\t=]*:([^=]|$)/ {split($1,A,/ /);for(i in A)print A[i]}' | sort -u`,
+  script: `grep -E '^[a-zA-Z_-]+:.*?(## .*)?$' Makefile | awk 'BEGIN {FS = ":.*?## "}; {split($0,a,\/:\/); split($0,b,\/##\/); print a[1] b[2]}' | sort -u`,
   postProcess: function (out) {
     const lines = out.split("\n");
     const targets = [];
     for (let i = 1; i < lines.length; i++) {
+      let first_space = lines[i].indexOf(' ');
+      let desc = first_space == -1 ? "Make target" : lines[i].substring(first_space + 1).trim();
+
       targets.push({
-        name: lines[i].trim(),
-        description: "Make target",
+        name: lines[i].substring(0, first_space == -1 ? lines[i].length : first_space).trim(),
+        description: desc,
         icon: "🎯",
         priority: 80,
       });


Clement-Jean:
👍

fedeci:
Hey @Clement-Jean, thanks for this first contribution! Currently there is no standard for descriptions of makefile targets and I am a little bit hesitant in approving this change because of that. Some people may add comments above the target name, use just one `#` or use more than two.
/cc @SeparateRecords @mschrage WDYT?

Clement-Jean:
@fedeci totally agree on that ! There is no convention. I was just trying to find a way to make the description more meaningful. I would understand that you reject this PR due to the mentioned concern.

SeparateRecords:
> There is no convention. I was just trying to find a way to make the description more meaningful

There's no standard but from a _very cursory look_ it seems like most people use comments in a similar way. I think we should actually make this more general! For example, [Python's makefile](https://github.com/python/cpython/blob/defbbd68f7f68f4edb3a6b256f26e0532727b3da/Makefile.pre.in#L704-L718) uses multiline comments above the target, as do many makefiles in the google org.

I think adding this would be really useful. Worst case: the "description" is unrelated. Best case: people are delighted to see their target descriptions _used as descriptions_. If I'm missing something, please tell me! But overall I like this.

Clement-Jean:
The multiline comment on top of the run is quite tricky and can be pretty long, I was thinking of the ## at the end of the rule as a 'summary'. Generally I use that as a usage guide, eg:

```
about: ## Display info related to the build
	@echo "OS: example"

help: ## Show this help
	grep -E '^[a-zA-Z_-]+:.*?\#\# .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?\#\# "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```

and then when I do a `make help`, I have something like:

```
about     Display info related to the build
help       Show this help
```

fedeci:
@Clement-Jean that's really cool, you should propose that to the `make` project itself. I am just concerned that writing comments next to the target name is not used much and adding it above may actually be more useful in "predicting user behaviour".

SeparateRecords:
This regex looks like it's correct to me. It matches leading comments, ignores .PHONY rules that immediately follow it, and matches recipe names and inline comments.
```regexp
((?:^#.*\n)*)(?:^\.PHONY:.*\n)?(^\S*):.*?(?:\s#\s*(.*))?\n\t
```

As an example, for this text:
```
# Create build directory and generate the sysconfig build-time data there.
# pybuilddir.txt contains the name of the build dir and is used for
# sys.path fixup -- see Modules/getpath.c.
# Since this step runs before shared modules are built, try to avoid bootstrap
# problems by creating a dummy pybuilddir.txt just to allow interpreter
# initialization to succeed.  It will be overwritten by generate-posix-vars
# or removed in case of failure.
pybuilddir.txt: $(BUILDPYTHON)  # inline comment
	@echo "none" > ./pybuilddir.txt
```

* $1 is the leading comment (hashes included, can easily be stripped out in JS)
* $2 is the recipe name
* $3 is the inline comment, leading hash not included

SeparateRecords:
```ts
const listTargets: Fig.Generator = {
  script: `cat [Mm]akefile`,
  postProcess: function (out) {
    const matches = out.matchAll(/((?:^#.*\n)*)(?:^\.PHONY:.*\n)?(^\S*):.*?(?:\s#+\s*(.*))?\n\t/gm);
    const targets = [];
    for (const match of matches) {
      const [leadingComment, target, inlineComment] = match;
      const description = inlineComment
        ? inlineComment.trim()
        : leadingComment
        ? leadingComment.replaceAll(/^#+\s*/gm)
        : undefined;

      targets.push({
        name: target.trim(),
        description,
        icon: "🎯",
        priority: 80,
      });
    }
    return targets;
  },
};
```

Clement-Jean:
@SeparateRecords Question, do you really need the `\n\t` at the end? Because we could also have:

```
# Another
another: both
```

and this is not matching for now

Clement-Jean:
This also doesn't work for now:

```
# Another
another: both # another one
```

This should print `another one`

SeparateRecords:
Ahhh good catch. Totally forgot about this case. I'll look into this.

SeparateRecords:
Tested this on the python and linux makefiles, looks like it works!
```regex
((?:^#.*\n)*)(?:^\.PHONY:.*\n)?(^\S*?):.*?(?:\s#+\s*(.*))?$
```

This change will also match special targets (I didn't know about most of these, damn make is cool): https://www.gnu.org/software/make/manual/html_node/Special-Targets.html

You could make a set of these outside the for-loop, eg `new Set([".PHONY", ".SUFFIXES", ".DEFAULT", ...])`, then at the top of the loop add `if (specialTargets.has(target)) continue;`

Clement-Jean:
Sorry for late response, I added the changes. Let me know

SeparateRecords:
Fantastic!! Would you mind adding the other special targets to that set too? Scraped from that page:

```
['.PHONY', '.SUFFIXES', '.DEFAULT', '.PRECIOUS', '.INTERMEDIATE', '.SECONDARY', '.SECONDEXPANSION', '.DELETE_ON_ERROR', '.IGNORE', '.LOW_RESOLUTION_TIME', '.SILENT', '.EXPORT_ALL_VARIABLES', '.NOTPARALLEL', '.ONESHELL', '.POSIX']
```

fedeci (approved):
Thanks!

PR #1138, opened 2022-04-10
fix(git): correct `log` option

mbd-s:
Bug fix

When the user types `git log `, fig suggests the autocompletion `--p`, which results in the error `fatal: unrecognized argument: --p`. Per the [git docs](https://git-scm.com/docs/git-log#_diff_formatting), the available options for `git log` to generate a patch are `-p`, `-u`, and `--patch`.

To display the correct options.


mschrage (approved):
Thanks for fixing this!

PR #1140, opened 2022-04-11
feat: add `dtm` spec

imxw:
Feature

`dtm` command are not currently supported

Autocompletion support for `dtm`

**Additional info:**
`dtm` project found [here](https://github.com/devstream-io/devstream)

### DIFF
--- a/src/dtm.ts
+++ b/src/dtm.ts
@@ -1,0 +1,284 @@
+const cmds: Fig.Generator = {
+  script: `dtm -h | sed -e '/Available Commands:/,/Flags:/!d' | grep -Ev "Available Commands:|Flags:|^$" | awk '{print $1}'`,
+  postProcess: (output) => {
+    if (output.startsWith("fatal:")) {
+      return [];
+    }
+    return output.split("\n").map((cmd) => {
+      return { name: cmd.trim(), description: "Command" };
+    });
+  },
+};
+
+const plugins: Fig.Generator = {
+  script: "dtm list plugins",
+  postProcess: (output) => {
+    if (output.startsWith("fatal:")) {
+      return [];
+    }
+    return output.split("\n").map((plugin) => {
+      return { name: plugin.trim(), description: "Plugin" };
+    });
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "dtm",
+  description: "DevStream is an open-source DevOps toolchain manager",
+  subcommands: [
+    {
+      name: "apply",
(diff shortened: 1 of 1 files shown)


src/dtm.ts
```diff
@@ -0,0 +1,284 @@
...
+        },
+      ],
+    },
+
+    {
+      name: "help",
+      description: "Help about any command",
+      args: {
+        name: "command",
+        generators: cmds,
```
fedeci, line 225:
```suggestion
        templates: "help",
```

src/dtm.ts
```diff
@@ -0,0 +1,284 @@
...
+      description:
+        "Verify DevOps tools according to DevStream config file and state",
+      options: [
+        {
+          name: ["-f", "--config-file"],
+          description: 'Config file (default "config.yaml")',
+          args: {
+            name: "file",
+            template: "filepaths",
+          },
```
fedeci, line 265: Suggestion: if this `--config` only accepts certain file extensions you can use the `@fig/autocomplete-generators` package.

fedeci (changes requested):
Thanks, there are a couple nits but overall it looks good!

src/dtm.ts
imxw: Thanks for your help

mschrage (approved):
This looks great. Thanks so much for contributing @imxw!

src/dtm.ts
imxw: Thanks for your help!

imxw:
> Looks like these flags are global:
> 
> ```
>   -f, --config-file string   config file (default "config.yaml")
>   -d, --debug                debug level log
>   -p, --plugin-dir string    plugins directory (default ".devstream")
>   -y, --yes                  apply/delete directly without confirmation
> ```
> 
> Could you add these to the top-level `options` array, and give them all `isPersistent: true`?
> 
> I think this would be the last change! Looks good 🙂

Thanks for your rigor!  With more subcommands, these flags can no longer be used globally. We will move them to the appropriate subcommands in the next version.

imxw:
> @imxw would you be interested in having automatic updates and versioning for this completion spec? I've seen you are using cobra in your CLI and we provide a cobra integration for autogenerating specs. I can open a PR in the next days, WDYT? I am merging this one in the meanwhile, thanks again!

Yes, I've noticed the cobra integration. Once I remove some unnecessary global flags from `dtm`, I'll use it.

PR #1143, opened 2022-04-12
feat: add `hexo` completion spec

patrick330602:

feature: autocomplete for `hexo-cli` for [Hexo](https://hexo.io)

does not provide autocomplete for hexo-cli

provides autocomplete for hexo-cli commands

### DIFF
--- a/src/hexo.ts
+++ b/src/hexo.ts
@@ -276,6 +276,7 @@ const completionSpec: Fig.Spec = {
       name: "--config",
       description:
         "Uses a custom config file (instead of `_config.yml`). Also accepts a comma-separated list (no spaces) of JSON or YAML config files that will combine the files into a single `_multiconfig.yml`",
+      isPersistent: true,
       args: {
         name: "config file",
         template: "filepaths",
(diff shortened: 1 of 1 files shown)


src/hexo.ts
```diff
@@ -0,0 +1,310 @@
...
+  script: "hexo list post | grep -E ^Draft",
+  postProcess: (out) => {
+    return out.split("\n").map(function (file) {
+      const title = file
+        .replace(/^.*   *_drafts\//g, "")
+        .replace(/  *(.*)$/g, "")
+        .replace(/.md$/g, "");
+      return {
+        name: title,
+        description: "draft for " + title,
```
SeparateRecords, line 11: Just a quick case change 🙂 (we try to keep to "conventional descriptions", eg. upper case first letters, no periods at the end)
```suggestion
        description: "Draft for " + title,
```

SeparateRecords (changes requested):
Amazing!! Just one quick change and then I think it's ready to merge!

src/hexo.ts
patrick330602: Gotcha, and updated the description

src/hexo.ts
```diff
@@ -0,0 +1,310 @@
...
+    },
+    {
+      name: "--cwd",
+      description: "Customizes the path of current working directory",
+      args: {
+        name: "directory",
+        template: "folders",
+      },
+    },
+  ],
```
SeparateRecords, line 308: One more thing, these look like they're global options - could you add `isPersistent` to them? Sorry for that 😅

patrick330602: Gotcha and added

SeparateRecords (approved):

SeparateRecords:
Thanks so much for adding this! Looks good 👍

PR #1148, opened 2022-04-14
feat(fastly): use diff-versioned spec

fedeci:
/cc @kailan It is working now! Any chance we can work to get `push-to-autocomplete-action` working for fastly?


mschrage:
recheck

brendanfalk:
Too big of a change and we have likely had so many changes since. I am going to close for now. We will work on a better integration with fastly in the future.

PR #1156, opened 2022-04-19
feat: add shell syntax "commands"

SeparateRecords:
Fixes https://github.com/withfig/fig/issues/1317
Also adds support for basic things like `if` (and `then`/`elif`/`else`), `for`, etc.

### DIFF
--- a/src/where.ts
+++ b/src/where.ts
@@ -1,5 +1,5 @@
 const completionSpec: Fig.Spec = {
-  name: "whence",
+  name: "where",
   description: "For each name, indicate how it should be interpreted",
   args: {
     name: "names",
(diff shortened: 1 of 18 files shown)


src/where.ts
```diff
@@ -0,0 +1,41 @@
+const completionSpec: Fig.Spec = {
+  name: "whence",
+  description: "For each name, indicate how it should be interpreted",
```
mschrage, line 3: Seems like `name` is incorrect?

src/which.ts
```diff
@@ -0,0 +1,22 @@
+const completionSpec: Fig.Spec = {
+  name: "whence",
+  description: "For each name, indicate how it should be interpreted",
```
mschrage, line 3: Same thing here! File is `which.ts` but name is `whence`

src/where.ts
SeparateRecords: Fixed!

fedeci (approved):
🚀

PR #1161, opened 2022-04-20
feat(scp): add hosts and filepaths completions

SeparateRecords:
Currently the signature for `scp` is wrong. Like `cp`, it can also take _multiple_ sources and copy to a single destination.


fedeci (approved):
LGTM!

PR #1166, opened 2022-04-21
feat: hide shell syntax commands

SeparateRecords:
This should be merged before https://github.com/withfig/public-site-nextjs/pull/127

These commands will be hidden from the manual, but still accessible.

Sidenote: It's so nice to be back on my _actual_ dev machine instead of my girlfriend's laptop. Back to vim!! 🚀


fedeci (approved):
👍

PR #1171, opened 2022-04-22
feat: add tldr spec

AMATEURTOSS:
docs update

add spec

add spec

**Additional info:**
about tldr: [https://github.com/tldr-pages/tldr](https://github.com/tldr-pages/tldr)


src/tldr.ts
```diff
@@ -0,0 +1,52 @@
...
+      })
+      .map((line) => {
+        return {
+          name: line.slice(0, -3),
+          description: "Tldr page",
+          icon: "fig://icon?type=string",
+        };
+      });
+  },
+};
```
SeparateRecords, line 35: Love this! I think this might have issues with other `tldr` implementations though, like tealdeer (which is my personal choice).

Both `tldr` implementations have a `--list` option which lists all cached pages, is it possible to use that? Tealdeer outputs `\n`-separated names, the JS cli outputs `, ` separated names.

src/tldr.ts
```diff
@@ -0,0 +1,52 @@
...
+  options: [
+    {
+      name: "--update",
+      description: "Download the latest pages and generate search index",
+    },
+    {
+      name: "--clear-cache",
+      description: "Delete the entire local cache",
+    },
+  ],
```
SeparateRecords, line 50: iirc tldr has more options than this - is this only the common options between implementations?

SeparateRecords (commented):
Looking good! Can't believe we didn't have tldr already, thanks for adding it.

Left a couple of questions - not dealbreakers for merging, just wanted to understand your reasoning :)

src/tldr.ts
AMATEURTOSS: i didn't know there is any other client for tldr.

Is there a tldr client that does not use .tldr/cache as the path to store the cache?
Otherwise, I think the current method is not bad.

src/tldr.ts
AMATEURTOSS: I'll add the rest of the options.

src/tldr.ts
SeparateRecords: There's actually quite a few implementations! Check out this page: https://github.com/tldr-pages/tldr/wiki/tldr-pages-clients

I'm in favor of merging this because it's the official CLI. If I add the tealdeer spec, I'll figure out some way to show the correct suggestions (probably reading the hashbang, `head -n1 $(command -v tldr)`)

src/tldr.ts
```diff
@@ -0,0 +1,165 @@
...
+    const sunos = `~/.tldr/cache/pages/sunos/`;
+    const windows = `~/.tldr/cache/pages/windows/`;
+    return `{
+      ls -l ${android} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9 &
+      ls -l ${common} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9 &
+      ls -l ${linux} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9 &
+      ls -l ${osx} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9 &
+      ls -l ${sunos} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9 &
+      ls -l ${windows} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9;
+    } | cat`;
```
SeparateRecords, line 16: Can this be simplified? You might be able to do `ls -l1 ~/.tldr/cache/pages/{android,common,linux,osx,sunos,windows} 2>/dev/null | sort -u | tr -s ' ' | cut -d ' ' -f 9 | cat | head` then remove blank lines and lines starting with `/` in JS

src/tldr.ts
```diff
@@ -0,0 +1,150 @@
+const android = `~/.tldr/cache/pages/android/`;
+const common = `~/.tldr/cache/pages/common/`;
+const linux = `~/.tldr/cache/pages/linux/`;
+const osx = `~/.tldr/cache/pages/osx/`;
+const sunos = `~/.tldr/cache/pages/sunos/`;
+const windows = `~/.tldr/cache/pages/windows/`;
+
+const wholeTldrPages: Fig.Generator = {
+  script: () => {
+    return `ls -l ${android} ${common} ${linux} ${osx} ${sunos} ${windows} 2>/dev/null | tr -s ' ' | cut -d ' ' -f 9 | sort -u`;
```
mschrage, line 10: Can this parsing logic be moved to JS (then we don't need to worry about aliases or incompatibilities between different flavors of coreutils?

mschrage (approved):

PR #1189, opened 2022-04-25
feat: add qodana spec

bxb100:

feature

add spec

add spec

**Additional info:**

qodana cli: https://github.com/JetBrains/qodana-cli


src/qodana.ts
```diff
@@ -0,0 +1,348 @@
...
+      name: ["-h", "--help"],
+      description: "Show help page for command",
+      isPersistent: true,
+      priority: 1,
+      exclusiveOn: exclusives,
+    },
+    {
+      name: ["-v", "--version"],
+      description: "Version for qodana",
+      exclusiveOn: exclusives,
```
fedeci, line 80: This exclusive won't have any effect because none of the options listed are options of the `quodana` command itself but are options of its subcommands.
```suggestion
```

src/qodana.ts
```diff
@@ -0,0 +1,348 @@
...
+          name: "--script",
+          description: 'Override the run scenario (default "default")',
+          args: {
+            name: "string",
+          },
+        },
+        {
+          name: "--send-report",
+          description:
+            "Send the inspection report to Qodana Cloud, requires the '--token' option to be specified",
```
fedeci, line 248:
```suggestion
            "Send the inspection report to Qodana Cloud, requires the '--token' option to be specified",
          dependsOn: ["--token"],
```

src/qodana.ts
```diff
@@ -0,0 +1,348 @@
...
+          name: ["-u", "--user"],
+          description:
+            "User to run Qodana container as. Please specify user id – '$UID' or user id and group id $(id -u):$(id -g). Use 'root' to run as the root user (default: the current user)",
+          args: {
+            name: "string",
+          },
+        },
+        {
+          name: ["-v", "--volume"],
+          description:
```
fedeci, line 285:
```suggestion
          isRepeatable: true,
          description:
```

src/qodana.ts
```diff
@@ -0,0 +1,348 @@
...
+            "Clear the local Qodana cache before running the analysis",
+        },
+        {
+          name: "--disable-sanity",
+          description:
+            "Skip running the inspections configured by the sanity profile",
+        },
+        {
+          name: ["-e", "--env"],
```
fedeci, line 151:
```suggestion
          name: ["-e", "--env"],
          isRepeatable: true,
```

src/qodana.ts
```diff
@@ -0,0 +1,348 @@
...
+  "-d",
+  "--dir-only",
+  "-p",
+  "--port",
+  "-i",
+  "--project-dir",
+  "-r",
+  "--report-dir",
+  "-f",
+  "--sarif-file",
```
fedeci, line 63: nit:
`exclusiveOn` does not require both the long and short options to be mentioned, I would remove "duplicates" because of maintainability.
e.g. `--help` is enough and `-h` can be omitted

fedeci (changes requested):
I left a few comments, thanks for opening this PR!

src/qodana.ts
bxb100: Thank review, is it any way to avoid type `option + space` not showing other subcommands or options behind?

src/qodana.ts
fedeci: Likely `isPersistent` does not work correctly with `exclusiveOn`, I'll fix it!

src/qodana.ts
fedeci: I don't think anyone will ever add commands or other flags after `--version`, but is seems like a good addition for the future!

fedeci (approved):
Thanks!

PR #1190, opened 2022-04-25
feat(fig): update spec

mschrage:


mschrage:
@fedeci seems like the linting rules aren't applied by the automated GH action?

fedeci:
They are applied but they are not yet configured for diff based versioning and it will be pretty though to get them working easily. It's WIP however!

PR #1206, opened 2022-05-04
feat: add mikro-orm completion spec

noftaly:
feature: added Mikro-ORM completion Spec

No Spec

There is a spec

**Additional info:**
MikroORM is a popular NodeJS ORM, and has a CLI to manage seeds, migrations etc.
This PR adds its spec. It has also been added to `npx`.

All the info here was taken from its CLI on version `5.1.3` (latest to this day)


fedeci (approved):
Thanks!

PR #1202, opened 2022-05-03
feat: add flyctl spec

BogDAAAMN:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

Added completion specs for [flyctl](https://fly.io/docs/flyctl/) CLI, the command-line interface for fly.io! 

It is auto-generated with the [`github.com/withfig/autocomplete-tools/packages/cobra`](https://github.com/withfig/autocomplete-tools/blob/main/packages/cobra/README.md) package. With the help of amazing @davidalpert we found a work-around and made the package to work with the current [`flyctl` cobra implementation](https://github.com/superfly/flyctl). You can see the implementation here https://github.com/BogDAAAMN/flyctl/commit/c0f338a6f3c59acc8286a8c82836fc2e73124b6f.

I am not familiar enough with Go to propose a better implementation or perhaps a GitHub Action, although I would love to try better! Anyhow I'm pinging the fly.io folks here, maybe there's something they might consider it ever. I'd love to try more. @michaeldwan @codepope @ThisIsRahmat @rugwirobaker

**Additional info:**

Added the following generators:
- `template: "filepaths"` for all the `config` arguments
- Generate the apps and orgs using the `flyctl list apps` command. It changes the icon based on `status`
  https://github.com/withfig/autocomplete/blob/72c5328f330775ce3e4954d918e26eb0b147b63a/src/flyctl.ts#L9-L26

PR #1207, opened 2022-05-05
feat: add completion spec for `fly` (alias of `flyctl`)

BogDAAAMN:

In addition to #1202, adding completion spec for `fly` command, alias of `flyctl`. Suggested by @jsierles. 

I propose we leave it like this until `flyctl` transitions to `fly`.

> Note: Most installations of `flyctl` also alias `flyctl` to `fly` as a command name and **this will become the default name in the future**. During the transition, note that where you see `flyctl` as a command it can be replaced with `fly`.

Source: https://github.com/superfly/flyctl

### DIFF
--- a/src/fly.ts
+++ b/src/fly.ts
@@ -1,0 +1,14 @@
+// Completion spec for fly, alias to the flyctl CLI.
+// See completion spec in the ./flyctl.ts file
+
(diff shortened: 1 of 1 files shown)


fedeci (approved):
Thanks!

PR #1208, opened 2022-05-05
feat: add mosh completion spec

brandonleon:
feature, new completion spec for mosh.

missing completion for mosh, mobile shell

new completion spec

**Additional info:**

### DIFF
--- a/src/mosh.ts
+++ b/src/mosh.ts
@@ -22,20 +22,20 @@ const completionSpec: Fig.Spec = {
       description: 'Mosh server on remote machine (default: "mosh-server")',
     },
     {
-      name: "--predict=adaptive",
-      description: "Local echo for slower links [default]",
-    },
-    {
-      name: ["--predict=always", "-a"],
-      description: "Use local echo even on fast links",
-    },
-    {
-      name: ["--predict=never", "-n"],
-      description: "Never use local echo",
-    },
-    {
-      name: "--predict=experimental",
-      description: "Aggressively echo even when incorrect",
+      name: "--predict",
+      description: "Local echo options",
+      sugestions: [
+        {
+          name: "adaptive",
+          description: "Local echo for slower links [default]",
+        },
+        { name: "always", description: "Use local echo even on fast links" },
+        { name: "never", description: "Never use local echo" },
+        {
+          name: "experimental",
+          description: "Aggressively echo even when incorrect",
+        },
+      ],
     },
     {
       name: "-4",
--- a/src/mosh.ts
+++ b/src/mosh.ts
@@ -24,18 +24,24 @@ const completionSpec: Fig.Spec = {
     {
       name: "--predict",
       description: "Local echo options",
-      sugestions: [
-        {
-          name: "adaptive",
-          description: "Local echo for slower links [default]",
-        },
-        { name: "always", description: "Use local echo even on fast links" },
-        { name: "never", description: "Never use local echo" },
-        {
-          name: "experimental",
-          description: "Aggressively echo even when incorrect",
-        },
-      ],
+      requiresEqual: true,
+      args: {
+        suggestions: [
+          {
+            name: "adaptive",
+            description: "Local echo for slower links [default]",
+          },
+          { name: "always", description: "Use local echo even on fast links" },
+          {
+            name: "never",
+            description: "Never use local echo",
+          },
+          {
+            name: "experimental",
+            description: "Aggressively echo even when incorrect",
+          },
+        ],
+      },
     },
     {
       name: "-4",
@@ -48,9 +54,8 @@ const completionSpec: Fig.Spec = {
     {
       name: "--family",
       description: "Network Type",
+      requiresEqual: true,
       args: {
-        name: "auto",
-        description: "Autodetect network type for single-family hosts only",
         suggestions: [
           { name: "inet", description: "Use IPv4 only" },
           { name: "ine6", description: "Use IPv6 only" },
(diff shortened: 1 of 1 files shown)


src/mosh.ts
```diff
@@ -0,0 +1,134 @@
...
+      description: "Show help for mosh",
+    },
+    {
+      name: ["--client"],
+      description: "mosh client on local machine (default: \"mosh-client\")",
+    },
+    {
+      name: ["--server"],
+      description: "mosh server on remote machine (default: \"mosh-server\")",
+    },
```
mschrage, line 57: Do these options take an argument?

src/mosh.ts
```diff
@@ -0,0 +1,134 @@
...
+      description: "do not send terminal initialization string",
+    },
+    {
+      name: ["--local"],
+      description: "run mosh-server locally without using ssh",
+    },
+    {
+      name: ["--experimental-remote-ip="],
+      description: "{local|remote|proxy}, select the method for discovering the remote IP address to use for mosh (default: \"proxy\")",
+    },
```
fedeci, line 125:
```suggestion
      name: "--experimental-remote-ip",
      description: "Select the method for discovering the remote IP address to use for mosh",
      requiresSeparator: true,
      args: {
        name: "method",
        suggestions: ["local", "remote", "proxy"],
        default: "proxy",
      },
    },
```

src/mosh.ts
```diff
@@ -0,0 +1,134 @@
...
+    {
+      name: ["--port", "-p"],
+      description: "server-side UDP port or range, (No effect on server-side SSH port)",
+    },
+    {
+      name: ["--bind-server="],
+      description: "{ssh|any|IP}, ask the server to reply from an IP address, (default: \"ssh\")",
+    },
+    {
+      name: ["--ssh="],
```
fedeci, line 107:
```suggestion
      name: "--ssh",
      requiresSeparator: true,
      args: {
        name: "command",
        isCommand: true,
      },
```

src/mosh.ts
```diff
@@ -0,0 +1,134 @@
...
+      description: "try all network types",
+    },
+    {
+      name: ["--family=prefer-inet"],
+      description: "use all network types, but try IPv4 first [default]",
+    },
+    {
+      name: ["--family=prefer-inet6"],
+      description: "use all network types, but try IPv6 first",
+    },
```
fedeci, line 97: I wonder if it's better to keep these options separate or to merge them and then duplicate `-4` and `-6`.
/cc @SeparateRecords @mschrage WDYT?

src/mosh.ts
```diff
@@ -0,0 +1,134 @@
...
+    return out
+      .split("\n")
+      .filter((line) => line.trim().startsWith("Host ") && !line.includes("*"))
+      .map((host) => ({
+        name: host.split(" ")[1],
+        description: "SSH host",
+        priority: 90,
+      }));
+  },
+};
```
fedeci, line 35: You can avoid to duplicate these generators and import them from the ssh spec:
```suggestion
import { knownHosts, configHosts } from "./ssh"
```

fedeci (changes requested):
Thanks for this spec 🚀! I left a couple comments. Can you also run `npm run lint:fix` before pushing so that we can get the CI not to fail?

src/mosh.ts
brandonleon: Agreed

src/mosh.ts
brandonleon: I missed this feature. I like this method much better.

src/mosh.ts
brandonleon: I considered that as well. I decided to stick to how the --help had it laid out, but that's not to say collapsing it all into -4 and -6 isn't a better option. 

I'm actually leaning toward your suggestion of making it simpler.

mschrage: I think these options should be split out. Any option that has a `=` character in the name is really is the combination of _an option_ and _an argument_.

fedeci: Let's merge them in a single option + argument then!

brandonleon: I'll get that done this evening.

src/mosh.ts
```diff
@@ -0,0 +1,124 @@
...
+      name: ["--predict=always", "-a"],
+      description: "Use local echo even on fast links",
+    },
+    {
+      name: ["--predict=never", "-n"],
+      description: "Never use local echo",
+    },
+    {
+      name: "--predict=experimental",
+      description: "Aggressively echo even when incorrect",
```
fedeci, line 38: Could you apply the same changes of `--family` to `--predict` too?

src/mosh.ts
```diff
@@ -0,0 +1,124 @@
...
+      name: "-4",
+      description: "Use IPv4 only",
+    },
+    {
+      name: "-6",
+      description: "Use IPv6 only",
+    },
+    {
+      name: "--family",
+      description: "Network Type",
```
fedeci, line 50: If an equals is required between the option and the argument you should add this option:
```suggestion
      description: "Network Type",
      requiresSeparator: true,
```

src/mosh.ts
```diff
@@ -0,0 +1,124 @@
...
+      args: {
+        name: "auto",
+        description: "Autodetect network type for single-family hosts only",
```
fedeci, line 53:
```suggestion
        name: "type",
```

fedeci (changes requested):
Great, a couple more changes and we are done!

brandonleon:
Will fix the last few errors tomorrow.

fedeci (approved):
Thanks!

PR #1209, opened 2022-05-07
feat(mob): add subcommands aliases

thmuch:
Updates the "mob" completion spec with frequently used, but undocumented abbreviations of some subcommands.

### DIFF
--- a/src/mob.ts
+++ b/src/mob.ts
@@ -3,7 +3,7 @@ const completionSpec: Fig.Spec = {
   description: "Fast git handover for remote collaboration with mob.sh",
   subcommands: [
     {
-      name: "start",
+      name: ["start", "s"],
       description: "Start session from base branch in wip branch",
       options: [
         {
@@ -25,7 +25,7 @@ const completionSpec: Fig.Spec = {
       },
     },
     {
-      name: "next",
+      name: ["next", "n"],
       description: "Handover changes in wip branch to next person",
       options: [
         {
@@ -46,7 +46,7 @@ const completionSpec: Fig.Spec = {
       ],
     },
     {
-      name: "done",
+      name: ["done", "d"],
       description: "Squashes all changes in wip branch to index in base branch",
       options: [
         {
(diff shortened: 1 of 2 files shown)


src/mob.ts
```diff
@@ -105,19 +105,19 @@ const completionSpec: Fig.Spec = {
...
     },
     {
-      name: "version",
+      name: ["version", "--version", "-v"],
       description: "Show the version",
     },
     {
-      name: "help",
+      name: ["help", "--help", "-h"],
       description: "Show help",
```
fedeci, line 121: This needs to be separated to an option and a subcommand.

fedeci (changes requested):
Thanks! I left a comment to resolve before merging this.

src/mob.ts
thmuch: @fedeci Thanks for your review!

Actually, these "options" aren't really options, but just aliases for the subcommands. I had a look at the [source code](https://github.com/remotemobprogramming/mob/blob/cf479c3018dcabb3c05f09f1a3f727c314113484/mob.go#L775) of the "mob" tool: `--version`, `-v`, `--help`and `-h` cannot be used as options, they're just there for convenience.

Should I better remove them from the completion spec (if this is a problem for Fig in any way)?

fedeci: It makes sense they cannot be used as options as it makes no sense to combine them with anything else and I don't think anyone will ever try to use them along with other options. We may leave them as subcommands if they are correctly suggested right now, but I am concerned for future changes to the engine that may lead to errors with them, so that's why I prefer to have them under options.
Probably a more appropriate solution would be to create a property that only suggest `--help` and `--version` when nothing else has been inserted, I'll work on that in the future.

thmuch: Thinking about it... These pseudo options are there just for convenience if someone tries the usual help and version options.

Since we have Fig suggest the `version` and `help`subcommands, I think I can safely remove these pseudo options (which aren't documented anyway). I'll do that tomorrow.

Later on (in a separate PR) I'll add them as options, once the new property is available.

For this PR, having the one-letter abbreviations of the other subcommands in far more important.

fedeci: 👍

fedeci (approved):
Thanks!

PR #1210, opened 2022-05-07
feat(git): add `git-daemon` subcommand

j-f1:

feature?

`git` autocompletions do not include the `daemon` command

`git daemon` is now supported*

**Additional info:**

~I’m not sure how to implement this, but several `git-daemon` options (including `--base-path`) require that they be specified as `--base-path=value` rather than `--base-path value`. Is there a way to tell Fig to insert an `=` instead of a space?~ That’s the `requiresEquals` option, will push an update shortly.

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -8065,6 +8065,7 @@ const completionSpec: Fig.Spec = {
           name: "--base-path",
           description:
             "Remap all the path requests as relative to the given path",
+          requiresEquals: true,
           args: { name: "path", template: "folders" },
         },
         {
@@ -8074,6 +8075,7 @@ const completionSpec: Fig.Spec = {
         },
         {
           name: "--interpolated-path",
+          requiresEquals: true,
           description:
             "To support virtual hosting, an interpolated path template can be used to dynamically construct alternate paths. The template supports %H for the target hostname as supplied by the client but converted to all lowercase, %CH for the canonical hostname, %IP for the server’s IP address, %P for the port number, and %D for the absolute path of the named repository. After interpolation, the path is validated against the directory whitelist",
           args: { name: "path-template" },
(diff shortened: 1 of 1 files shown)


src/git.ts
```diff
@@ -8028,6 +8046,186 @@ const completionSpec: Fig.Spec = {
...
+        {
+          name: "--informative-errors",
+          description:
+            'When informative errors are turned on, git-daemon will report more verbose errors to the client, differentiating conditions like "no such repository" from "repository not exported". This is more convenient for clients, but may leak information about the existence of unexported repositories. When informative errors are not enabled, all errors report "access denied" to the client',
+        },
+        {
+          name: "--no-informative-errors",
+          description:
+            "Turn off informative errors. This option is the default. See --informative-errors for more information",
+        },
```
fedeci, line 8219: These are mutually [`exclusiveOn`](https://fig.io/docs/reference/option#exclusiveon)

src/git.ts
```diff
@@ -8028,6 +8046,186 @@ const completionSpec: Fig.Spec = {
...
+        {
+          name: "--verbose",
+          description:
+            "Log details about the incoming connections and requested files",
+        },
+        {
+          name: "--detach",
+          description: "Detach from the shell. Implies --syslog",
+        },
```
fedeci, line 8163:
```suggestion
        {
          name: "--detach",
          description: "Detach from the shell",
          dependsOn: ["--syslog"],
        },
```

src/git.ts
```diff
@@ -8028,6 +8046,186 @@ const completionSpec: Fig.Spec = {
...
+        {
+          name: "--export-all",
+          description:
+            "Allow pulling from all directories that look like Git repositories (have the objects and refs subdirectories), even if they do not have the git-daemon-export-ok file",
+        },
+        {
+          name: "--inetd",
+          description: "Have the server run as an inetd service",
+        },
```
fedeci, line 8091:
```suggestion
        {
          name: "--inetd",
          description: "Have the server run as an inetd service",
          exclusiveOn: ["--pid-file"],
        },
```

fedeci (changes requested):
Thanks for this addition! We recently renamed `requiresEquals` to `requiresSeparator`, can you change those, please?
I know the docs require to be updated!

src/git.ts
j-f1: I don’t think this is a correct interpretation of Git’s behavior. Specifically, `git daemon --detach` is equivalent to `git daemon --syslog --detach`, so `--syslog` doesn’t have to be passed at all.

fedeci: Yeah sorry I read implies as requires 🤦‍♂️

fedeci (approved):
Thanks!

PR #1222, opened 2022-05-10
feat: add new fig CLIs

fedeci:
# Likely these specs will be subject to changes in the next days so we should wait before merging them


src/@fig/publish-spec-to-team.ts
```diff
@@ -0,0 +1,59 @@
...
+    },
+    {
+      name: ["-p", "--spec-path"],
+      description: "The local path of the spec to publish",
+      args: { name: "path" },
+    },
+    {
+      name: ["-b", "--binary-path"],
+      description: "The path of the binary to run to generate the spec",
+      args: { name: "path" },
```
fedeci, line 39:
```suggestion
      args: { name: "path", template: ["filepaths"] },
```

src/create-completion-spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const completionSpec: Fig.Spec = {
+  name: "create-completion-spec",
+  description: "Setup fig folder and create spec with the given name",
+  subcommands: [
+    {
+      name: "help",
+      description: "Display help for command",
+      priority: 49,
+      args: { name: "command", isOptional: true },
```
fedeci, line 11:
```suggestion
      args: { name: "command" },
```

PR #1224, opened 2022-05-11
feat: add croc spec

brandonleon:
Feature, adding spec for croc, the command line tool for sending files and folders.

No current spec for croc.

Adding spec for the croc command.

**Additional info:**

### DIFF
--- a/src/croc.ts
+++ b/src/croc.ts
@@ -1,0 +1,109 @@
+const completionSpec: Fig.Spec = {
+  name: "croc",
+  description: "",
+  subcommands: [
+    {
+      name: "send",
+      description: "Send file(s), or folder",
+    },
+    {
+      name: "relay",
+      description: "Start your own relay",
+    },
+  ],
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show help for croc",
+    },
+    {
+      name: "--internal-dns",
+      description:
+        "Use a built-in DNS stub resolver rather than the host operating system",
+    },
+    {
+      name: "--remember",
+      description: "Save these settings to reuse next time",
+    },
+    {
+      name: "--debug",
+      description: "Toggle debug mode",
+    },
+    {
+      name: "--yes",
+      description: "Automatically agree to all prompts",
+    },
+    {
+      name: "--stdout",
+      description: "Redirect file to stdout",
+    },
+    {
+      name: "--no-compress",
+      description: "Disable compression",
+    },
+    {
+      name: "--ask",
+      description: "Make sure sender and recipient are prompted",
+    },
+    {
+      name: "--local",
+      description: "Force to use only local connections",
+    },
+    {
+      name: "--ignore-stdin",
+      description: "Ignore piped stdin",
+    },
+    {
+      name: "--overwrite",
+      description: "Do not prompt to overwrite",
+    },
+    {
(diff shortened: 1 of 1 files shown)


src/croc.ts
```diff
@@ -0,0 +1,110 @@
...
+    "Croc is a tool that allows any two computers to simply and securely transfer files and folders",
+  subcommands: [
+    {
+      name: "send",
+      description: "Send file(s), or folder",
+    },
+    {
+      name: "relay",
+      description: "Start your own relay",
+    },
```
SeparateRecords, line 13: `send` and `relay` both have options but it looks like they're missing here

src/croc.ts
```diff
@@ -0,0 +1,110 @@
...
+      name: "--throttleUpload",
+      description: "Throttle the upload speed e.g. 500k",
+    },
+    {
+      name: ["--version", "-v"],
+      description: "Print the version",
+    },
+  ],
+  // Only uncomment if croc takes an argument
+  // args: {}
```
SeparateRecords, line 108: It should take an argument which is the secret code

SeparateRecords (changes requested):
Thanks for the PR! Left some notes - not much left to add 🙂

src/croc.ts
brandonleon: Added receive code argument.

src/croc.ts
```diff
@@ -0,0 +1,170 @@
...
+      args: {
+        name: "command",
+        template: "help",
+      },
+    },
+  ],
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show help for croc",
```
SeparateRecords, line 76:
```suggestion
      description: "Show help for croc",
      isPersistent: true,
```

src/croc.ts
```diff
@@ -0,0 +1,170 @@
...
+      description: 'Password for the relay (default: "pass123")',
+    },
+    {
+      name: "--socks5",
+      description: "Add a socks5 proxy",
+    },
+    {
+      name: "--throttleUpload",
+      description: "Throttle the upload speed e.g. 500k",
+    },
```
SeparateRecords, line 159: These should all take an argument, `args: { name: "value" }` on each of them would be good enough!

src/croc.ts
```diff
@@ -0,0 +1,170 @@
...
+      description: "Start your own relay",
+      options: [
+        {
+          name: "--host",
+          description: "Host of the relay",
+        },
+        {
+          name: "--ports",
+          description: "Ports of the local relay",
+        },
```
SeparateRecords, line 61: These both take a value, can just do `args: { name: "value" }`

SeparateRecords (changes requested):
Done a final pass on it, checking for missing args. I've marked the options that should take an arg, once that's done it's ready to merge!

src/croc.ts
brandonleon: Got a little more detailed on a few of them.

SeparateRecords (approved):
LGTM! Thanks for adding all of that 😄

Note that you probably won't instantly get this spec since the package size is now over 100MB, which causes issues for the CDN we use (jsdelivr).

I have an issue open to increase the size limit for us and have a workaround PR up internally, so hopefully it'll be usable soon!

PR #1227, opened 2022-05-13
feat: add tmuxinator completion spec.

kmjohny:
Add new completion spec for [tmuxinator](https://github.com/tmuxinator/tmuxinator)


src/tmuxinator.ts
```diff
@@ -0,0 +1,271 @@
...
+      ],
+    },
+    {
+      name: "debug",
+      description: "Output the shell commands that are generated by tmuxinator",
+      options: [
+        {
+          name: ["-a", "--attach"],
+          description: "Attach to tmux session after creation",
+          args: {},
```
SeparateRecords, line 85: Would you be able to give this a name? Description not necessary, name just helps in the popup

src/tmuxinator.ts
```diff
@@ -0,0 +1,271 @@
...
+      name: "doctor",
+      description: "Look for problems in your configuration",
+    },
+    {
+      name: "help",
+      description: "Describe available commands or one specific command",
+      args: {
+        name: "command",
+        description: "The tmuxinator commands",
+        generators: commands,
```
SeparateRecords, line 126: I think you might be able to just use `template: "help"` instead? It displays all the commands on the parent spec object as suggestions. Unless there's some other behavior (eg. user-defined commands) that should be fine.

SeparateRecords (commented):
Looks great! As is, we could definitely merge it, looks like everything would work fine. Just 2 small notes then I think it'll be perfect.

src/tmuxinator.ts
kmjohny: Done!

src/tmuxinator.ts
kmjohny: Makes sense. Done!.

SeparateRecords (approved):
Fantastic! Thank you so much for this. Looks brilliant.

PR #1229, opened 2022-05-15
feat(fig): update spec

mschrage:


fedeci (approved):
Finally working again!

PR #1233, opened 2022-05-21
Add completion for the LastPass CLI `lpass`

mamercad:

New completion

There is no completion for `lpass` currently

There will be completion for `lpass`

**Additional info:**

The LastPass CLI lives [here](https://github.com/lastpass/lastpass-cli).

### DIFF
--- a/src/lpass.ts
+++ b/src/lpass.ts
@@ -1,0 +1,921 @@
+import { deserialize } from "v8";
+
+const completionSpec: Fig.Spec = {
+  name: "lpass",
+  description: "",
+  subcommands: [
+    {
+      name: "login",
+      description: "Logs in to LastPass",
+      options: [
+        {
+          name: "--trust",
+          description:
+            "Subsequent logins will not require multifactor authentication",
+        },
+        {
+          name: "--plaintext-key",
+          description:
+            "The decryption key will be saved to the hard disk in plaintext",
+        },
+        {
+          name: ["--force", "-f"],
+          description: "Not documented in the man page :(",
+        },
+        {
+          name: "--color",
+          description: "Controls colored output to the terminal",
+          args: {
+            name: "color",
+            suggestions: ["auto", "never", "always"],
+          },
+        },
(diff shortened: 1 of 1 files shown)


src/lpass.ts
```diff
@@ -0,0 +1,964 @@
...
+            "Subsequent logins will not require multifactor authentication",
+        },
+        {
+          name: "--plaintext-key",
+          description: "Save decryption key to the hard disk in plaintext",
+          args: {
+            name: "--force",
+            description: "Skip user confirmation",
+            isOptional: true,
+          },
```
SeparateRecords, line 23: I think this is probably a mistake, should this be a file path instead of an arg named `--force`?

src/lpass.ts
```diff
@@ -0,0 +1,964 @@
...
+          args: {
+            name: "color",
+            suggestions: ["auto", "never", "always"],
+            isOptional: true,
+            default: "auto",
+          },
+        },
+        {
+          name: ["UNIQENAME", "UNIQUEID"],
+          description: "Shows entry with this unique name or ID",
```
SeparateRecords, line 277: Should this be an arg instead? (haven't used `lpass` so please let me know if this is indeed an option)

SeparateRecords (changes requested):
Looking good 😄 not much to change on this, only a couple places to clean up.

One thing though, you can replace `requiresEquals` with `requiresSeparator` which does the same thing. We recently deprecated `requiresEquals` since it's not generic enough, will be removed in v3.0 of the library

src/lpass.ts
mamercad: From the man page:

```
lpass login [--trust] [--plaintext-key [--force, -f]] [--color=auto|never|
always] USERNAME
...
It will prompt the user to confirm, unless --force
       is specified.
```

src/lpass.ts
mamercad: Yep, it should, nice catch!

PR #1236, opened 2022-05-22
feat(who): add completion

mamercad:

New completion

There is no completion for `who` currently

There will be completion for `who`

**Additional info:**

`man who`

### DIFF
--- a/src/who.ts
+++ b/src/who.ts
@@ -90,6 +90,7 @@ const completionSpec: Fig.Spec = {
     description:
       "By default, who gathers information from the file /var/run/utmpx; an alternative file may be specified",
     isOptional: true,
+    template: "filepaths",
   },
 };
 export default completionSpec;
--- a/src/who.ts
+++ b/src/who.ts
@@ -9,63 +9,49 @@ const completionSpec: Fig.Spec = {
       name: "-a",
       description: "Same as -bdlprTtu",
       insertValue: "-bdlprTtu",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
     {
       name: "-b",
       description: "Time of last system boot",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
     {
       name: "-d",
       description: "Print dead processes",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
     {
       name: "-H",
       description: "Write column headings above the regular output",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
     {
       name: "-l",
       description: "Print system login processes (unsupported)",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
     {
       name: "-m",
       description: "Only print information about the current terminal",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
     {
       name: "-p",
       description: "Print active processes spawned by launchd(8) (unsupported)",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
@@ -73,7 +59,6 @@ const completionSpec: Fig.Spec = {
       name: "-q",
       description:
         "'Quick mode': List only names and number of users currently logged on",
-      isDangerous: false,
       exclusiveOn: [
         "-a",
         "-b",
@@ -96,9 +81,7 @@ const completionSpec: Fig.Spec = {
     {
       name: "-r",
       description: "Print the current runlevel",
-      isDangerous: false,
       args: {
-        isDangerous: false,
         isOptional: true,
       },
     },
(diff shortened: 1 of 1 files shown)


src/who.ts
```diff
@@ -0,0 +1,153 @@
...
+  name: "who",
+  description: "Display who is logged in",
+  parserDirectives: {
+    optionsMustPrecedeArguments: true,
+  },
+  options: [
+    {
+      name: "-a",
+      description: "Same as -bdlprTtu",
+      insertValue: "-bdlprTtu",
```
SeparateRecords, line 11: I think inserting the options explicitly isn't necessary since `-a` is an option itself, although if you prefer this UX then I'm happy to go with your judgement on it

src/who.ts
```diff
@@ -0,0 +1,153 @@
...
+      isDangerous: false,
+      args: {
+        isDangerous: false,
+        isOptional: true,
+      },
```
SeparateRecords, line 16: As far as I know the options shouldn't take any args, these can be safely deleted from the options

src/who.ts
```diff
@@ -0,0 +1,153 @@
+const completionSpec: Fig.Spec = {
```
SeparateRecords, line 1: This is a bit of a weird suggestion but since it also supports `who am I` as an invocation, you can add this:
```typescript
subcommands: [
  {
    name: "am",
    insertValue: "am I",
    displayName: "am I",  // You can change this to make it look better if `am I` looks strange
    subcommands: [{ name: "I" }],
  },
],
```
Of course with appropriate descriptions, I omitted them for brevity 😄

Modelling the `who am I` invocation as an `am` subcommand with an `I` subcommand, and selecting the suggestion for `am` inserts that too. This also gives some other behavior for free (eg. it's optional and can't be used with options)

And if it looks weird, you can set `priority: 40` (or any number below 50) to put it at the bottom of the list

src/who.ts
```diff
@@ -0,0 +1,153 @@
...
+      },
+    },
+  ],
+  args: {
+    name: "file",
+    description:
+      "By default, who gathers information from the file /var/run/utmpx; an alternative file may be specified",
+    isDangerous: false,
+    isOptional: true,
+  },
```
SeparateRecords, line 151: Can use `template: "filepaths"` here

SeparateRecords (changes requested):
Looks good - same suggestion as the other review, plus a few more notes 🙂

src/who.ts
mamercad: Hehe, yeah, reading `Same as -bdlprTtu` is a bit ridiculous; I'll leave it in the description (since it's straight from the man page), but, I'll drop the `insertValue`.

src/who.ts
mamercad: Yeah, not going to lie, I was wondering how to do this -- thanks for the help!

mamercad:
@SeparateRecords I think this is ready for another look when you have time please!

src/who.ts
```diff
@@ -0,0 +1,96 @@
+const completionSpec: Fig.Spec = {
+  name: "who",
+  description: "Display who is logged in",
+  parserDirectives: {
+    optionsMustPrecedeArguments: true,
+  },
+  subcommands: [
+    {
+      name: "am",
+      insertValue: "am I",
```
SeparateRecords, line 10:
```suggestion
      insertValue: "am I{cursor}",
```
Small change, but it puts the cursor directly after `I`, eg, `who am I|` instead of `who am I |`

src/who.ts
```diff
@@ -0,0 +1,96 @@
...
+      displayName: "am I",
+      subcommands: [{ name: "I" }],
+      priority: 40,
```
SeparateRecords, line 13: Just adds descriptions
```suggestion
      description: "Returns the invoker's real user name",
      subcommands: [{
      	name: "I",
        description: "Returns the invoker's real user name",
      }],
      priority: 40,
```

SeparateRecords (approved):
Two very small tweaks to the `am I` subcommand. Everything else looks great!

(I'd recommend making the changes yourself instead of accepting them on GH to keep the sole credit for the PR - just a vanity thing 😄)

src/who.ts
mamercad: Thanks for the help through this @SeparateRecords !

src/who.ts
fedeci: I wonder if it would work with `additionalSuggestions`, instead. Let me check!

fedeci: @mamercad this works fine and you can remove the insertValue and displayValue from the `am` subcommand. 
```js
additionalSuggestions: [
    {
      name: "am I",
      insertValue: "am I{cursor}",
      icon: "fig://icon?type=command"
    }
  ],
```

mamercad: I think I had to drop the `am ` from `insertValue` though (I was getting `who am am I` (it's pushed).

fedeci: 👍

fedeci (approved):
Thanks!

PR #1237, opened 2022-05-22
feat(ansible-doc): add completion

mamercad:

New completion

There is no completion for `ansible-doc` currently

There will be completion for `ansible-doc`

**Additional info:**

`man ansible-doc`

### DIFF
--- a/src/ansible-doc.ts
+++ b/src/ansible-doc.ts
@@ -1,0 +1,158 @@
+const allPluginsGenerator: Fig.Generator = {
+  script: "ansible-doc --list --json 2>/dev/null",
+  postProcess: function (output) {
+    const plugins = JSON.parse(output);
+    return Object.keys(plugins).map((key) => ({
+      name: key,
+      description: plugins[key],
+    }));
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "ansible-doc",
+  description: "Displays information on modules installed in Ansible libraries",
+  options: [
+    {
+      name: "--metadata-dump",
+      description:
+        "For internal testing only Dump json metadata for all plugins",
+      args: {
+        isOptional: true,
+      },
+    },
+    {
+      name: "--playbook-dir",
+      description:
+        "Sets the relative path for many features including roles/ group_vars/ etc",
+      args: {
+        name: "BASEDIR",
+        description: "Base directory",
+        template: ["filepaths"],
+      },
+    },
+    {
+      name: "--version",
+      description:
+        "Shows version number, config file location, module search path, module location, executable location and exit",
+      args: {
+        isOptional: true,
+      },
+    },
+    {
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):

src/ansible-doc.ts
```diff
@@ -0,0 +1,173 @@
+const allPluginsGenerator = (
+  suggestOptions?: Partial<Fig.Suggestion>
+): Fig.Generator => ({
```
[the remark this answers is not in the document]
mamercad: It enumerates the installed plugins like this:

image

src/ansible-doc.ts
```diff
@@ -0,0 +1,173 @@
...
+  name: "ansible-doc",
+  description: "Displays information on modules installed in Ansible libraries",
+  options: [
+    {
+      name: "--metadata-dump",
+      description:
+        "For internal testing only Dump json metadata for all plugins",
+      args: {
+        isOptional: true,
+        isDangerous: false,
```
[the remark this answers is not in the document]
mamercad: Cool; I'll push without them shortly.

src/ansible-doc.ts
```diff
@@ -0,0 +1,160 @@
...
+): Fig.Generator => ({
+  script: "ansible-doc --list --json 2>/dev/null",
+  postProcess: function (output) {
+    const plugins = JSON.parse(output);
+    return Object.keys(plugins).map((key) => ({
+      name: key,
+      description: plugins[key],
+    }));
+  },
+});
```
fedeci, line 12: I mean, can we just refactor to this?
```suggestion
const allPluginsGenerator: Fig.Generator = {
  script: "ansible-doc --list --json 2>/dev/null",
  postProcess: function (output) {
    const plugins = JSON.parse(output);
    return Object.keys(plugins).map((key) => ({
      name: key,
      description: plugins[key],
    }));
  },
};
```

mamercad: I've committed this, but, it's failing `typecheck`. I don't know enough TypeScript to know what to do, can you help?

src/ansible-doc.ts
```diff
@@ -0,0 +1,158 @@
...
+      args: {
+        isOptional: true,
+      },
+    },
+  ],
+  args: {
+    name: "plugin",
+    isDangerous: true,
+    isVariadic: true,
+    generators: allPluginsGenerator(),
```
fedeci, line 154: This was the cause of the typecheck failing @mamercad 
```suggestion
    generators: allPluginsGenerator,
```

mamercad: Ah; thanks, will push.

mamercad:
Hrm; I rebased and the linter is exploding on all sorts of things that aren't mind 😄

fedeci (approved):
Thanks!

PR #1238, opened 2022-05-22
feat(ansible-config): add completion

mamercad:

New completion

There is no completion for `ansible-config` currently

There will be completion for `ansible-config`

**Additional info:**

`man ansible-config`

### DIFF
--- a/src/ansible-config.ts
+++ b/src/ansible-config.ts
@@ -7,6 +7,32 @@ const completionSpec: Fig.Spec = {
       description: "List and output available configs",
       options: [
         {
+          name: ["--help", "-h"],
+          description: "Show help and exit",
+          args: {
+            isOptional: true,
+          },
+        },
+        {
+          name: "--verbose",
+          description:
+            "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+          args: {
+            isOptional: true,
+          },
+          exclusiveOn: ["-v"],
+        },
+        {
+          name: "-v",
+          description:
+            "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+          args: {
+            isOptional: true,
+          },
+          exclusiveOn: ["--verbose"],
+          isRepeatable: 4,
+        },
+        {
           name: ["--config", "-c"],
           description:
             "Path to configuration file, defaults to first file found in precedence",
@@ -41,12 +67,45 @@ const completionSpec: Fig.Spec = {
           },
         },
       ],
+      args: {
+        name: "args",
+        description:
+          "Specific plugin to target, requires type of plugin to be set",
+        isOptional: true,
+        isVariadic: true,
+      },
     },
     {
       name: "dump",
       description:
         "Shows the current settings, merges ansible.cfg if specified",
       options: [
+        {
+          name: ["--help", "-h"],
+          description: "Show help and exit",
+          args: {
+            isOptional: true,
+          },
+        },
+        {
+          name: "--verbose",
+          description:
+            "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+          args: {
+            isOptional: true,
+          },
+          exclusiveOn: ["-v"],
+        },
+        {
+          name: "-v",
+          description:
+            "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+          args: {
+            isOptional: true,
+          },
+          exclusiveOn: ["--verbose"],
+          isRepeatable: 4,
+        },
         {
           name: ["--only-changed", "--changed-only"],
           description:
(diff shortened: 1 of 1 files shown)


src/ansible-config.ts
```diff
@@ -0,0 +1,246 @@
...
+    },
+    {
+      name: "--verbose",
+      description:
+        "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+      exclusiveOn: ["-v"],
+      args: {
+        isOptional: true,
+        isDangerous: false,
+      },
```
SeparateRecords, line 231: Does this need to be here? I think --verbose/-v don't take args (could be wrong!)

src/ansible-config.ts
```diff
@@ -0,0 +1,246 @@
...
+    },
+    {
+      name: ["--help", "-h"],
+      description: "Show help and exit",
+      args: {
+        isOptional: true,
+        isDangerous: false,
+      },
```
SeparateRecords, line 221: Same thing, does --help take an arg? If it does, what value is it expecting? (Eg subcommand name?)

src/ansible-config.ts
```diff
@@ -0,0 +1,246 @@
...
+  ],
+  options: [
+    {
+      name: "--version",
+      description:
+        "Shows version number, config file location, module search path, module location, executable location and exit",
+      args: {
+        isOptional: true,
+        isDangerous: false,
+      },
```
SeparateRecords, line 213: Should this be here?

SeparateRecords (changes requested):
Same thing as the others 🙂 `args` in some extra places, `isDangerous: false` too, but aside from that it looks great!

src/ansible-config.ts
mamercad: Right; they don't ... I've been unclear about the need for this (if the option doesn't take any arguments):

```
      args: {
        isOptional: true,
      },
```

Is it best to omit this if that's the case, or, include it to be explicit?

mamercad: (I'll push without it.)

src/ansible-config.ts
mamercad: I *think* this is fixed now.

src/ansible-config.ts
mamercad: I *think* this is fixed now.

src/ansible-config.ts
SeparateRecords: > Right; they don't ... I've been unclear about the need for this (if the option doesn't take any arguments):

If the option doesn't take any arguments then you can omit `args` entirely since the parser will then that the token following `--verbose` is the argument _for_ `--verbose`. Meaning if the next token is another option, you won't get suggestions for it since you're in an argument

SeparateRecords: > Is it best to omit this if that's the case, or, include it to be explicit?

You're correct! If no args, you can omit args entirely. Would you be able to update the other options that don't take args too?

mamercad:
@SeparateRecords I think this is ready for another look when you have time please!

src/ansible-config.ts
```diff
@@ -0,0 +1,296 @@
...
+            "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+          exclusiveOn: ["-v"],
+        },
+        {
+          name: "-v",
+          description:
+            "Verbose mode (-vvv for more, -vvvv to enable connection debugging)",
+          exclusiveOn: ["--verbose"],
+          isRepeatable: 4,
+        },
```
SeparateRecords, line 25: Are the long/short forms of this option treated differently? (eg. `-v` can be repeated but `--verbose` can't be)

Not a merge blocker but I'd like to know :)

SeparateRecords (approved):
Looks great! Thanks for making all those changes. This can definitely be merged as-is. Left one question then we're good to go 🚀

src/ansible-config.ts
fedeci: Friendly ping @mamercad. I'm merging this one in the meanwhile, thank you for the contribution!

mamercad: > Are the long/short forms of this option treated differently? (eg. `-v` can be repeated but `--verbose` can't be)
> 
> Not a merge blocker but I'd like to know :)

Yep, I think that's how it works. I think `ansible-config --verbose` is the same as `ansible-config -v` is the same as `ansible-config --verbose -v` (sigh).

But, I think we do have a problem (that doesn't break anything) with this -- it looks like `ansible-config` doesn't actually support stacking the `-v`s. I probably copy/pasta'd too much, i.e., `ansible-playbook` supports stacking the `-v's to get more and more verbosity.

Looking more closely, it looks like only `-v` and `--verbose` (which are the same) are supported to `ansible-config`.

I'll PR a fix to this shortly.

mamercad: Err, I stand corrected:

```
❯ ansible-config -h 2>&1 | grep verbose
  -v, --verbose         verbose mode (-vvv for more, -vvvv to enable
```

```
❯ ansible-config list -h 2>&1 | grep verbose
  -v, --verbose         verbose mode (-vvv for more, -vvvv to enable
```

```
❯ ansible-config dump -h 2>&1 | grep verbose
  -v, --verbose         verbose mode (-vvv for more, -vvvv to enable
```

```
❯ ansible-config view -h 2>&1 | grep verbose
  -v, --verbose         verbose mode (-vvv for more, -vvvv to enable
```

```
❯ ansible-config init -h 2>&1 | grep verbose
  -v, --verbose         verbose mode (-vvv for more, -vvvv to enable
```

mamercad:
```bash
❯ cat v.sh
#!/usr/bin/env bash

ansible-config list       >/tmp/0 2>&1
ansible-config list -v    >/tmp/1 2>&1
ansible-config list -vv   >/tmp/2 2>&1
ansible-config list -vvv  >/tmp/3 2>&1
ansible-config list -vvvv >/tmp/4 2>&1

echo "=== no v versus -v ==="
diff -q /tmp/0 /tmp/1 >/dev/null && echo same || echo diff
echo

echo "=== -v versus -vv ==="
diff -q /tmp/1 /tmp/2 >/dev/null && echo same || echo diff
echo

echo "=== -vv versus -vvv ==="
diff -q /tmp/2 /tmp/3 >/dev/null && echo same || echo diff
echo

echo "=== -vvv versus -vvvv ==="
diff -q /tmp/3 /tmp/4 >/dev/null && echo same || echo diff
echo
```

```
❯ ./v.sh

=== no v versus -v ===
diff

=== -v versus -vv ===
diff

=== -vv versus -vvv ===
same

=== -vvv versus -vvvv ===
same
```

PR #1239, opened 2022-05-22
feat(nc): add completion

mamercad:

New completion

There is no completion for `nc` currently

There will be completion for `nc`

**Additional info:**

`man nc`

### DIFF
--- a/src/nc.ts
+++ b/src/nc.ts
@@ -139,6 +139,7 @@ const completionSpec: Fig.Spec = {
         name: "port",
         isDangerous: false,
       },
+      exclusiveOn: ["-p", "-s", "-z", "-w"],
     },
     {
       name: "-L",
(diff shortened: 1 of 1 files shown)


src/nc.ts
```diff
@@ -0,0 +1,324 @@
...
+  parserDirectives: {
+    optionsMustPrecedeArguments: true,
+  },
+  options: [
+    {
+      name: "-4",
+      description: "Forces nc to use IPv4 addresses only",
+      args: {
+        isOptional: true,
+        isDangerous: false,
```
SeparateRecords, line 15: This line isn't necessary because `isDangerous: false` is the default

SeparateRecords (changes requested):
Looks good to me! One minor change then it's ready to merge: `isDangerous: false` is in a few places, which isn't necessary since that's the default behavior. I've annotated the first one but it should be easy enough to find/delete the rest!

src/nc.ts
mamercad: I'll fix this up in the rest of my open PRs as well.

SeparateRecords (approved):

SeparateRecords:
Fantastic, thanks so much for this! Will review your other PRs as well

PR #1241, opened 2022-05-23
feat(deno): update to 1.22

SeparateRecords:
This uses the completions generated by `deno completions fig`, which should make it easier to switch to automatic PRs in the future if we go down that route with it. Even if not, this _massively_ simplifies the amount of work it'll be to update the deno spec.

Because it's generated by the CLI itself it's missing some quality of life stuff that I've manually added to it over the last 8 months, when it's added back in I'll mark the PR as ready.

### DIFF
--- a/src/deno.ts
+++ b/src/deno.ts
@@ -2508,6 +2508,9 @@ const completion: Fig.Spec = {
     {
       name: "task",
       description: "Run a task defined in the configuration file",
+      parserDirectives: {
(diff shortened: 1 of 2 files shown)


fedeci:
I need to fix linting for the `push-to-autocomplete-action` then we should be good to propose deno to integrate directly with the autocomplete repo.

fedeci (approved):

PR #1243, opened 2022-05-24
Update Remotion completions for 3.0

JonnyBurger:

Update Remotion completions with new commands, flags and Remotion Lambda feature

Represents the capability of Remotion 2.0 while 3.0 is out now

Represents 3.0 functionality

**Additional info:**

### DIFF
--- a/src/remotion.ts
+++ b/src/remotion.ts
@@ -124,6 +124,7 @@ const renderOptions: Fig.Option[] = [
         {
           type: "arg",
           displayName: "[json string]",
+          insertValue: "'{cursor}'",
         },
       ],
     },
@@ -669,6 +670,7 @@ const completionSpec: Fig.Spec = {
               {
                 type: "arg",
                 displayName: "[json string]",
+                insertValue: "'{cursor}'",
               },
             ],
           },
--- a/src/remotion.ts
+++ b/src/remotion.ts
@@ -489,7 +489,10 @@ const completionSpec: Fig.Spec = {
                     "Type of CPU architecture to use for the Lambda function",
                   args: {
                     name: "architecture",
-                    suggestions: [{ name: "x86_64" }, { name: "arm64" }],
+                    suggestions: [
+                      { name: "x86_64", icon: "fig://icon?type=cpu" },
+                      { name: "arm64", icon: "fig://icon?type=cpu" },
+                    ],
                   },
                 },
                 {
--- a/src/remotion.ts
+++ b/src/remotion.ts
@@ -1,9 +1,227 @@
+const globalLambdaOptions: Fig.Option[] = [
+  {
+    name: "--quiet",
+    description: "Print less output",
+  },
+  {
+    name: "-q",
+    hidden: true,
+    description: "Print less output",
+  },
+  {
+    name: "--yes",
+    description: "Skip confirmation",
+  },
+  {
+    name: "--force",
+    hidden: true,
+    isDangerous: true,
+    description: "Skip confirmation",
+  },
+  {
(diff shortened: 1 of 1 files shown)


src/remotion.ts
```diff
@@ -1,184 +1,615 @@
...
+  },
+  {
+    name: "--frames-per-lambda",
+    insertValue: "--frames-per-lambda=",
+    description: "How many frames should be rendered per chunk",
+    args: {
+      name: "framesPerLambda",
+    },
+  },
+];
```
SeparateRecords, line 79: Do these options require an `=` separator? Federico fixed the implementation in the autocomplete engine (and we renamed a property) so now you can omit the `insertValue` and set `requiresSeparator: true`, you'll get this behavior for free.

(same goes for the other options further down in this file)

src/remotion.ts
```diff
@@ -1,184 +1,615 @@
...
+    name: "--props",
+    insertValue: "--props=",
+    description: "Pass input props as filename or as JSON",
+    args: {
+      template: ["filepaths"],
+      suggestions: [
+        {
+          type: "arg",
+          displayName: "[json string]",
         },
```
SeparateRecords, line 133: Can use `insertValue` to automatically put the cursor between two quotes, which is likely what you want if you're writing JSON in a shell
```suggestion
        {
          type: "arg",
          displayName: "[json string]",
          insertValue: "'{cursor}'",
        },
```

src/remotion.ts
```diff
@@ -1,184 +1,615 @@
...
+  },
+  {
+    name: "--quality",
+    insertValue: "--quality=",
+    description: "Quality for rendered frames, JPEG only, 0-100",
+    args: {
+      default: "80",
+      suggestions: [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100].map((e) => ({
+        name: e.toString(),
+      })),
```
SeparateRecords, line 145: Clever! 😄

src/remotion.ts
```diff
@@ -1,184 +1,615 @@
...
+                  },
+                },
+                {
+                  name: "--architecture",
+                  description:
+                    "Type of CPU architecture to use for the Lambda function",
+                  insertValue: "--architecture=",
+                  args: {
+                    name: "architecture",
+                    suggestions: [{ name: "x86_64" }, { name: "arm64" }],
```
SeparateRecords, line 511: This isn't documented anywhere but there's actually an icon for CPU architecture! `fig://icon?type=cpu`

SeparateRecords (changes requested):
Looking really good! Not much to change, awesome job on this update to the CLI. Looks like it was all hand written 😮

src/remotion.ts
JonnyBurger: This is nice to hear! I must have been on the old version, yes, I wanted that when you autocomplete this, the options pop up right away.

src/remotion.ts
JonnyBurger: Thanks! 😄 Credits to @MatteoGauthier!

src/remotion.ts
JonnyBurger: Nice, we'll take that!

JonnyBurger:
Thanks a lot for these great improvement suggestions! 🙌 
All fixed up!

SeparateRecords (approved):

SeparateRecords:
Perfect - awesome job on this (and on 3.0!) 🎉🚀

MatteoGauthier:
Extending fig is so easy, thanks for that

PR #1247, opened 2022-05-27
feat(screen): add completion

mamercad:

New completion

There is no completion for `screen`

There will be completion for `screen`

**Additional info:**

`man screen`

### DIFF
--- a/src/screen.ts
+++ b/src/screen.ts
@@ -1,0 +1,247 @@
+const identifierArg: Fig.Arg = {
+  name: "identifier",
+  suggestions: [
+    "pid",
+    "pid.tty",
+    "pid.tty.host",
+    "sessionowner/pid",
+    "sessionowner/pid.tty",
+    "sessionowner/pid.tty.host",
+  ],
+};
+
+const completionSpec: Fig.Spec = {
+  name: "screen",
+  description: "Screen manager with VT100/ANSI terminal emulation",
+  parserDirectives: {
+    optionsMustPrecedeArguments: true,
+    flagsArePosixNoncompliant: true,
+  },
+  subcommands: [
+    {
+      name: "-d",
+      description:
+        "Does not start screen, but detaches the elsewhere running screen session",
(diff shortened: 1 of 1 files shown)


src/screen.ts
```diff
@@ -0,0 +1,240 @@
...
+      description: "Tell screen to use UTF-8 encoding",
+    },
+    {
+      name: "-v",
+      description: "Print 'Screen version 4.00.03 (FAU) 23-Oct-06'",
+    },
+    {
+      name: "-wipe",
+      description: "Do nothing, just clean up SockDir",
+    },
```
fedeci, line 217: Does this have double dash? If it doesn't you should add the `flagsArePosixNoncompliant` parserOption: https://fig.io/docs/reference/subcommand#parserdirectives

fedeci: I've just seen that many options are not simple flags (e.g. `-RR` or `-fn`) so that `parserOption` is required for correct parsing.

src/screen.ts
```diff
@@ -0,0 +1,240 @@
...
+    {
+      name: "-dmS",
+      description: "Start as daemon: Screen session in detached mode",
+      icon: "fig://icon?type=option",
+      args: {
+        name: "name",
+        description: "Name of the screen session",
+      },
+    },
+  ],
```
fedeci, line 98: Why these are considered as subcommands?

fedeci (changes requested):

src/screen.ts
mamercad: What's the mechanism for "chaining arguments"?

fedeci: Do you mean that `-d` is exclusive on `-D` and they both allow some kind of subcommand (`-r`, `R`, `-RR`) after them?

fedeci:
It would be great if you could rebase and lint:fix everything after reinstalling npm deps.

fedeci (approved):
Thanks!

PR #1249, opened 2022-05-27
feat(tr): add completion

mamercad:

New completion

There is no completion for `tr`

There will be completion for `tr`

**Additional info:**

`man tr`

### DIFF
--- a/src/tr.ts
+++ b/src/tr.ts
@@ -1,0 +1,195 @@
+const stringSuggestions: Fig.Suggestion[] = [
+  {
+    name: "a",
+    description: "Any single character",
+    priority: 40,
+  },
+  {
+    name: "\\a",
+    description: "Alert character",
+    priority: 39,
+  },
+  {
+    name: "\\b",
+    description: "Backspace character",
+    priority: 39,
+  },
+  {
+    name: "\\f",
+    description: "Form feed character",
+    priority: 39,
+  },
+  {
+    name: "\\n",
+    description: "Newline character",
+    priority: 39,
+  },
+  {
+    name: "\\r",
+    description: "Carriage return character",
+    priority: 39,
+  },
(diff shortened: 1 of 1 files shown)


src/tr.ts
```diff
@@ -0,0 +1,195 @@
...
+  options: [
+    {
+      name: "-C",
+      description:
+        "Complement the set of characters in string1, that is '-C ab' includes every character except for 'a' and 'b'",
+    },
+    {
+      name: "-c",
+      description: "Same as '-C' but complement the set of values in string1",
+    },
```
fedeci, line 157: I believe these accept an argument

src/tr.ts
```diff
@@ -0,0 +1,195 @@
...
+      description:
+        "Squeeze multiple occurrences of the characters listed in the last operand (either string1 or string2) in the input into a single instance of the character. This occurs after all deletion and translation is completed",
+      exclusiveOn: ["-d", "-sd"],
+    },
+    {
+      name: "-ds",
+      description:
+        "Delete characters in string1 and squeeze multiple occurrences of the characters listed in string2 into a single instance of the character",
+      exclusiveOn: ["-s", "-d"],
+    },
```
fedeci, line 174: This does not need to be added as a separate option since it is the combination of -s and -d.

fedeci (changes requested):

src/tr.ts
mamercad: Isn't that handled by `string1`?

```bash
SYNOPSIS
     tr [-Ccsu] string1 string2
     tr [-Ccu] -d string1
     tr [-Ccu] -s string1
     tr [-Ccu] -ds string1 string2
```

fedeci: It is!

PR #1250, opened 2022-05-27
feat(vault): add completion

mamercad:

New completion (opening this as a Draft to prevent duplication of effort)

There is no completion for `vault`

There will be completion for `vault`

**Additional info:**

`vault -h`

### DIFF
--- a/src/vault.ts
+++ b/src/vault.ts
@@ -1,0 +1,2874 @@
+const vaultCommonOptions: Fig.Option[] = [
+  {
+    name: ["--help", "-h"],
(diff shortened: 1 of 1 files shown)


fedeci (approved):
Thanks!

fedeci (changes requested):
I've seen `requiresEquals` has been used but that has been deprecated in favour of `requiresSeparator`. Setting that to `true` behaves the same as `requiresEquals: true`. Can you update this small nit please?

fedeci (approved):

PR #1254, opened 2022-05-28
feat(tac): add completion

mamercad:

New completion

There is no completion for `tac`

There will be completion for `tac`

**Additional info:**

```bash
 tac --help
Usage: tac [OPTION]... [FILE]...
Write each FILE to standard output, last line first.

With no FILE, or when FILE is -, read standard input.

Mandatory arguments to long options are mandatory for short options too.
  -b, --before             attach the separator before instead of after
  -r, --regex              interpret the separator as a regular expression
  -s, --separator=STRING   use STRING as the separator instead of newline
      --help        display this help and exit
      --version     output version information and exit

GNU coreutils online help: <https://www.gnu.org/software/coreutils/>
Full documentation <https://www.gnu.org/software/coreutils/tac>
or available locally via: info '(coreutils) tac invocation'
```

### DIFF
--- a/src/tac.ts
+++ b/src/tac.ts
@@ -1,0 +1,34 @@
+const completionSpec: Fig.Spec = {
+  name: "tac",
+  description: "Concatenate and print files in reverse",
+  options: [
+    {
+      name: "--help",
+      description: "Display this help and exit",
+    },
+    {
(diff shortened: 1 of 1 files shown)


fedeci (approved):
👍

PR #1256, opened 2022-05-28
feat(od): add completion

mamercad:

New completion

There is no completion for `od`

There will be completion for `od`

**Additional info:**

`man od`


SeparateRecords (approved):
Awesome job! Nothing needs to be changed 😄

PR #1260, opened 2022-05-29
feat(pr): add completion

mamercad:

New completion

There is no completion for `pr`

There will be completion for `pr`

**Additional info:**

`man pr`

### DIFF
--- a/src/pr.ts
+++ b/src/pr.ts
@@ -18,35 +18,23 @@ const completionSpec: Fig.Spec = {
       // NOTE: I'm not sure how to do this; could use help
       name: "-",
       hidden: true,
-      description: `Produce output that is columns wide (default is 1) that is
-      written vertically down each column in the order in which the
-      text is received from the input file.  The options -e and -i are
-      assumed.  This option should not be used with -m.  When used with
-      -t, the minimum number of lines is used to display the output.
-      (To columnify and reshape text files more generally and without
-      additional formatting, see the rs(1) utility.)`,
+      description:
+        "Produce output that is columns wide (default is 1) that is written vertically down each column in the order in which the text is received from the input file. The options -e and -i are assumed. This option should not be used with -m. When used with -t, the minimum number of lines is used to display the output. (To columnify and reshape text files more generally and without additional formatting, see the rs(1) utility.)",
     },
     {
       name: "-a",
-      description: `Modify the effect of the -column option so that the columns are
-      filled across the page in a round-robin order (e.g., when column
-      is 2, the first input line heads column 1, the second heads
-      column 2, the third is the second line in column 1, etc.).  This
-      option requires the use of the -column option.`,
+      description:
+        "Modify the effect of the -column option so that the columns are filled across the page in a round-robin order (e.g., when column is 2, the first input line heads column 1, the second heads column 2, the third is the second line in column 1, etc.). This option requires the use of the -column option",
     },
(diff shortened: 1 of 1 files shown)


src/pr.ts
```diff
@@ -0,0 +1,194 @@
...
+      name: "-",
+      hidden: true,
+      description: `Produce output that is columns wide (default is 1) that is
+        written vertically down each column in the order in which the
+        text is received from the input file.  The options -e and -i are
+        assumed.  This option should not be used with -m.  When used with
+        -t, the minimum number of lines is used to display the output.
+        (To columnify and reshape text files more generally and without
+        additional formatting, see the rs(1) utility.)`,
+    },
```
SeparateRecords, line 28: @fedeci I'm not sure what to do for this! Using `-` and `+` as options isn't _super_ common but a few included binaries do use this. I think this should be supported but it doesn't seem to be?

src/pr.ts
```diff
@@ -0,0 +1,194 @@
+const completionSpec: Fig.Spec = {
+  name: "pr",
+  description: "",
```
SeparateRecords, line 3:
```suggestion
  description: "A printing and pagination filter for text files",
```

SeparateRecords (commented):
Looking great!

src/pr.ts
```diff
@@ -0,0 +1,194 @@
+const completionSpec: Fig.Spec = {
+  name: "pr",
+  description: "A printing and pagination filter for text files",
+  parserDirectives: {
+    optionsMustPrecedeArguments: true,
+  },
+  options: [
+    {
+      // NOTE: I'm not sure how to do this; could use help
```
fedeci, line 9: I'm unsure how to make this work. Because if the option token does not start with a dash we look up for it in the subcommands and not in the options. Maybe you can add both the `+` and `-` as the suggestions of an optional argument before the file(s) one.
```suggestion
```

src/pr.ts
```diff
@@ -0,0 +1,194 @@
...
+      name: "+",
+      hidden: true,
+      description: `Begin output at page number page of the formatted input.`,
+      args: {
+        name: "page",
+      },
+    },
+    {
+      // NOTE: I'm not sure how to do this; could use help
```
fedeci, line 18: This works correctly.
```suggestion
```

src/pr.ts
```diff
@@ -0,0 +1,194 @@
...
+      },
+    },
+  ],
+  args: {
+    name: "file",
+    description: "File(s) to be printed",
+    isOptional: true,
+    isVariadic: true,
+    template: "filepaths",
+  },
```
fedeci, line 192:
```js
  args: [{
    suggestions: [{ name: "+", ... } , { name: "-", ... }],
		isOptional: true,
  }, {
    name: "file",
    ...
  }],
```

fedeci: Tested it and it does not work because for `pr ` it only suggests `+` and `-` along with the options but no filepaths.

fedeci:
I am also currently under exam session so I am not really into making potentially breaking changes to the engine, @SeparateRecords we can chat with Sean about this but it's likely a bug that if I have two following arguments where the first of them is optional I am not receiving suggestions for both the args.

fedeci:
Thanks!

mamercad:
...I'm not sure what to do, closing.

PR #1266, opened 2022-05-31
feat(railway): update options and subcommands

saliven:
Update options for some commands, and add `down` and `shell` completion.

No completion for `down` and `shell` commands, and `-s` or `--service` options.

Completion for `down` and `shell` commands, and `-s` or `--service` options.

**Additional info:**


fedeci (approved):
Thanks!

PR #1270, opened 2022-06-04
fix(dig): correct subcommands

### LINKED ISSUE #1240: dig autocomplete inserts escape characters

jordi-adame:
Buf fix

#1240 

**Additional info:**

https://user-images.githubusercontent.com/1081448/172023284-598a85a7-4190-4a4f-bd37-9b18a9013e6f.mov

### DIFF
--- a/src/dig.ts
+++ b/src/dig.ts
@@ -92,6 +92,8 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: "+nocomments +noquestion +noauthority +noadditional +nostats",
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
Thanks for the PR, but I'm wondering how this spec could work, I doubt those subcommands gets correctly suggested, would you be open to update this pr with a more wide fix for the whole spec according to the [docs](https://fig.io/docs/getting-started/first-completion-spec)?

PR #1274, opened 2022-06-08
feat: add `op` completions

SeparateRecords:
Adds 1Password CLI completions

Fig autocomplete open with 1Password suggestions

This spec is [generated by a ~~very scuffed~~ script](https://gist.github.com/SeparateRecords/1e73dd825bbc0724d930da2d3bbdff0e). Icons, suggestions, templates, and generators are added by hand


fedeci (approved):
The script looks good and the generator at the beginning of the file looks good!

PR #1275, opened 2022-06-08
feat: add `newman` completions

SeparateRecords:

**Additional info:**


src/newman.ts
```diff
@@ -0,0 +1,223 @@
...
+      args: {
+        name: "collection",
+        template: "filepaths",
+      },
+    },
+    {
+      name: "help",
+      description: "Display help for command",
+      priority: 49,
+      args: { name: "command", isOptional: true },
```
fedeci, line 210: Can you add an help template here? I think it isn't embedded directly in the integration because of some reasons I can't remember :)

fedeci (approved):
Just one subtle thing to fix then you can merge this!

PR #1279, opened 2022-06-10
feat: create `trunk` completion spec

t-bre:
Feature

No completion spec for `trunk`

Creates completion spec for `trunk`

**Additional info:**
[trunk website](https://trunk.io/products/check)

### DIFF
--- a/src/trunk.ts
+++ b/src/trunk.ts
@@ -247,7 +247,7 @@ const completionSpec: Fig.Spec = {
       isPersistent: true,
       args: {
         name: "value",
-        suggestions: ["0", "1"],
+        suggestions: ["true", "false"],
       },
     },
     {
--- a/src/trunk.ts
+++ b/src/trunk.ts
@@ -77,7 +77,6 @@ const checkAndFormatSharedOptions: Fig.Option[] = [
 const completionSpec: Fig.Spec = {
   name: "trunk",
   description: "",
-
   subcommands: [
     {
       name: "init",
--- a/src/trunk.ts
+++ b/src/trunk.ts
@@ -76,7 +76,8 @@ const checkAndFormatSharedOptions: Fig.Option[] = [
 
 const completionSpec: Fig.Spec = {
   name: "trunk",
-  description: "",
+  description:
+    "An all-in-one tool for scalably checking, formatting, and monitoring code",
   subcommands: [
     {
       name: "init",
--- a/src/trunk.ts
+++ b/src/trunk.ts
@@ -89,10 +89,12 @@ const completionSpec: Fig.Spec = {
         {
           name: "--check-sample",
           description: "Run `trunk check sample` without prompting",
+          exclusiveOn: ["--nocheck-sample"],
         },
         {
           name: "--nocheck-sample",
           description: "Do not run `trunk check sample` post-init",
+          exclusiveOn: ["--check-sample"],
         },
         {
           name: "--force",
@@ -265,12 +267,14 @@ const completionSpec: Fig.Spec = {
       name: "--no-progress",
       description: "Don't show progress updates",
       isPersistent: true,
+      exclusiveOn: ["--ci-progress"],
     },
     {
       name: "--ci-progress",
       description:
         "Show updates every 30 seconds without clearing terminal screen (implicit with --ci)",
       isPersistent: true,
+      exclusiveOn: ["--no-progress"],
     },
     {
       name: "--action_timeout",
--- a/src/trunk.ts
+++ b/src/trunk.ts
@@ -1,0 +1,292 @@
+const checkAndFormatSharedOptions: Fig.Option[] = [
+  {
+    name: ["-a", "--all"],
+    description: "Run on all files instead of only changed files",
+  },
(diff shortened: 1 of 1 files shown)


src/trunk.ts
```diff
@@ -0,0 +1,291 @@
...
+      name: "--no-progress",
+      description: "Don't show progress updates",
+      isPersistent: true,
+    },
+    {
+      name: "--ci-progress",
+      description:
+        "Show updates every 30 seconds without clearing terminal screen (implicit with --ci)",
+      isPersistent: true,
+    },
```
fedeci, line 274: Are these mutually [exclusiveOn](https://fig.io/docs/reference/option#exclusiveon)

src/trunk.ts
```diff
@@ -0,0 +1,291 @@
...
+          description: "Add sha256s to trunk.yaml for additional verification",
+        },
+        {
+          name: "--check-sample",
+          description: "Run `trunk check sample` without prompting",
+        },
+        {
+          name: "--nocheck-sample",
+          description: "Do not run `trunk check sample` post-init",
+        },
```
fedeci, line 96: These too are likely `exclusiveOn`

fedeci (changes requested):
Thanks for this spec, I left two comments to be solved before merging but overall it looks good!

src/trunk.ts
t-bre: Yes you're right, fixed

src/trunk.ts
t-bre: Same here, also fixed

src/trunk.ts
```diff
@@ -0,0 +1,295 @@
...
+      "Upstream branch used to compute changed files (autodetected by default)",
+    args: {
+      name: "branch",
+    },
+  },
+];
+
+const completionSpec: Fig.Spec = {
+  name: "trunk",
+  description: "",
```
SeparateRecords, line 79: Would you be able to add a description here?

src/trunk.ts
```diff
@@ -0,0 +1,295 @@
...
+      isPersistent: true,
+    },
+    {
+      name: ["-m", "--monitor"],
+      description:
+        "Enable the trunk daemon to monitor file changes in your repo",
+      isPersistent: true,
+      args: {
+        name: "value",
+        suggestions: ["0", "1"],
```
SeparateRecords, line 249: Should this be `true`/`false` instead of `0`/`1`?

src/trunk.ts
```diff
@@ -0,0 +1,295 @@
...
+    },
+    {
+      name: "--ci-progress",
+      description:
+        "Show updates every 30 seconds without clearing terminal screen (implicit with --ci)",
+      isPersistent: true,
+      exclusiveOn: ["--no-progress"],
+    },
+    {
+      name: "--action_timeout",
```
SeparateRecords, line 280: All the other options use dashes, should this be a dash instead of underscore?

src/trunk.ts
```diff
@@ -0,0 +1,295 @@
...
+  options: [
+    {
+      name: ["-h", "--help"],
+      description: "Usage information",
+      isPersistent: true,
+    },
+    {
+      name: "--version",
+      description: "The version",
+      isPersistent: true,
```
SeparateRecords, line 240: Is this persistent on each subcommand? Usually `--version` is only on the root (I don't have trunk installed to check)

SeparateRecords (changes requested):
Looks good! Couple of quetions in reviews, but aside from that this spec is in a really good place :)

src/trunk.ts
t-bre: Good point, just checked and both ways work. I think `true`/`false` is more intuitive, so I've changed it to that.

src/trunk.ts
t-bre: Yes sorry I missed that, done now.

src/trunk.ts
t-bre: Yeah I thought that as well, but no for some reason this one has an underscore. Tried it with a dash and it throws an error. According to the help text `--action_timeout` is correct:

```
usage
  trunk [flags] [subcommand]

flags
  -h, --help                  usage information
      --version               the version
  -m, --monitor               enable the trunk daemon to monitor file changes in your repo
      --ci                    run in continuous integration mode
  -o, --output                output format {text | json}
      --no-progress           don't show progress updates
      --ci-progress           show updates every 30 seconds without clearing terminal screen (implicit with --ci
      --action_timeout        how long actions (downloads, lint runs, etc.) have to time out
  -v, --verbose               output details about what's happening under the hood
```

SeparateRecords: That's super weird! Thanks for double checking 😄

src/trunk.ts
SeparateRecords: Neat! Unusual behavior but good to have it modeled perfectly.

SeparateRecords (commented):
LGTM! Thanks so much for contributing this 😄

SeparateRecords (approved):

PR #1280, opened 2022-06-10
feat(split): add completion

mamercad:

New completion

There is no completion for `split`

There will be completion for `split`

**Additional info:**

`man split`


fedeci (approved):
Thanks!

PR #1282, opened 2022-06-12
feat(join): add completion

mamercad:

New completion

There is no completion for `join`

There will be completion for `join`

**Additional info:**

`man join`

### DIFF
--- a/src/join.ts
+++ b/src/join.ts
@@ -1,0 +1,92 @@
+const completionSpec: Fig.Spec = {
+  name: "join",
+  description:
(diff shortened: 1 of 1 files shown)


src/join.ts
```diff
@@ -0,0 +1,92 @@
...
+        name: "field",
+      },
+    },
+    {
+      name: "-j2",
+      description: "Join on the field'th field of file2",
+      args: {
+        name: "field",
+      },
+    },
```
fedeci, line 72: These can be added as `additionalSuggestion` or removed since are a combination of options

fedeci (approved):
👍

PR #1295, opened 2022-06-17
feat(mknod): add completion

mamercad:

New completion

There is no completion for `mknod`

There will be completion for `mknod`

**Additional info:**

`man mknod`

### DIFF
--- a/src/mknod.ts
+++ b/src/mknod.ts
@@ -1,6 +1,16 @@
 const completionSpec: Fig.Spec = {
   name: "mknod",
   description: "Create device special file",
+  subcommands: [
+    {
+      name: "c",
+      description: "Create (c)haracter device",
+    },
+    {
+      name: "b",
+      description: "Create (b)lock device",
+    },
+  ],
   options: [
     {
       name: "-F",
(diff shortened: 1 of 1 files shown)


src/mknod.ts
```diff
@@ -0,0 +1,63 @@
...
+    {
+      name: "c",
+      description: "Create (c)haracter device",
+      exclusiveOn: ["b"],
+    },
+    {
+      name: "b",
+      description: "Create (b)lock device",
+      exclusiveOn: ["c"],
+    },
```
fedeci, line 44: These are subcommands since they are mutually exclusive and do not have a dash.

src/mknod.ts
```diff
@@ -0,0 +1,63 @@
...
+    {
+      name: "major minor",
+      description:
+        "The major device number is an integer number which tells the kernel which device driver entry point to use; the minor device number tells the kernel which one of several similar devices the node corresponds to; for example, it may be a specific serial port or pty",
+    },
+    {
+      name: "major unit subunit",
+      description:
+        "The major device number is an integer number which tells the kernel which device driver entry point to use; the unit and subunit numbers select a subset of a device; for example, the unit may specify a particular SCSI disk, and the subunit a partition on that disk",
+    },
```
fedeci, line 60: These appear to be the same arg, likely the structure you want is: two required args (major and minor) and one optional arg (subunit)

fedeci (changes requested):
Thanks! Can you resolve the comments please?

src/mknod.ts
mamercad: Hrm, I'm not sure that's right, the second is "major unit subunit" (not "major minor").

mamercad: I've pushed an idea...

fedeci: Looks better!

fedeci (approved):

PR #1298, opened 2022-06-18
feat(`stat`,`readlink`): add completion

mamercad:

New completions

There are no completions for `stat` and `readlink`

There will be completions for `stat` and `readlink`

**Additional info:**

`man stat`


fedeci (approved):
Thanks!

PR #1300, opened 2022-06-18
feat(`basename`,`dirname`): add completion

mamercad:

New completions

There is no completion for `basename` and `dirname`

There will be completion for `basename` and `dirname`

**Additional info:**

`man basename`


fedeci (approved):
LGTM!

Issue #101, opened 2021-04-07
Docksal's `fin` client


mattbloomfield:
### CLI Tool
What CLI tool is missing autocompletion in Fig?

`fin` (Docksal's CLI)

### Useful Links
Please add some links to the docs of that tool or the `--help`text

https://docs.docksal.io/fin/fin-help/

### Top Workflows
What are the most used commands or workflows in that CLI tool (for you)?

fin restart
fin project reset cli
fin start
fin stop
fin <custom_command> (found in `.docksal/commands/*`

mattschrage:
@mattbloomfield is this blocked by #106 - should we try to build out the skeleton for the CLI or wait until we can dynamically generate subcommands?

SeparateRecords:
Done!

PR #1304, opened 2022-06-21
feat: add docksal/fin completion spec

XxHalbfettxX:
Adding completion spec for Docksal/fin

#101

**Additional info:**


src/fin.ts
```diff
@@ -0,0 +1,928 @@
...
+          name: "alias",
+          description:
+            "Manage aliases that allow fin @alias execution (fin help alias)",
+          subcommands: [
+            {
+              name: "list",
+              description: "Show aliases list",
+            },
+            {
+              name: "",
```
SeparateRecords, line 58: Looks like this is missing a name?

src/fin.ts
```diff
@@ -0,0 +1,928 @@
...
+              args: {
+                name: "file",
+              },
+            },
+            {
+              name: ["list", "ls"],
+              description: "Show list of existing databases",
+              args: {
+                name: "",
+                description: "",
```
SeparateRecords, line 106: Name/description can be omitted if you want them to be blank, adding a name is good though because it shows up on https://fig.io/manual :)

src/fin.ts
```diff
@@ -0,0 +1,928 @@
...
+              description: "Create a database",
+              args: {
+                name: "name",
+              },
+            },
+            {
+              name: "drop",
+              description: "Delete a database",
+              args: {
+                name: "name",
```
SeparateRecords, line 128: Might be able to use `fin db list` for a generator 🙂

src/fin.ts
```diff
@@ -0,0 +1,928 @@
...
+                name: "string",
+              },
+            },
+          ],
+          args: {
+            name: "SERVICE",
+          },
+        },
+        {
+          // TODO: ???????
```
SeparateRecords, line 718: Got a good chuckle from this line 😄

Does this subcommand still need its options? Is that what the todo is for?

SeparateRecords (changes requested):
Wow, big spec!! Thank you for the contribution. Looks like a lot of work to make this.

Most of it looks great. Only some (hopefully) easy additions :)

The biggest thing to change is moving the spec itself out of `generateSpec`, because generateSpec merges the output and spec.

SeparateRecords:
@XxHalbfettxX quick ping! Would love to get this merged, shouldn't take much time to get the spec ready :) happy to help out / provide more info if you want!

XxHalbfettxX:
I'm on holiday this week and would get on it next week :)

src/fin.ts
XxHalbfettxX: Oh, I forgot to do this ToDo. 😅
This subcommand was mentioned in the man page, but was removed in the current version of. Therefore, I will remove this subcommand.

src/fin.ts
```diff
@@ -0,0 +1,928 @@
+const completionSpec: Fig.Spec = {
+  name: "fin",
+  description: "Docksal command line utility",
+  generateSpec: async (tokens, executeShellCommand) => {
```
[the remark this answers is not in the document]
XxHalbfettxX: Thanks for the Tip :)

src/fin.ts
XxHalbfettxX: I forgot to remove this section :D

XxHalbfettxX:
Are the desired changes fulfilled by my adaptations?

SeparateRecords:
Hey hey! Sorry for the wait, thanks for the ping 🙂 Will review now.

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+            {
+              name: ["--all", "-a"],
+              description: "Stop all services on all Docksal projects",
+            },
+          ],
+          args: {
+            name: "service",
+            isOptional: true,
+          },
+        },
```
SeparateRecords, line 256: Is there a way to list the valid services? We can add in a follow up pr if you'd like

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+          description:
+            "Add a private SSH key from $HOME/.ssh by file name. mAdds all default keys (id_rsa/id_dsa/id_ecdsa/id_ed25519) if no file name is given",
+          args: {
+            name: "key-name",
+            isOptional: true,
+          },
+          options: [
+            {
+              name: "--quiet",
+            },
```
SeparateRecords, line 370: Could you add a description to this one?

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+      ],
+    },
+    {
+      name: "ssh-key",
+      description: "Manage SSH keys (fin help ssh-key)",
+      subcommands: [
+        {
+          name: "add",
+          description:
+            "Add a private SSH key from $HOME/.ssh by file name. mAdds all default keys (id_rsa/id_dsa/id_ecdsa/id_ed25519) if no file name is given",
```
SeparateRecords, line 362: Extra `m` in the text 🙂
```suggestion
            "Add a private SSH key from $HOME/.ssh by file name. Adds all default keys (id_rsa/id_dsa/id_ecdsa/id_ed25519) if no file name is given",
```

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+          args: {
+            name: "key-name",
+            isOptional: true,
+          },
```
SeparateRecords, line 366: If this uses keys in ~/.ssh we can use a generator like this:
```js
generators: {
  script: "\\command ls .ssh | grep --color=never -E '^id_'",
  splitOn: "\n",
}
```

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+            },
+          ],
+        },
+        {
+          name: "ls",
+          description: "List SSH keys loaded in the docksal-ssh-agent",
+        },
+        {
+          name: "rm",
+          description: "Remove all keys from the docksal-ssh-agent",
```
SeparateRecords, line 379: Can reuse the output from `fin ssh-key ls` in a generator?

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+          name: "remove",
+          description: "Remove Docksal machine and cleanup after it",
+        },
+        {
+          name: "ip",
+          description: "Show the machine IP address",
+        },
+        {
+          name: "ls",
+          description: "List all docker machines",
```
SeparateRecords, line 453: You could probably use this command for a generator for the other `fin vm` commands :D

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+              name: "--global",
+              description: "Set for global ENV file",
+            },
+            {
+              name: "--env",
+              description: "Set in environment specific project ENV file",
+              requiresEquals: true,
+              args: {
+                name: "name",
+              },
```
SeparateRecords, line 557:
```suggestion
              args: {
                name: "name",
                template: "filepaths",
              },
```

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+              description: "Get value from global ENV file",
+            },
+            {
+              name: "--env",
+              description:
+                "Get value from environment specific project ENV file",
+              requiresEquals: true,
+              args: {
+                name: "name",
+              },
```
SeparateRecords, line 603:
```suggestion
              args: {
                name: "name",
                template: "filepaths",
              },
```

src/fin.ts
```diff
@@ -0,0 +1,946 @@
...
+          },
+        },
+      ],
+    },
+    {
+      name: "exec",
+      description: "Execute a command or a script in cli",
+      args: {
+        name: "command|file",
+      },
```
SeparateRecords, line 617:
```suggestion
      args: {
        name: "command|file",
        isCommand: true,
      },
```
Will give correct suggestions if you provide a command name there!

SeparateRecords (changes requested):
Awesome job on this. Really killing it. It's a big spec so there's lots of places to add generators - most of the review comments I've left are things that can be worked on in a follow up. All the generator suggestions are just nice-to-haves.

There's a couple of easy improvements like adding `template: '...'` to some args which I think is important for good UX. So let's get that done for this PR at least, and if you want to do the generators then go for it! 🙂

src/fin.ts
XxHalbfettxX: I can write a generator for this, but there is no command into which the generated completions can be inserted. The rm command does not take an argument. It deletes all the keys at once.

src/fin.ts
XxHalbfettxX: The same as with the ssh-key ls command. I have not found a place where I can insert the vm as a completion. 

But it could also be that I lack experience with docksal

src/fin.ts
XxHalbfettxX: I have added a generator for the services.

src/fin.ts
```diff
@@ -359,14 +360,20 @@ const completionSpec: Fig.Spec = {
...
           name: "add",
           description:
-            "Add a private SSH key from $HOME/.ssh by file name. mAdds all default keys (id_rsa/id_dsa/id_ecdsa/id_ed25519) if no file name is given",
+            "Add a private SSH key from $HOME/.ssh by file name. Adds all default keys (id_rsa/id_dsa/id_ecdsa/id_ed25519) if no file name is given",
           args: {
             name: "key-name",
             isOptional: true,
+            generators: {
+              script:
+                "\\command ls .ssh | \\command grep --color=never -E '^id_'",
```
fedeci: I think \\command is wrong here, isn't it?

fedeci: Yeah, I meant which it should be used in a different way: `\\ls .ssh ...` is the correct one. It takes the original `ls` command and not eventual aliases the user may have create around it e.g. `alias ls="ls -ls"`

XxHalbfettxX: Hups :D I guess I didn't read that carefully enough :D

SeparateRecords: @fedeci actually `command` is correct - Bash has an `ls` builtin with different behavior, so using `command` means it will skip using the builtin and search the path! The reason `command` is "quoted" with `\\` is because users can also alias it, so we skip the alias lookup

fedeci: That's crazy😕

SeparateRecords (approved):
This is really fantastic. Sorry it's taken so long to merge this, been swamped with a million things to do. You've really nailed this spec @XxHalbfettxX :D

PR #1305, opened 2022-06-21
feat: add molecule completion spec

thmshmm:
New feature to support completion for https://molecule.readthedocs.io/en/latest/

### DIFF
--- a/src/molecule.ts
+++ b/src/molecule.ts
@@ -1,0 +1,301 @@
+const driverNameOption: Fig.Option = {
+  name: ["--driver-name", "-d"],
+  description: "Name of driver to use. (delegated)",
+  args: {
+    name: "TEXT",
+    suggestions: [
+      "delegated",
+      "docker",
+      "podman",
+      "vagrant",
+      "azure",
+      "hetzner",
+    ],
+  },
+};
+
+const scenarioNameOption: Fig.Option = {
+  name: ["--scenario-name", "-s"],
+  description: "Name of the scenario to target. (default)",
+  args: { name: "TEXT" },
+};
+
+const parallelOption: Fig.Option = {
+  name: "--parallel",
+  description: "Enable parallel mode",
+  exclusiveOn: ["--no-parallel"],
+};
+
+const noParallelOption: Fig.Option = {
+  name: "--no-parallel",
+  description: "Disable parallel mode",
+  exclusiveOn: ["--parallel"],
+};
+
+const dependencyNameOption: Fig.Option = {
+  name: "--dependency-name",
+  description: "Name of dependency to initialize. (galaxy)",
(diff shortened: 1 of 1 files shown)


src/molecule.ts
```diff
@@ -0,0 +1,284 @@
...
+  args: { name: "TEXT" },
+};
+
+const helpOption: Fig.Option = {
+  name: "--help",
+  description: "Show help",
+};
+
+const parallelOption: Fig.Option = {
+  name: ["--parallel", "--no-parallel"],
```
fedeci, line 29: These two options need to be separated. You can also add `exclusiveOn` to the two options.

src/molecule.ts
```diff
@@ -0,0 +1,284 @@
...
+
+const completionSpec: Fig.Spec = {
+  name: "molecule",
+  description: "Molecule aids in the development and testing of Ansible roles",
+  subcommands: [
+    {
+      name: "check",
+      description:
+        "Use the provisioner to perform a Dry-Run (destroy, dependency, create, prepare, converge)",
+      options: [scenarioNameOption, parallelOption, helpOption],
```
fedeci, line 77: Help option can be added just once in the root subcommand along with the `isPersistent: true` prop.

src/molecule.ts
```diff
@@ -0,0 +1,284 @@
...
+      options: [scenarioNameOption, helpOption],
+    },
+    {
+      name: "destroy",
+      description: "Use the provisioner to destroy the instances",
+      options: [
+        scenarioNameOption,
+        driverNameOption,
+        {
+          name: ["--all", "--no-all"],
```
fedeci, line 108: These should separated too

src/molecule.ts
```diff
@@ -0,0 +1,284 @@
...
+    },
+    {
+      name: "test",
+      description:
+        "Test (dependency, lint, cleanup, destroy, syntax, create, prepare, converge, idempotence, side_effect, verify, cleanup, destroy)",
+      options: [
+        scenarioNameOption,
+        driverNameOption,
+        {
+          name: ["--all", "--no-all"],
```
fedeci, line 258: These too

src/molecule.ts
```diff
@@ -0,0 +1,284 @@
...
+    },
+    {
+      name: "prepare",
+      description:
+        "Use the provisioner to prepare the instances into a particular starting state",
+      options: [
+        scenarioNameOption,
+        driverNameOption,
+        {
+          name: ["--force", "--no-force"],
```
fedeci, line 228: Same here

fedeci (approved):
Thanks! This spec looks really good apart for a couple issues I annoted, can you fix those please?

PR #1308, opened 2022-06-25
wip: add completions for workflows

mschrage:
Waiting on automatically generated completion spec for `fig`


SeparateRecords:
Looks good! Is src/fig2.ts supposed to exist? Looks like what I do when working on a big change and want to compare side-by-side with the old version 😄

mschrage:
Thnaks for the review! `fig2` should definitely not exist.

PR #1312, opened 2022-06-28
feat: add `secli`

AnishDe12020:
new spec for [secli](https://crates.io/crates/secli)

**Additional info:**
I have generated the spec using the clap fig crate but as I am on linux, wasn't able to test it

### DIFF
--- a/src/secli.ts
+++ b/src/secli.ts
@@ -1,0 +1,103 @@
+const completion: Fig.Spec = {
+  name: "secli",
+  description:
+    "A simple CLI to store secrets locally and retrieve them via the CLI",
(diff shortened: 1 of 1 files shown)


fedeci:
recheck

fedeci:
I am sorry for closing this PR but this CLI tool isn't safe for storing secrets. Operating systems offer encrypted built in alternatives to just a sqlite db (like keychain on MacOS) that are accessible through APIs. I know this is like censoring the tool and that is really the last thing I want, but I believe we can keep some users safe by not merging this. Sorry again, this is nothing of personal!

PR #1313, opened 2022-06-28
Fastly CLI 3.1.0

Integralist:

Updates the https://github.com/fastly/cli autocomplete spec to the latest `3.1.0` release.

The current version in this repo is `2.0.0`.


kailan:
thanks for doing this @Integralist –– it looks like the pre-commit formatter didn't run correctly because the diff is *huge*.

If you run `npm run lint:fix` it will resolve this and the diff should be more like what we saw in the figly repo: https://github.com/kailan/figly/commit/3237f16ad7fd7095f85b78b669fbe22368a01c04

kailan (approved):
✨ perfect 👍

mschrage (approved):

src/fastly.ts
```diff
@@ -11111,7 +11247,7 @@ const completionSpec: Fig.Spec = {
                 },
                 {
                   name: "--address",
-                  description: "The hostname or IPv4 address",
+                  description: "The hostname or IPv4 addres",
```
SeparateRecords, line 11250: Looks like this should be fixed upstream, but for now could you edit manually?

SeparateRecords (approved):
Looks good! Most of this diff is adding empty newlines, which can also be removed. If you're a vim user, should be as easy as `:%s/,\n\n/,\r/g` (replace comma followed by two newlines with a command and newline)

src/fastly.ts
Integralist: Fixed.

Integralist:
@SeparateRecords regarding your comment...

> Looks good! Most of this diff is adding empty lines, which can also be removed. If you're a vim user, should be as easy as :%s/,\n\n/,\r/g (replace comma followed by two newlines with a comma and newline).

I am a [vim user](https://link.springer.com/book/10.1007/978-1-4842-0250-0)! 😉

I've now replaced all the line breaks. I think the tool I used to generate the updated file (https://github.com/kailan/figly) might have added those by accident.

SeparateRecords:
Awesome, appreciate the changes. LGTM & thanks for updating the spec!!

(Also, that's incredible - I had no idea that was your book, I think I saw it recommended years ago while I was still in the steep part of the Vim learning curve!)

fedeci:
I think we can integrate oss projects like fastly directly, and use diff versioned specs for them!

PR #1314, opened 2022-06-28
feat(fig): update spec

mschrage:


fedeci:
Mh, this generates linting issues, likely it is just a wrong config for diff versioned specs.

PR #1317, opened 2022-06-29
feat: add `meteor` completion spec

getninjaN:
New spec for [meteor](https://www.meteor.com/)

There is no spec for meteor

I went through the entire Meteor CLI and added them to the spec.

**Additional info:**
There already is a pull request (#1009) but that one only adds the npm part, so I figured I'd create the entire thing.


src/meteor.ts
```diff
@@ -0,0 +1,1073 @@
...
+      loadSpec: "npm",
+    },
+    {
+      name: "help",
+      description: "Get help on meteor command line usage",
+      args: {
+        name: "command",
+        description: "Built-in command",
+        isOptional: true,
+        suggestions: [
```
fedeci, line 57: Instead of manually adding suggestions here you can use the `help` template.

src/meteor.ts
```diff
@@ -0,0 +1,1073 @@
+import { filepaths } from "@fig/autocomplete-generators";
+
+const examplesGenerator: Fig.Generator = {
+  script: "meteor create --list",
+  postProcess: (output) => {
+    return output.split("\n").map((example) => {
+      if (example.includes("github.com")) {
+        return { name: example.split(":")[0].trim() };
+      }
+    });
```
fedeci, line 10: `.map` must always return some value. You can use `.reduce` or `.filter().map()` here.

src/meteor.ts
```diff
@@ -0,0 +1,1073 @@
...
+            name: "command",
+            suggestions: [
+              "maintainers",
+              "recommend-release",
+              "change-homepage",
+              "list-organizations",
+              "members",
+            ],
+          },
+        },
```
fedeci, line 102: These should be added as subcommands

src/meteor.ts
```diff
@@ -0,0 +1,1073 @@
...
+          name: ["--port", "-p"],
+          description: "Port to listen on (instead of the default 3000)",
+          args: {
+            name: "port",
+            isOptional: true,
+            default: "3000",
+          },
+        },
+        {
+          name: ["--inspect", "--inspect-brk"],
```
fedeci, line 125: In node these two flags are different, it seems it is the same for meteor

src/meteor.ts
```diff
@@ -0,0 +1,1073 @@
...
+          description: "Create a basic blaze-based app",
+        },
+        {
+          name: "--tailwind",
+          description:
+            "Create a basic react-based app, with tailwind configured",
+        },
+        {
+          name: "path",
+        },
```
fedeci, line 288: Is this a flag or an argument?

fedeci (dismissed):
Thanks @getninjaN! I left some comments to fix before merging this spec!

getninjaN:
> Thanks @getninjaN! I left some comments to fix before merging this spec!

Great! I'll take a look at it asap.

getninjaN:
Changes has been made. 🍻

fedeci (approved):
Thanks!

PR #1323, opened 2022-07-01
Update For ng 14

mreouven:
Bug fix

There is no completion for ng.ts

In the angular version 14 its moved the version checking 

**Additional info:**

check https://github.com/angular/angular-cli/blob/8da926966e9f414ceecf60b89acd475ce1b55fc5/packages/angular/cli/src/commands/version/cli.ts

### DIFF
--- a/src/ng.ts
+++ b/src/ng.ts
@@ -127,8 +127,14 @@ const completionSpec: Fig.Spec = {
     },
     {
       name: "version",
-      description: "View your Angular CLI version",
+      description: "View your Angular CLI version (update for Angular 14+)",
     }
+  ],
+  options: [
+    {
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
Thanks for the PR, I would still keep the option because not everyone may be using this angular version.

mreouven:
> Thanks for the PR, I would still keep the option because not everyone may be using this angular version.

I add the possibility of using both options 
The problem is that in version 14 to have the version with the use of the flag create an error 
`Error: You need to specify a command before moving on. Use '--help' to view the available commands.`

fedeci (approved):
Thanks!

PR #1327, opened 2022-07-04
feat: add `bw` BitWarden CLI

SeparateRecords:
This is the initial PR that just adds it to this repo.

BitWarden is using commander.js v7.2.0 which the Fig integration is incompatible with, that needs to be upgraded to at least 8.0.0 before we can submit a PR for integration


fedeci:
Commander 7 is practically identical to 8. The integration works even with the commander version used by babel which is v3 if I am not wrong!

yyyyyae:
Sent from my Metro by T-Mobile 5G Device
-------- Original message --------From: Federico Ciardi ***@***.***> Date: 7/4/22  4:42 AM  (GMT-06:00) To: withfig/autocomplete ***@***.***> Cc: Subscribed ***@***.***> Subject: Re: [withfig/autocomplete] feat: add `bw` BitWarden CLI (PR #1327) 
Commander 7 is practically identical to 8. The integration works even with the commander version used by babel which is v3 if I am not wrong!

src/bw.ts
```diff
@@ -0,0 +1,829 @@
...
+        {
+          name: ["-h", "--help"],
+          description: "Display help for command",
+          priority: 49,
+        },
+      ],
+    },
+    {
+      name: "share",
+      description: "--DEPRECATED-- Move an item to an organization",
```
fedeci, line 518:
```suggestion
      description: "--DEPRECATED-- Move an item to an organization",
      deprecated: true,
```

fedeci (approved):
Thanks!

PR #1332, opened 2022-07-06
feat(php): add bin/console (Symfony) command

matthiashamacher:
Feature

Adding completions for Symfony bin/console

### DIFF
--- a/src/php/bin-console.ts
+++ b/src/php/bin-console.ts
@@ -3,21 +3,18 @@ const completionSpec: Fig.Spec = {
   description: "Symfony bin/console command",
   generateSpec: async (tokens, executeShellCommand) => {
     const out = await executeShellCommand("php bin/console list --format=json");
-    const subcommands = [];
+    let subcommands = [];
 
     try {
       const commandDefinitions = JSON.parse(out);
 
-      commandDefinitions.commands.map((command) => {
-        if (command.name === "_complete") {
-          return;
-        }
-
-        subcommands.push({
+      subcommands = commandDefinitions.commands
+        .filter((command) => command.name !== "_complete")
+        .map((command) => ({
           name: command.name,
           description: command.description,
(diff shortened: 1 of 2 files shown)


src/php/bin-console.ts
```diff
@@ -0,0 +1,56 @@
...
+  generateSpec: async (tokens, executeShellCommand) => {
+    const out = await executeShellCommand("php bin/console list --format=json");
+    const subcommands = [];
+
+    try {
+      const commandDefinitions = JSON.parse(out);
+
+      commandDefinitions.commands.map((command) => {
+        if (command.name === "_complete") {
+          return;
```
fedeci, line 13: Map should always return some value, cau you run a filter before or combine both the map and the filter in a reduce statement?

src/php/bin-console.ts
```diff
@@ -0,0 +1,56 @@
...
+          icon: "https://cdn.iconscout.com/icon/free/png-128/symfony-282493.png",
+          args: Object.keys(command.definition.arguments).map((argumentKey) => {
+            const argument = command.definition.arguments[argumentKey];
+
+            return {
+              name: argument.name,
+              description: argument.description,
+              isOptional: !argument.is_required,
+            };
+          }),
```
fedeci, line 28: nit: you can directly map through `command.definition.arguments`.
```js
command.definition.arguments.map(argument => {...})
```

src/php/bin-console.ts
```diff
@@ -0,0 +1,56 @@
...
+          options: Object.keys(command.definition.options).map((optionKey) => {
```
fedeci, line 29: Same here

fedeci (changes requested):
Thanks, there are a couple of nits but the spec looks awesome!

src/php/bin-console.ts
matthiashamacher: Done.

fedeci (approved):
Thanks!

PR #1335, opened 2022-07-06
[z] Added generated path based on navigation history

Rawnly:
Feature

Autocomplete on 'z' is shown using the current path.

It uses navigation history to autocomplete the path (it makes more sense since `z` is used for jumping into recent directories without the need of writing the full path)

**Additional info:**
CleanShot 2022-07-07 at 00 36 02@2x

### DIFF
--- a/src/z.ts
+++ b/src/z.ts
@@ -3,10 +3,29 @@ const completionSpec: Fig.Spec = {
   name: "z",
   description: "CLI tool to jump around directories",
   args: {
-    template: ["folders"],
     name: "regex",
     isVariadic: true,
     isOptional: true,
+    // Generate a completion for the given argument, via zoxide CLI.
+    generators: {
+        script: context => {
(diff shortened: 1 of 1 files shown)


src/z.ts
```diff
@@ -3,10 +3,29 @@ const completionSpec: Fig.Spec = {
...
+            const shortPath = fullPath.replace(homeRegex, '~') // .split('/').slice(3).join('/')
+
+            return {
+              name: shortPath,
+              description: fullPath,
+              displayName: shortPath
+            };
+          });
+        }
+      }
```
fedeci, line 28: This is a generator for zoxide, I understand some people are aliasing it to z but we have alias resolving that should work so I would prefer to investigate that issue before merging this PR.

Rawnly: Absolutely, this is why I created a draft, then I realised that this Z and the one I'm using via zoxide aren't the same.

PR #1337, opened 2022-07-07
feat: add appwrite

BalliAsghar:

Adding completion spec for **appwrite**  - Open-Source End-to-End Backend Server

**Additional info:**


src/appwrite.ts
```diff
@@ -0,0 +1,5882 @@
...
+  subcommands: [
+    {
+      name: "login",
+      description:
+        "The login command allows you to authenticate and manage a user account",
+      icon,
+
+      options: [
+        {
+          name: ["-h", "--help"],
```
fedeci, line 15: You can add isPersistent to the `--help` option in the root of the spec and it will be added by default to all the subcommands, this should make the file a bit smaller.

fedeci (approved):
Thanks! This is super organized and was super smooth to review! I think we should use this PR as an example in our contributing guidelines! Can you fix the only comment I left please?

fedeci:
Can you also run `npm run lint:fix`? That was my fault merging a PR that was not passing lint.

src/appwrite.ts
BalliAsghar: done!

fedeci (approved):
Thanks again!

PR #1339, opened 2022-07-08
feat: add `create-t3-app` spec

fernandoxu:
Feature add [create-t3-app](https://create.t3.gg/)

### DIFF
--- a/src/create-t3-app.ts
+++ b/src/create-t3-app.ts
@@ -1,3 +1,12 @@
+const dirArgument: Fig.Arg = {
+  name: "dir",
+  description:
+    "The name of the application, as well as the name of the directory to create",
+  template: "folders",
+  isOptional: true,
+  suggestCurrentToken: true,
+};
+
 const completionSpec: Fig.Spec = {
   name: "create-t3-app",
   description: "A CLI for creating web applications with the t3 stack",
@@ -18,6 +27,7 @@ const completionSpec: Fig.Spec = {
       description: "Display help for command",
     },
   ],
+  args: dirArgument,
 };
 
 export default completionSpec;
--- a/src/create-t3-app.ts
+++ b/src/create-t3-app.ts
@@ -1,13 +1,5 @@
-const dirArgument: Fig.Arg = {
-  name: "dir",
-  description:
-    "The name of the application, as well as the name of the directory to create",
-  template: "folders",
-  isOptional: true,
-};
-
 const completionSpec: Fig.Spec = {
-  name: "create-t3-app@latest",
+  name: "create-t3-app",
   description: "A CLI for creating web applications with the t3 stack",
   icon: "https://create.t3.gg/favicon.svg",
   options: [
@@ -26,7 +18,6 @@ const completionSpec: Fig.Spec = {
       description: "Display help for command",
     },
   ],
-  args: dirArgument,
 };
 
 export default completionSpec;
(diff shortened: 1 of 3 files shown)


src/create-t3-app@latest.ts
```diff
@@ -0,0 +1,32 @@
+const dirArgument: Fig.Arg = {
+  name: "dir",
+  description:
+    "The name of the application, as well as the name of the directory to create",
+  template: "folders",
+  isOptional: true,
+};
+
+const completionSpec: Fig.Spec = {
+  name: "create-t3-app@latest",
```
fedeci, line 10: Can you remove @latest please (from the filename also)? Eventually if you are used to run this CLI with the trailing @latest you can create another file named `src/create-t3-app@latest.ts` that contains the following:
```ts
import create_t3_app from "./create_t3_app";

  ...create_t3_app,
  name: "create_t3_app@latest",
};

export default completionSpec;
```

src/create-t3-app@latest.ts
```diff
@@ -0,0 +1,32 @@
+const dirArgument: Fig.Arg = {
+  name: "dir",
+  description:
+    "The name of the application, as well as the name of the directory to create",
+  template: "folders",
```
fedeci, line 5: Do folders need to exist or they may not exist?

src/npx.ts
```diff
@@ -75,6 +75,10 @@ const suggestions: Fig.Suggestion[] = [
     name: "create-next-app",
     icon: "https://nextjs.org/static/favicon/favicon-16x16.png",
   },
+  {
+    name: "create-t3-app@latest",
+    icon: "https://create.t3.gg/favicon.svg",
+  },
```
fedeci, line 81: Here you should duplicate the suggestion adding "create-t3-app" also.

fedeci (changes requested):
Thanks! Great first spec!

src/create-t3-app@latest.ts
fernandoxu: OK! thanks a lot for your check!

src/create-t3-app@latest.ts
SeparateRecords: IIRC you're correct, folder doesn't need to exist

You can add `suggestCurrentToken: true` to make the behavior correct! (I think that's the name of the prop...)

src/create-t3-app.ts
```diff
@@ -0,0 +1,23 @@
...
+      name: ["-y", "--default"],
+      description:
+        "Bypass the CLI and use all default options to bootstrap a new t3-app (default: false)",
+    },
+    {
+      name: ["--help", "-h"],
+      description: "Display help for command",
+    },
+  ],
+};
```
fedeci, line 21: This wasn't to remove, we have options for everything! `suggestCurrentToken` was what this argument needed.
```suggestion
const dirArgument: Fig.Arg = {
  name: "dir",
  description:
    "The name of the application, as well as the name of the directory to create",
  template: "folders",
  isOptional: true,
  suggestCurrentToken: true,
};

const completionSpec: Fig.Spec = {
  name: "create-t3-app",
  description: "A CLI for creating web applications with the t3 stack",
  icon: "https://create.t3.gg/favicon.svg",
  options: [
    {
      name: "--noGit",
      description:
        "Explicitly tell the CLI to not initialize a new git repo in the project (default: false)",
    },
    {
      name: ["-y", "--default"],
      description:
        "Bypass the CLI and use all default options to bootstrap a new t3-app (default: false)",
    },
    {
      name: ["--help", "-h"],
      description: "Display help for command",
    },
  ],
  args: dirArgument,
};
```

fedeci (changes requested):

fedeci (approved):
Thanks again!

PR #1342, opened 2022-07-12
fix(ruby): update a description to be grammatically correct

gipcompany:

This PR is just a grammatical error fix.

No issues created yet because this seems a very tiny thing.

You can see like below that this is correct description.

スクリーンショット 2022-07-12 22 09 56

**Additional info:**

For your information, the Ruby man says like below.

https://github.com/ruby/ruby/blob/067a5f1a0024ebd270f363cd440147da9dc176b5/man/ruby.1#L212-L213

### DIFF
--- a/src/ruby.ts
+++ b/src/ruby.ts
@@ -40,7 +40,7 @@ const completionSpec: Fig.Spec = {
     {
       name: "-S",
       description:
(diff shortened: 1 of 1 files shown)


fedeci (approved):
Thanks!

PR #1344, opened 2022-07-13
fix(cargo): use API for search instead of `cargo`

SeparateRecords:


src/cargo.ts
```diff
@@ -41,21 +39,28 @@ const binList: Fig.Generator = {
...
+  newest_version: string;
+}
+
 const searchGenerator: Fig.Generator = {
   script: function (context) {
-    if (context[context.length - 1] === "") return "";
-    const searchTerm = context[context.length - 1];
-    return `cargo search "${searchTerm}" | \\grep -E "^\\w"`;
+    const query = encodeURIComponent(context[context.length - 1]);
+    return `curl -sfL 'https://crates.io/api/v1/crates?q=${query}&per_page=60'`;
```
fedeci, line 55: Are 60 results really what we want? I mean I think 10-15 may be enough, it's difficult someone ever goes that low.

src/cargo.ts
```diff
@@ -3923,6 +3928,8 @@ const completionSpec: Fig.Spec = {
       args: {
         name: "query",
         isVariadic: true,
+        generators: searchGenerator,
```
fedeci, line 3931: Can you add debounce to this argument?

fedeci (changes requested):
Two things to fix and you can merge this!

src/cargo.ts
SeparateRecords: 🤨 I thought I did... wack

src/cargo.ts
SeparateRecords: Yeah the reason for 60 is that the API isn't guaranteed to return results that actually start with the search term, it also fuzzy matches on the description and rest of the title. I found 60 to be a good middle ground for non-fuzzy-search testing but I can definitely lower it too

fedeci: It sounds good to have 60 then!

fedeci:
Lgtm!

fedeci (approved):

PR #1345, opened 2022-07-14
feat(docker-compose): update to v2

SeparateRecords:
~~This seems like a good opportunity for version based specs, will implement this before merge~~

### DIFF
--- a/src/docker-compose.ts
+++ b/src/docker-compose.ts
@@ -1,5 +1,3 @@
-import { filepaths } from "@fig/autocomplete-generators";
-
 const servicesGenerator: Fig.Generator = {
   script: (tokens) => {
     const files: string[] = [];
@@ -15,40 +13,48 @@ const servicesGenerator: Fig.Generator = {
   splitOn: "\n",
 };
 
-const completionSpec: Fig.Spec = {
+const compose: Fig.Spec = {
   name: "docker-compose",
-  description: "Define and run multi-container applications with Docker",
+  description: "Docker Compose",
   subcommands: [
     {
       name: "build",
       description: "Build or rebuild services",
       args: {
-        name: "service",
-        isOptional: true,
+        name: "services",
         isVariadic: true,
+        isOptional: true,
         generators: servicesGenerator,
       },
       options: [
         {
           name: "--build-arg",
           description: "Set build-time variables for services",
+          isRepeatable: true,
           args: {
-            name: "key=val",
+            name: "key=value",
+            generators: servicesGenerator,
           },
         },
         {
           name: "--compress",
-          description: "Compress the build context using gzip",
+          description: "Compress the build context using gzip. DEPRECATED",
+          hidden: true,
+          deprecated: true,
         },
         {
           name: "--force-rm",
-          description: "Always remove intermediate containers",
+          description: "Always remove intermediate containers. DEPRECATED",
+          hidden: true,
+          deprecated: true,
         },
         {
-          name: ["-m", "--memory"],
-          description: "Set memory limit for the build container",
+          name: ["--memory", "-m"],
+          description:
+            "Set memory limit for the build container. Not supported on buildkit yet",
+          hidden: true,
           args: {
-            name: "MEM",
+            name: "memory",
           },
         },
         {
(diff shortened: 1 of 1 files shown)


src/docker-compose.ts
```diff
@@ -15,40 +13,48 @@ const servicesGenerator: Fig.Generator = {
   splitOn: "\n",
 };
 
-const completionSpec: Fig.Spec = {
+const compose: Fig.Spec = {
   name: "docker-compose",
-  description: "Define and run multi-container applications with Docker",
+  description: "Docker Compose",
```
SeparateRecords, line 18:
```suggestion
  description: "Define and run multi-container applications with Docker",
```

src/docker-compose.ts
```diff
@@ -202,73 +280,79 @@ const completionSpec: Fig.Spec = {
...
-            isVariadic: true,
+            name: "user",
           },
         },
         {
-          name: ["-w", "--workdir"],
-          description: "DIR Path to workdir directory for this command",
+          name: ["--workdir", "-w"],
+          description: "Path to workdir directory for this command",
           args: {
```
SeparateRecords, line 338: This should have `template: "folders"`

src/docker-compose.ts
```diff
@@ -295,34 +382,81 @@ const completionSpec: Fig.Spec = {
...
+          },
         },
         {
           name: "--tail",
           description:
             "Number of lines to show from the end of the logs for each container",
           args: {
-            name: "Num of Lines",
-            suggestions: ["all"],
+            name: "tail",
```
SeparateRecords, line 415:
```suggestion
            name: "lines",
```

src/docker-compose.ts
```diff
@@ -295,34 +382,81 @@ const completionSpec: Fig.Spec = {
...
-          description: "Don't print prefix in logs",
+          name: ["--timestamps", "-t"],
+          description: "Show timestamps",
+        },
+        {
+          name: "--until",
+          description:
+            "Show logs before a timestamp (e.g. 2013-01-02T13:23:37Z) or relative (e.g. 42m for 42 minutes)",
+          args: {
+            name: "until",
```
SeparateRecords, line 428:
```suggestion
            name: "timestamp",
```

src/docker-compose.ts
```diff
@@ -331,38 +465,34 @@ const completionSpec: Fig.Spec = {
...
         { name: "private_port" },
       ],
       options: [
         {
-          name: "--protocol",
-          description: "Tcp or udp [default: tcp]",
+          name: "--index",
+          description:
+            "Index of the container if service has multiple replicas",
           args: {
```
SeparateRecords, line 485: Should have `suggestions: ["tcp", "udp"]`

src/docker-compose.ts
```diff
@@ -15,40 +13,48 @@ const servicesGenerator: Fig.Generator = {
   splitOn: "\n",
 };
 
-const completionSpec: Fig.Spec = {
+const compose: Fig.Spec = {
```
SeparateRecords, line 16:
```suggestion
const completionSpec: Fig.Spec = {
```

src/docker-compose.ts
```diff
@@ -1,5 +1,3 @@
-import { filepaths } from "@fig/autocomplete-generators";
-
 const servicesGenerator: Fig.Generator = {
```
SeparateRecords, line 1: May want to update this?

SeparateRecords:
@fedeci ready for review - decided not to go ahead with version based spec.

Reason is that with v2 there are two ways of running docker compose: `docker-compose` and `docker compose`
Version based specs can't determine which to load based on the command being run, so the "solution" is a really ugly hack: have two version-based specs, one for docker-compose directly and the other loaded by `docker`

fedeci:
> Reason is that with v2 there are two ways of running docker compose: `docker-compose` and `docker compose` Version based specs can't determine which to load based on the command being run, so the "solution" is a really ugly hack: have two version-based specs, one for docker-compose directly and the other loaded by `docker`

Isn't `loadSpec` properly working?

PR #1349, opened 2022-07-16
feat(gh): add browse core command

johnmerrill:
feature

`gh` command `browse` is not supported with autocomplete

`gh` command `browse` will be an option

**Additional info:**
`gh browse` documentation https://cli.github.com/manual/gh_browse

### DIFF
--- a/src/gh.ts
+++ b/src/gh.ts
@@ -372,6 +372,49 @@ const completionSpec: Fig.Spec = {
       ],
     },
     {
+      name: "browse",
+      description: "Open the repository in the browser",
+      options: [ghOptions.help],
+      subcommands: [
+        {
+          name: ["-b", "--branch"],
+          description: "Select another branch by passing in the branch name",
+          args: {
+            name: "branch",
+            generators: ghGenerators.remoteBranches,
+          },
+        },
+        {
+          name: ["-c", "--commit"],
+          description: "Select another branch by passing in the branch name",
+        },
+        {
+          name: ["-n", "--no-browser"],
+          description: "Print destination URL instead of opening the browser",
+        },
+        {
+          name: ["-p", "--projects"],
+          description: "Open repository projects",
+        },
+        {
+          name: ["-R", "--repo"],
+          description:
+            "Select another repository using the [HOST/]OWNER/REPO format",
+          args: {
+            name: "[HOST/]OWNER/REPO",
+          },
+        },
+        {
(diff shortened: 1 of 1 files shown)


src/gh.ts
```diff
@@ -371,6 +371,49 @@ const completionSpec: Fig.Spec = {
...
+          name: ["-b", "--branch"],
+          description: "Select another branch by passing in the branch name",
+          args: {
+            name: "branch",
+            generators: ghGenerators.remoteBranches,
+          },
+        },
+        {
+          name: ["-c", "--commit"],
+          description: "Select another branch by passing in the branch name",
```
fedeci, line 389:
```suggestion
          description: "Open the last commit",
```

src/gh.ts
```diff
@@ -371,6 +371,49 @@ const completionSpec: Fig.Spec = {
         },
       ],
     },
+    {
+      name: "browse",
+      description: "Open the repository in the browser",
+      options: [ghOptions.help],
+      subcommands: [
```
fedeci, line 378: All these are options and should be moved to the `options` property array

src/gh.ts
```diff
@@ -371,6 +371,49 @@ const completionSpec: Fig.Spec = {
         },
       ],
     },
+    {
+      name: "browse",
+      description: "Open the repository in the browser",
```
fedeci, line 376: The browse command also accepts arguments, we may provide generators for this but I think we can tackle that in another PR
```suggestion
      description: "Open the repository in the browser",
      args: {
        name: "pr | issue | path[:line]"
      },
```

fedeci (changes requested):
Thanks for the PR 🎉, can you solve the comments please?

src/gh.ts
johnmerrill: I added the existing generator `ghGenerators.listPR` to this argument. Not sure how you feel about using a generator that only satisfies part of the argument.

src/gh.ts
```diff
@@ -371,6 +371,53 @@ const completionSpec: Fig.Spec = {
...
+      args: {
+        name: "[pr | issue | path[:line]",
+        generators: ghGenerators.listPR,
```
fedeci, line 379: I'm fine with this as long as this also gets added:
```suggestion
        generators: ghGenerators.listPR,
        suggestCurrentToken: true,
```

fedeci (approved):
Thanks!

PR #1350, opened 2022-07-16
feat: add hop

Looskie:
Adding completion spec for Hop's CLI  - Interact with Hop in your terminal

**Additional info:**

### DIFF
--- a/src/hop.ts
+++ b/src/hop.ts
@@ -132,6 +132,10 @@ const completionSpec: Fig.Spec = {
       ],
     },
     {
+      name: "link",
+      description: "Link an existing deployment to a hopfile",
+    },
+    {
       name: "projects",
       description: "Interact with projects",
       subcommands: [
--- a/src/hop.ts
+++ b/src/hop.ts
@@ -33,35 +33,66 @@ const completionSpec: Fig.Spec = {
         {
           name: ["--type", "-t"],
           description: "Type of the container, defaults to `ephemeral`",
+          args: {
+            name: "container type",
+            suggestions: ["ephemeral", ""],
+          },
         },
         {
           name: ["--containers", "-i"],
           description:
             "Number of containers to use, defaults to 1 if `scaling` is manual",
+          args: {
+            name: "container amount",
+            suggestions: ["1", "2", "4", "8", "10"],
+          },
         },
         {
           name: ["--cpu", "-c"],
           description: "The number of CPUs to use, defaults to 1",
+          args: {
+            name: "cpu amount",
+            suggestions: ["1", "2", "4", "8"],
+          },
         },
         {
           name: ["--env", "-e"],
           description: "Environment variables to set, in the form of KEY=VALUE",
+          args: {
+            name: "environment variable",
+            suggestions: ["port="],
+          },
         },
         {
           name: ["--name", "-n"],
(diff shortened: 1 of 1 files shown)


Looskie:
👍

src/hop.ts
```diff
@@ -0,0 +1,173 @@
...
+      description: "Deploy a new container",
+      args: {
+        name: "dir",
+        isOptional: true,
+        description: "Directory to deploy, defaults to current directory",
+      },
+      options: [
+        {
+          name: ["--type", "-t"],
+          description: "Type of the container, defaults to `ephemeral`",
```
fedeci, line 35:
```suggestion
          description: "Type of the container, defaults to `ephemeral`",
          args: {
            name: "container type",
          },
```

src/hop.ts
```diff
@@ -0,0 +1,173 @@
...
+        },
+        {
+          name: ["--ram", "-m"],
+          description: "Amount of RAM to use, defaults to 512MB",
+        },
+        {
+          name: ["--scaling", "-s"],
+          description: "Scaling strategy, defaults to `manual`",
+        },
+      ],
```
fedeci, line 66: All of these need an argument, see a simple example above, I would suggest you to give a look to this property too that may be useful [`suggestions`](https://fig.io/docs/reference/arg#suggestions), let me know if you need help!

src/hop.ts
```diff
@@ -0,0 +1,173 @@
...
+        },
+        {
+          name: "secrets",
+          description: "Get help on project secrets",
+        },
+        {
+          name: "whoami",
+          description: "Get help on whoami",
+        },
+      ],
```
fedeci, line 96: We have some nice sugar to avoid having to maintain this list, the `help` template which creates some argument suggestions in a smart way!
```suggestion
      args: {
        name: "command",
        template: "help,"
      },
```

fedeci (changes requested):
Thanks for this first spec 🎉! I left a couple comments to solve before merging.

src/hop.ts
Looskie: oh this is so beautiful

src/hop.ts
Looskie: added! thank you for the `suggestions` tip, thats going to be REALLY handy :).

Looskie:
Thanks for the speedy review! Let me know if theres anything else with these changes :)

src/hop.ts
```diff
@@ -0,0 +1,182 @@
...
+            suggestions: ["1", "2", "4", "8"],
+          },
+        },
+        {
+          name: ["--env", "-e"],
+          description: "Environment variables to set, in the form of KEY=VALUE",
+          args: {
+            name: "environment variable",
+            suggestions: ["port="],
+          },
```
fedeci, line 64: @SeparateRecords has key value generators been merged?

fedeci (approved):
Love it! This really improves the spec a lot! Thanks you so much @Looskie

Looskie:
Hi @fedeci ! I've added a small command to the spec, hopefully this isn't a burden!

src/hop.ts
SeparateRecords: keyValue works fine but there's no generator for this so no point to adding it :)

SeparateRecords:
lgtm!

PR #1355, opened 2022-07-20
feat(git): don't insert leading `remotes/.../` in branch name

SeparateRecords:
This is a fairly common complaint, finally fixing it


fedeci (approved):
Nice

PR #1363, opened 2022-07-24
feat: add `ykman` spec

vaaleyard:
feature, new spec for `ykman`
Done

**Additional info:**
This is the yubikey-manager CLI: https://github.com/Yubico/yubikey-manager


src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+      description: "Enable or disable applications",
+      subcommands: [
+        {
+          name: "mode",
+          description: "Manage connection modes (USB Interfaces)",
+          args: {
+            name: "MODE",
+            description:
+              "MODE can be a string, such as 'OTP+FIDO+CCID', or a shortened form: 'o+f+c'. It can also be a mode number",
+          },
```
SeparateRecords, line 52: Would love to add suggestions to this. Because it takes predefined values but they're separated by `+` we need to use a generator:
```suggestion
          args: {
            name: "MODE",
            description:
              "MODE can be a string, such as 'OTP+FIDO+CCID', or a shortened form: 'o+f+c'. It can also be a mode number",
            generators: {
              getQueryTerm: "+",
              custom: () => Promise.resolve(["OTP", "FIDO", "CCID"]),
            }
          },
```

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+              name: ["-f", "--force"],
+              description: "Confirm the action without prompting",
+              isDangerous: true,
+            },
+            {
+              name: ["-e", "--enable"],
+              description: "Enable applications",
+              args: {
+                name: "OTP|U2F|FIDO2|OATH|PIV|OPENPGP|HSMAUTH",
+              },
```
SeparateRecords, line 100:
```suggestion
              args: {
                name: "application",
                suggestions: ["OTP", "U2F", "FIDO2", "OATH", "PIV", "OPENPGP", "HSMAUTH"],
              },
```

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+            },
+            {
+              name: ["-d", "--disable"],
+              description: "Disable applications",
+              args: {
+                name: "OTP|U2F|FIDO2|OATH|PIV|OPENPGP|HSMAUTH",
+              },
```
SeparateRecords, line 107:
```suggestion
              args: {
                name: "application",
                suggestions: ["OTP", "U2F", "FIDO2", "OATH", "PIV", "OPENPGP", "HSMAUTH"],
              },
```

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+              name: ["-f", "--force"],
+              description: "Confirm the action without prompting",
+              isDangerous: true,
+            },
+            {
+              name: ["-e", "--enable"],
+              description: "Enable applications",
+              args: {
+                name: "OTP|U2F|FIDO2|OATH|PIV|OPENPGP|HSMAUTH",
+              },
```
SeparateRecords, line 186: Suspecting there might be a couple more of these - could you replace all of them with the suggestions?

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+                    default: "6",
+                  },
+                },
+                {
+                  name: ["-a", "--algorithm"],
+                  description: "Algorithm to use for code generation",
+                  args: {
+                    name: "SHA1|SHA256|SHA512",
+                    default: "SHA1",
+                  },
```
SeparateRecords, line 660: Same thing, probably easiest to just search for all instances of `|` and add the different items as `suggestions`

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+                },
+                {
+                  name: ["-p", "--protect"],
+                  description:
+                    "Store new management key on the YubiKey, protected by PIN. A random key will be used if no key is provided",
+                },
+                {
+                  name: ["-g", "--generate"],
+                  description:
+                    "Generate a random management key. Implied by --protect unless --new-management-key is also given. Conflicts with --new-management-key",
```
SeparateRecords, line 1632: Can you add an `exclusiveOn` property for this option?
Eg. `exclusiveOn: ["--new-management-key"]`

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+              args: [
+                {
+                  name: "SLOT",
+                  description: "PIV slot of the certificate",
+                },
+                {
+                  name: "PUBLIC-KEY",
+                  description:
+                    "File containing a public key. Use '-' to use stdin",
+                },
```
SeparateRecords, line 1943: Same thing for the arg below this one, can have a template and suggestions.
```suggestion
                {
                  name: "PUBLIC-KEY",
                  description:
                    "File containing a public key. Use '-' to use stdin",
                  template: "filepaths",
                  suggestions: [{ name: "-", hidden: true }],
                },
```

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+              args: [
+                {
+                  name: "SLOT",
+                  description: "PIV slot of the private key",
+                },
+                {
+                  name: "CERTIFICATE",
+                  description:
+                    "File to write attestation certificate to. Use '-' to use stdout",
+                },
```
SeparateRecords, line 2005: Same as above! And for any of the other args -- you can probably search for "stdout"

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+        name: "DEBUG|INFO|WARNING|ERROR|CRITICAL",
+      },
+    },
+    {
+      name: "--log-file",
+      description:
+        "Write logs to the given FILE instead of standard error; ignore unless --log-level is also set",
+      args: {
+        name: "FILE",
+      },
```
SeparateRecords, line 2334:
```suggestion
      args: {
        name: "FILE",
        template: "filepaths",
      },
```

src/ykman.ts
```diff
@@ -0,0 +1,2354 @@
...
+        name: "SERIAL",
+      },
+    },
+    {
+      name: ["-r", "--reader"],
+      description:
+        "Use an exteral smart card reader. Conflicts with --device and list",
+      args: {
+        name: "NAME",
+      },
```
SeparateRecords, line 2319: Can we know the available readers at runtime? Is there a command you can use to list them all out?

SeparateRecords (changes requested):
Hey @vaaleyard, looks good! There's a few superficial / polish improvements to make but I'd even be happy to merge this and add those in the future if you like. Shouldn't be too hard to add them, so it's up to you!

src/ykman.ts
vaaleyard: I get this error message when copying this code:
``` 
Type 'Promise<string[]>' is not assignable to type 'Promise<Suggestion[]>'.  Type 'string[]' is not assignable to type 'Suggestion[]'.    Type 'string' has no properties in common with type 'Suggestion'. typescript (2322) [55, 29]
```
No idea why, could you help me?

src/ykman.ts
vaaleyard: Resolved in 9498b07611edacca15c095724da90551af8799e4

src/ykman.ts
vaaleyard: Fixed in 9498b07611edacca15c095724da90551af8799e4

src/ykman.ts
vaaleyard: Fixed in 9498b07611edacca15c095724da90551af8799e4

src/ykman.ts
vaaleyard: Fixed in 9498b07611edacca15c095724da90551af8799e4

src/ykman.ts
vaaleyard: Fixxed in c4e66847af0377a0f9a6a91105ce8a464c263274

src/ykman.ts
vaaleyard: Fixed in ce1664646dac9df1dd1d72f8c0130935bfe66b38

src/ykman.ts
vaaleyard: Fixed in ce1664646dac9df1dd1d72f8c0130935bfe66b38

src/ykman.ts
vaaleyard: Fixed in ce1664646dac9df1dd1d72f8c0130935bfe66b38

src/ykman.ts
vaaleyard: Fixed in ce1664646dac9df1dd1d72f8c0130935bfe66b38

vaaleyard: Actually, there is, I've made some regex to get it and made another generator in the above option.
Here: 8b98bc8513b3798489767be723f5b53ae8bf6900.

It's suggesting the right stuff.

src/ykman.ts
SeparateRecords: Ah, sorry - totally forgot that in a generator it has to be a `Suggestion` object. You can make it `{ name: "OTP" }`, etc.

SeparateRecords:
Fantastic, will be ready to merge once that simple change is made - I get a lot of notifications on github so please do bump this if I miss it :)

src/ykman.ts
vaaleyard: I'm sorry, but I didn't understand at all. What has to be a `Suggestion` object? I'm trying to find another spec that use some generator to see if I can get it.

SeparateRecords: Ah, my bad, should've been clearer.

The `Suggestion` type is what everything inherits from. It has all the `name`, `description`, `displayName`, etc... properties. So instead of returning strings, you can return Suggestions.

I don't think there's a particular reason that suggestions _couldn't_ return strings, will add this to my to-dos for the month :)

vaaleyard: I've commited some code, could you see if it's right? My mind is telling that is definitely wrong...

Do I have to create a generator like [this](https://github.com/withfig/autocomplete/blob/master/src/gpg.ts#L1), for example?

fedeci: @vaaleyard here we want a generator that triggers on each + but also that removes previously inserted tokens each time the suggestions are shown. Currently we do not provide any sugar to do this so the generator has to be written manually.

```
const generator: {
  trigger: "+",
  custom: async (tokens, executeShellCommand) => {
    const previouslyAdded = tokens[tokens.length - 1].split("+")
    // this has to be an alias map of the possible suggestions
    const possibleSuggestionsMap = new Map([
      ["OTP", ["o"]],
      ["o", ["OTP"]],
      // ...all other values
    ])
    const suggestions = new Set(possibleSuggestionsMap.values())
    
    for (const token of previouslyAdded) {
      suggestions.delete(token)
      const aliases = possibleSuggestionsMap.get(token)
      if (aliases) {
        aliases.forEach(v => { suggestions.delete(v) })
      }
    }
    return [...suggestions].map(v => ({ name: v }))
  },
}
```

Also consider this: image

I will add a new property as a sugar for this problem this week!

vaaleyard: I guess instead of
```js
      if (aliases) {
        aliases.forEach(v => { suggestions.delete(v) })
      }
```

should be
```js
      if (aliases) {
        suggestions.delete(aliases)
      }
```
?

because it only works with []string.

fedeci: That's correct @vaaleyard

fedeci: `Set.prototype.delete` accepts a single primitive as argument, not an array🤔

vaaleyard: @fedeci there's no more errors now 😅. Could you check if it's all ok? thanks

src/ykman.ts
```diff
@@ -0,0 +1,2503 @@
...
+      ["c", ["CCID"]],
+    ]);
+    const suggestions = new Set(possibleSuggestionsMap.values());
+
+    for (const token of previouslyAdded) {
+      suggestions.delete(token);
+      const aliases = possibleSuggestionsMap.get(token);
+      if (aliases) {
+        aliases.forEach((v) => {
+          suggestions.delete([v]);
```
fedeci, line 21:
```suggestion
          suggestions.delete(v);
```

src/ykman.ts
```diff
@@ -0,0 +1,2503 @@
...
+    // this has to be an alias map of the possible suggestions
+    const possibleSuggestionsMap = new Map([
+      ["OTP", ["o"]],
+      ["o", ["OTP"]],
+      ["FIDO", ["f"]],
+      ["f", ["FIDO"]],
+      ["CCID", ["c"]],
+      ["c", ["CCID"]],
+    ]);
+    const suggestions = new Set(possibleSuggestionsMap.values());
```
fedeci, line 14:
```suggestion
  const suggestions = new Set([...possibleSuggestionsMap.values()].flat());
```

fedeci (changes requested):

src/ykman.ts
vaaleyard: committed

src/ykman.ts
vaaleyard: Committed

vaaleyard:
I guess it's all okay now?

fedeci (approved):
Thanks for the big spec @vaaleyard!

PR #1366, opened 2022-07-27
prisma: update and add new commands

Jolg42:
Update to `prisma` CLI commands

**Additional info:**

I work at Prisma and this is our official docs https://www.prisma.io/docs/reference/api-reference/command-reference

### DIFF
--- a/src/prisma.ts
+++ b/src/prisma.ts
@@ -37,7 +37,14 @@ const completionSpec: Fig.Spec = {
           description: "Define the datasource provider to use",
           args: {
             description: "Choose provider",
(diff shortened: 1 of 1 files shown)


SeparateRecords:
@Jolg42 looks good to me! Would be great to set up an integration so the schema can be generated automatically - what framework (if any) are you using for the prisma CLI?

SeparateRecords (approved):

Jolg42:
Would be great indeed! @SeparateRecords do you have some examples in mind? The CLI does not use a framework for now.

PR #1367, opened 2022-07-27
fix: Add workspace to `fig run` spec

chaynabors:


chaynabors:
@mschrage

fedeci:
If this was added in a minor version it have to be added as a new minor version.

grant0417:
The aliases should have always been there, unsure why they were not there before

PR #1373, opened 2022-07-29
feat: add `mamba` spec

Lunchb0ne:
[Mamba](https://github.com/mamba-org/mamba) is a faster alternative for conda written in C++.

New Spec for `mamba` 

Most of it is taken from the spec for `conda`

**Additional info:**

### DIFF
--- a/src/mamba.ts
+++ b/src/mamba.ts
@@ -138,6 +138,7 @@ const update_options: Fig.Option[] = [
     description:
       "Read package versions from the given file. Repeated file specifications can be passed (e.g. --file=file1 --file=file2)",
     args: {
+      name: "file",
       template: "filepaths",
     },
   },
@@ -146,7 +147,7 @@ const update_options: Fig.Option[] = [
     name: ["-c", "--channel"],
     description:
       "They are given (including local directories using the 'file://' syntax or simply a path like '/home/conda/mychan' or '../mychan'). Then, the defaults or channels from .condarc are searched (unless --override-channels is given). You can use 'defaults' to get the default packages for conda. You can also use any name and the .condarc channel_alias value will be prepended. The default channel_alias is http://conda.anaconda.org/",
-    args: {},
+    args: { name: "channel" },
   },
   {
     name: "--use-local",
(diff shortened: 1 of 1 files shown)


src/mamba.ts
```diff
@@ -0,0 +1,1120 @@
...
+          name: parts[0],
+          description: parts[parts.length - 1],
+          priority: isActive ? 100 : 50,
+          icon: isActive ? "✅" : "🐍",
+        });
+      }
+      return availableEnvs;
+    } catch (e) {
+      return availableEnvs;
+    }
```
fedeci, line 40: Is the trycatch really required?

src/mamba.ts
```diff
@@ -0,0 +1,1120 @@
...
+};
+
+const getInstalledPackages: Fig.Generator = {
+  script: "conda list --json",
+  scriptTimeout: 10000,
+  cache: {
+    strategy: "stale-while-revalidate",
+  },
+  postProcess: function (out) {
+    const parsed: Array<Package> = JSON.parse(out);
```
fedeci, line 51: This should be trycatched

src/mamba.ts
```diff
@@ -0,0 +1,1120 @@
...
+        {
+          name: "--stack",
+          description:
+            "Stack the environment being activated on top of the previous active environment",
+        },
+        {
+          name: "--no-stack",
+          description:
+            "Do not stack the environment. Overrides 'auto_stack' setting",
+        },
```
fedeci, line 373: These are mutually `exclusiveOn` (see https://fig.io/docs/reference/option#exclusiveon)

fedeci (dismissed):
Great one, thanks! I left two small comments to solve and we are ready to merge this.

src/mamba.ts
Lunchb0ne: Ah, it's not necessary now. Updated!

src/mamba.ts
Lunchb0ne: Done!

src/mamba.ts
Lunchb0ne: I updated some options with `exclusiveOn` and `dependsOn`.

fedeci: Great, thanks!

SeparateRecords (changes requested):
Spec is looking really good! One request - can you add a `name` prop everywhere you see `args: {}`? I know this is just inherited from conda but don't worry about fixing it there

Looks better in https://fig.io/manual because it describes what the argument is for :) no need to add a description to them

SeparateRecords (approved):
lgtm! Thank you so much for the contribution, nice job on it!

PR #1380, opened 2022-08-03
feat: add pipx

meowmeowmeowcat:

Add spec for [pipx](https://github.com/pypa/pipx)

### DIFF
--- a/src/pipx.ts
+++ b/src/pipx.ts
@@ -1,3 +1,15 @@
+const packagesGenerator: Fig.Generator = {
+  script: "pipx list --short",
+  postProcess: (out) => {
+    return out.split("\n").map((line) => {
+      return {
+        name: line.split(" ").at(0),
+        description: "Installed package",
+      };
+    });
+  },
+};
+
 const completionSpec: Fig.Spec = {
   name: "pipx",
   description: "Install and Run Python Applications in Isolated Environments",
@@ -117,6 +129,7 @@ const completionSpec: Fig.Spec = {
           name: "package",
           description:
             "Name of the existing pipx-managed Virtual Environment to inject into",
+          generators: packagesGenerator,
         },
         {
           name: "dependencies",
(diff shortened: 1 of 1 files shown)


src/pipx.ts
```diff
@@ -0,0 +1,444 @@
...
+        },
+        {
+          name: "--verbose",
+          description: "Show verbose output",
+        },
+      ],
+      args: {
+        name: "package_spec",
+        description: "Package name or pip installation spec",
+      },
```
SeparateRecords, line 66: You can import the suggestions from `./pip` and add them to this arg!

src/pipx.ts
```diff
@@ -0,0 +1,444 @@
...
+          name: "--verbose",
+          description: "Show verbose output",
+        },
+      ],
+      args: [
+        {
+          name: "package",
+          description:
+            "Name of the existing pipx-managed Virtual Environment to inject into",
+        },
```
SeparateRecords, line 120: Same thing as above 🙂

src/pipx.ts
```diff
@@ -0,0 +1,444 @@
...
+          },
+        },
+        {
+          name: "--verbose",
+          description: "Show verbose output",
+        },
+      ],
+      args: {
+        name: "package",
+      },
```
SeparateRecords, line 174: For this and other args where you're specifying a package already installed, you can use a `generator` (https://fig.io/docs/reference/generator) and `pipx list` I think

SeparateRecords (changes requested):
Looking good! Not much to change here. Sorry for the delay too 😅

src/pipx.ts
meowmeowmeowcat: pipx is used to install Python console applications, so not all Python packages can be installed by pipx.

SeparateRecords: That's fair. Will leave this up to your judgement in that case!

PR #1381, opened 2022-08-03
feat: add `glow` completions

SeparateRecords:
1. I wanted to add the 'wrapwidth' option to neovim. The suggestion has been open for _**six years**_ nobody has added the option to soft wrap at a set column number.
2. To do that, I had to read the documentation on how to build it.
3. To read the documentation, I wanted to use `glow`.
4. In the documentation was a link that was wrapping over two lines, so I couldn't click it.
5. I wanted to see if there was a flag for glow to use the proper escape codes to make the link clickable.
6. There were no fig completions for glow.

So here we are... I _was_ going to add an option to neovim...

### DIFF
--- a/src/glow.ts
+++ b/src/glow.ts
@@ -5,6 +5,7 @@ const completionSpec: Fig.Spec = {
   description: "Render markdown on the CLI, with pizzazz!",
   args: {
     name: "source|dir",
+    isOptional: true,
     generators: filepaths({
       extensions: ["md"],
       suggestFolders: "always",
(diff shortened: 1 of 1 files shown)


src/glow.ts
```diff
@@ -0,0 +1,94 @@
...
+  description: "Render markdown on the CLI, with pizzazz!",
+  args: {
+    name: "source|dir",
+    generators: filepaths({
+      extensions: ["md"],
+      suggestFolders: "always",
+      editFileSuggestions: { priority: 52 },
+      editFolderSuggestions: { priority: 51 },
+    }),
+  },
```
SeparateRecords, line 14: This arg is optional

PR #1383, opened 2022-08-05
feat: add `streamlit` spec

jrieke:

Feature

No autocompletion for [Streamlit](https://streamlit.io/) 😢

Added autocompletion for Streamlit 🎈, with all commands and options mentioned in the help message: 

CleanShot 2022-08-05 at 16 53 51@2x

**Additional info:**

1. Great tool!! Just discovered 2h ago and using fig + creating a spec was super smooth.
2. I'm a PM at Streamlit, so happy to maintain this if we update something on our side. 
3. If you want to know more about Streamlit, feel free to reach out 😉

### DIFF
--- a/src/streamlit.ts
+++ b/src/streamlit.ts
@@ -2,47 +2,56 @@ import { filepaths } from "@fig/autocomplete-generators";
 
 const completionSpec: Fig.Spec = {
   name: "streamlit",
-  description: "",
+  description: "Streamlit",
   subcommands: [
     {
       name: "activate",
       description: "Activate Streamlit by entering your email",
+      icon: "✉️",
     },
     {
       name: "cache",
       description: "Manage the Streamlit cache",
+      icon: "🗂",
       subcommands: [
         {
           name: "clear",
           description: "Clear st.cache, st.memo, and st.singleton caches",
+          icon: "🗑",
         },
       ],
     },
     {
       name: "config",
       description: "Manage Streamlit's config settings",
+      icon: "⚙️",
       subcommands: [
         {
           name: "show",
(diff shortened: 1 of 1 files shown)


src/streamlit.ts
```diff
@@ -0,0 +1,74 @@
...
+      name: "help",
+      description: "Print the help message",
+    },
+    {
+      name: "run",
+      description: "Run a Python script, piping stderr to Streamlit",
+      args: {
+        name: "file",
+        description: "The Python script to run",
+        template: "filepaths",
```
fedeci, line 47: This can use the `@fig/autocomplete-generators` package to only accept `.py` files (see https://github.com/withfig/autocomplete/blob/2a99474a1ba53fd6edacb7106f9d5bcd127d2bd8/src/python3.ts#L23)

fedeci (changes requested):
Thanks @jrieke it looks good! I would just improve the UX more by changing that generator.

src/streamlit.ts
jrieke: Oh that's fantastic, I was searching for that! Fixed it. Might be worth highlighting this a bit more in the docs, since I searched for this quite a while but couldn't find anything.

jrieke:
@fedeci Thanks for the quick review. I fixed the generator as you suggested. I also added emoji icons to all commands. Not sure if that's a "common" thing in Fig. We use a ton of emojis in Streamlit, so I'm sure our users will love it!

Let me know if you have any more comments!

fedeci (approved):
Thanks! Emojis are perfectly fine and are a quick and easy way to add custom icons to suggestions!

PR #1385, opened 2022-08-08
feat: add gibo completion spec

simonwhitaker:

Adds completions for [gibo](https://github.com/simonwhitaker/gibo)

n/a

fig supports gibo

**Additional info:**

### DIFF
--- a/src/gibo.ts
+++ b/src/gibo.ts
@@ -26,7 +26,14 @@ const completionSpec: Fig.Spec = {
         isVariadic: true,
       },
     },
-    { name: "search", description: "Search for boilerplates" },
+    {
+      name: "search",
+      description: "Search for boilerplates",
+      args: {
+        name: "search term",
+        description: "Name of the language or framework to search for",
+      },
+    },
     {
       name: "root",
       description: "Show the directory where gibo stores its boilerplates",
--- a/src/gibo.ts
+++ b/src/gibo.ts
@@ -1,0 +1,52 @@
+const boilerplates: Fig.Generator = {
+  script: 'find "$HOME/.gitignore-boilerplates" -name "*.gitignore"',
+  postProcess: (output) => {
+    return output
+      .split("\n")
+      .map((path) => {
+        return { name: path.replace(/.*\//, "").replace(".gitignore", "") };
+      })
+      .sort((a, b) => {
+        return a.name > b.name ? 1 : a.name < b.name ? -1 : 0;
+      });
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "gibo",
+  description: "Easy access to gitignore boilerplates",
+  subcommands: [
+    {
+      name: "dump",
+      description: "Dump one or more boilerplates to STDOUT",
+      args: {
+        name: "boilerplate",
+        description: "Name of the boilerplate to dump",
+        generators: boilerplates,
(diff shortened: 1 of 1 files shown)


src/gibo.ts
```diff
@@ -0,0 +1,52 @@
...
+  script: 'find "$HOME/.gitignore-boilerplates" -name "*.gitignore"',
+  postProcess: (output) => {
+    return output
+      .split("\n")
+      .map((path) => {
+        return { name: path.replace(/.*\//, "").replace(".gitignore", "") };
+      })
+      .sort((a, b) => {
+        return a.name > b.name ? 1 : a.name < b.name ? -1 : 0;
+      });
```
fedeci, line 11: nitpick:
```suggestion
      .sort();
```

src/gibo.ts
```diff
@@ -0,0 +1,52 @@
+const boilerplates: Fig.Generator = {
+  script: 'find "$HOME/.gitignore-boilerplates" -name "*.gitignore"',
+  postProcess: (output) => {
+    return output
+      .split("\n")
```
fedeci, line 5:
```suggestion
      .trim()
      .split("\n")
```

fedeci (changes requested):
I didn't know about this tool but I am installing it immediately! Can you just address the comments I left please?

src/gibo.ts
SeparateRecords: @fedeci why sort the objects themselves? the name property is the only thing we care about sorting by

There's also [`String.prototype.localCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#sort_an_array) which is what I'd have used in this case

src/gibo.ts
```diff
@@ -0,0 +1,59 @@
+const boilerplates: Fig.Generator = {
+  script: 'find "$HOME/.gitignore-boilerplates" -name "*.gitignore"',
```
SeparateRecords, line 2: Is there a reason you're using this instead of `gibo list`? The output is piped into `cat` automatically so it'll be newline-separated, seems like less work to process.

```ts
script: "gibo list",
splitOn: "\n",
```

src/gibo.ts
fedeci: Oops I fucked up thinking those were strings.

src/gibo.ts
simonwhitaker: Done!

src/gibo.ts
simonwhitaker: Ah, I forgot `localCompare` -- haven't written JS in a while. Thank you!

src/gibo.ts
simonwhitaker: Oooh, nice! Let me try that.

src/gibo.ts
SeparateRecords: @simonwhitaker If it weren't for helping maintain this repo I don't think I'd be sorting strings in JS very often either 😄

SeparateRecords:
Looks great! @simonwhitaker thank you so much for contributing this, and also for bringing attention to gibo because that looks _so nice_ for setting up projects 🤩

SeparateRecords (approved):

PR #1386, opened 2022-08-08
feat: add cf

howardwicks1:
Adding spec for `cf` cloudfoundry-cli

**Additional info:**

### DIFF
--- a/src/cf.ts
+++ b/src/cf.ts
@@ -1,7 +1,8 @@
 const completionSpec: Fig.Spec = {
   name: "cf",
   description: "",
-  subcommands: [{
+  subcommands: [
+  {
     name: "app",
     description: "Display health and status for app",
     args: {
@@ -77,6 +78,43 @@ const completionSpec: Fig.Spec = {
   {
     name: "login",
     description: "Log user in",
+    options: [
+      {
+        name: "-a",
+        description: "API endpoint (e.g. https://api.example.com)",
+        args: {
+          name: "API endpoint"
+        }
+      },
+      {
+        name: "-o",
+        description: "Org",
+        args: {
+          name: "Org"
+        }
+      },
+      {
+        name: "-s",
+        description: "Space",
+        args: {
+          name: "Space"
+        }
+      },
+      {
+        name: "-p",
+        description: "Password",
+        args: {
+          name: "Password"
+        }
+      },
+      {
+        name: "-u",
+        description: "Username",
+        args: {
+          name: "Username"
+        }
+      },
+    ]
   },
   {
     name: "logout",
@@ -89,6 +127,22 @@ const completionSpec: Fig.Spec = {
   {
     name: "target",
     description: "Set or view the targeted org or space",
+    options: [
+      {
+        name: "-o",
+        description: "Org",
+        args: {
+          name: "Org"
+        }
+      },
+      {
+        name: "-s",
+        description: "Space",
+        args: {
+          name: "Space",
+        }
+      },
+    ]
   },
   {
     name: "api",
@@ -105,6 +159,72 @@ const completionSpec: Fig.Spec = {
   {
     name: "push",
     description: "Push a new app or sync changes to an existing app",
+    options: [
+      {
+        name: "-b",
+        description: "Custom buildpack by name or git url",
+        args: {
+          name: "Buildpack"
+        }
+      },
+      {
+        name: "-c",
+        description: "Startup command, set to null to reset to default start command",
+        args: {
+          name: "Command",
+          isOptional: true,
+        }
+      },
+      {
+        name: "-d",
+        description: "Domain",
+        args: {
+          name: "Domain",
+        }
+      },
+      {
+        name: ["-o", "--docker-image"],
+        description: "Docker image",
+        args: {
+          name: "Docker image",
+        }
+      },
+      {
+        name: "-f",
+        description: "Path to manifest",
+        args: {
+          name: "Path to manifest",
+        }
+      },
+      {
+        name: ["-n", "--host-name"],
+        description: "Hostname",
+        args: {
+          name: "Hostname",
+        }
+      },
+      {
+        name: "-i",
+        description: "Number of instances",
+        args: {
+          name: "Number of instances",
+        }
+      },
+      {
+        name: "-k",
+        description: "Disk limit",
+        args: {
+          name: "Disk limit",
+        }
+      },
+      {
+        name: "-m",
+        description: "Memory limit",
+        args: {
+          name: "Memory limit",
+        }
+      },
+    ]
   },
   {
     name: "scale",
--- a/src/cf.ts
+++ b/src/cf.ts
@@ -138,11 +138,369 @@ const completionSpec: Fig.Spec = {
     name: "restart-app-instance",
     description: "Terminate the running application Instance at the given index and instantiate a new instance of the application with the same index",
   },
+  {
+    name: "run-task",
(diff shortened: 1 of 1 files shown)


fedeci:
Hi @howardwicks1, thanks for the PR! Was this created manually or generated via some integration?

src/cf.ts
```diff
@@ -0,0 +1,1967 @@
...
+        },
+      },
+      {
+        name: "--allow-paid-service-plans",
+        description: "Can provision instances of paid service plans",
+      },
+      {
+        name: "--disallow-paid-service-plans",
+        description: "Cannot provision instances of paid service plans",
+      },
```
fedeci, line 1769: These appear to be mutually `exclusiveOn`

src/cf.ts
```diff
@@ -0,0 +1,1967 @@
...
```
fedeci, line 1637: These too

src/cf.ts
```diff
@@ -0,0 +1,1967 @@
...
+        name: "-f",
+        description: "Force deletion without confirmation",
+      },
+    ],
+  },
+  {
+    name: "org-users",
+    description: "Show org users by role",
+    args: {
+      name: "org",
```
fedeci, line 1387: It would be cool to have a generator for orgs, I can point you in the right direction if needed!

fedeci (changes requested):
`npm run lint:fix` should fix all the linting issues automatically, then we can either merge the PR or add a ton of generators here and there :)

src/cf.ts
howardwicks1: Good spot will sort that out

src/cf.ts
howardwicks1: Yep will sort 👍

src/cf.ts
howardwicks1: I experimented with a generator for `app name` but ripped it out in this [commit](https://github.com/withfig/autocomplete/pull/1386/commits/1b714997cad5f4d837d61af5755c7a4f08d45801) as it wasnt working as intended. Happy to have another look at creating some generators though 😄

howardwicks1:
Hello @fedeci thank you for the comments. This completion spec was created manually.

src/cf.ts
```diff
@@ -0,0 +1,1912 @@
+const completionSpec: Fig.Spec = {
+  name: "cf",
+  description: "",
```
fedeci, line 3: The description can be important for manual pages https://fig.io/manual

fedeci (approved):
Thanks for the changes! We can merge it as is for the moment since it is already a big PR, if you want you can open an issue to discuss about generators. Can you just add a description to the root command?

src/cf.ts
howardwicks1: Updated description

src/cf.ts
```diff
@@ -0,0 +1,1924 @@
...
+          script: `cf apps | cut -d " " -f1`,
+          postProcess: function (out) {
+            return out
+              .split("\n")
+              .slice(4)
+              .map((APP_NAME) => ({
+                name: APP_NAME,
+                description: "APP_NAME",
+              }));
+          },
```
SeparateRecords, line 21: Beautiful! You can pull out this generator and assign it to a variable, then you can use it anywhere there's an app name arg.

src/cf.ts
```diff
@@ -0,0 +1,1924 @@
...
+    {
+      name: "run-task",
+      description: "Run a one-off task on an app",
+      args: [
+        {
+          name: "app name",
+        },
+        {
+          name: "command",
+        },
```
SeparateRecords, line 354: Is there a way to list the tasks / commands that are valid? Can add another generator for this!

src/cf.ts
```diff
@@ -0,0 +1,1924 @@
...
+          },
+        },
+      ],
+    },
+    {
+      name: "update-service",
+      description: "Update a service instance",
+      args: {
+        name: "service",
+      },
```
SeparateRecords, line 621: Same thing with services! Can build a generator to suggest them here + every other place that takes a service

src/cf.ts
```diff
@@ -0,0 +1,1924 @@
...
+          },
+        },
+      ],
+    },
+    {
+      name: "org",
+      description: "Show org info",
+      args: {
+        name: "org",
+      },
```
SeparateRecords, line 908: Same with orgs :)

SeparateRecords (changes requested):
This spec is great! Really nice job on it, that's absolutely huge. Lot of work to do.

Left some comments for additional generators. Wasn't exhaustive, there are more places you could add them - adding them will make this spec absolutely 🔥 🤩

howardwicks1:
Have added a basic generator for APP_NAME + description

src/cf.ts
howardwicks1: Ooh sounds good, will have a look at expanding the generator to work like that. 😄

src/cf.ts
howardwicks1: Not too sure about this one, afaik the command could be anything depending on what app a user has deployed via cloudfoundry 😄 ie a user could be passing a regular bash command, reference to a script or entrypoint.

src/cf.ts
howardwicks1: Will look at this one 👍

howardwicks1:
Added generators for orgs and spaces, going to look at one for services next 😄

src/cf.ts
```diff
@@ -0,0 +1,2009 @@
...
+  postProcess: function (out) {
+    return out
+      .split("\n")
+      .slice(3)
+      .map((org) => ({
+        name: org,
+        description: "Org",
+      }));
+  },
+};
```
SeparateRecords, line 38: Looks like the output for all of these is fairly consistent since it's basically the same generator - here's a quick simplification!

```suggestion
const postProcessCfList =
  (description: string, leadingLines: number) => (out: string) =>
    out
      .trim() // output can have a trailing newline, prevents empty suggestion
      .split("\n")
      .slice(leadingLines)
      .map((name) => ({ name, description }));

const generateAppNames: Fig.Generator = {
  script: `cf apps | cut -d " " -f1`,
  postProcess: postProcessCfList("App name", 4),
};

const generateOrgs: Fig.Generator = {
  script: `cf orgs`,
  postProcess: postProcessCfList("Org", 3),
};

const generateSpaces: Fig.Generator = {
  script: `cf spaces`,
  postProcess: postProcessCfList("Space", 3),
};
```

SeparateRecords:
So pumped for the spec to land. If you want we can merge now, then do the remaining generators later - what you've got already is _absolutely fantastic_.

src/cf.ts
howardwicks1: Ah this is really nice! Will give this a go 👍

howardwicks1:
> So pumped for the spec to land. If you want we can merge now, then do the remaining generators later - what you've got already is _absolutely fantastic_.

I will address the last couple comments first 😄 but then yes I would be happy for the merge to go ahead

howardwicks1:
Generator for services is now included, I have not attached the services generator to some of the `user-provided` service commands as I dont think they would work with the generator (as thats only able to poll the proper list of services present on the platform

SeparateRecords (approved):
LGTM! Thank you so much for working on this! 🚀

After merge it may take a bit for it to show up for you because of caching, but it's deployed automatically and _will_ work after a couple of days at most (depending on how recently Fig cached its specs)

PR #1389, opened 2022-08-11
feat(yarn): bump priority of scripts

SeparateRecords:
This does not affect `npm` since the scripts are only visible with `npm run`, but it improves the usability of `yarn` because the scripts are at the top


fedeci (approved):

SeparateRecords:
lightning fast @fedeci strikes again 😄

PR #1410, opened 2022-08-15
feat(northflank): add northflank autocompletion spec

MSuter6:
The PR introduces the complete Fig specification for the Northflank CLI.

**Additional info:**
Northflank is a comprehensive developer platform to build and scale microservices, jobs and managed databases with a powerful UI, API & CLI.
Northflank website: https://northflank.com/
CLI reference: https://www.npmjs.com/package/@northflank/cli

This specification was autogenerated by the `@fig/complete-commander` package.


SeparateRecords (approved):
Looks good to me! Would like @fedeci to take a look too, but all looks fine since this is auto generated.

There's a couple places where there could be `suggestions: [ ... ]` to make some specific suggestions (eg. `--output`) but afaik that's defined in the commander program, so we could probably deal with that upstream and regenerate the spec when the improvements are merged :)

MSuter6:
Hi @SeparateRecords, thank you for having a look and your feedback. 
Good point about `suggestions`. Commander supports `choices` (https://github.com/tj/commander.js/#more-configuration) which also seem to be supported by the fig autogenerator. We don't use `choices` in commander yet but that's something we should look into.
I was also thinking about adding a postprocessing step to add async generators after the autogeneration has run. This would allow to have dynamic `suggestions` which, for example load a list of project ID's to select from. 
As this will take a bit more time, it's something we would like to explore and add in a future PR.

fedeci:
Hi @MSuter6, I'm giving a look rn to the spec. If you could setup https://fig.io/docs/guides/integrating/push-to-autocomplete-action  to automatically push updates here on a new release it would be super cool since we have a merge tool running in that action allowing to retain generators without you having to create a post generation step.

fedeci (approved):

fedeci:
Thanks for the PR!

SeparateRecords:
> we have a merge tool running in that action allowing to retain generators without you having to create a post generation step

It's true! @fedeci worked some _magic_ to make it work, I still can't believe how effective it is. Definitely worth looking into since it will probably save you a lot of time.

MSuter6:
Hi @fedeci, thanks a lot for the tip. Generating and releasing the fig spec automatically alongside our current CLI release process is something we still need to do so we'll definitely have a look there. 👍🏻

PR #1411, opened 2022-08-16
feat(cargo): load suggestions when toolchain is specified

SeparateRecords:


grant0417:
Not sure if you want to generate the list at runtime, you can do that with `rustup toolchain list` and split on the first `-` and take the first part.

SeparateRecords:
@grant0417 done!

grant0417 (approved):

PR #1425, opened 2022-08-18
feat(exa): --no-icons

ben5vik:


src/exa.ts
```diff
@@ -252,6 +252,10 @@ const completionSpec: Fig.Spec = {
       description: "Shows icons before each folder or file",
       name: "--icons",
     },
+    {
+      description: "Does not display icons (This overwrites -icons)",
+      name: "--no-icons", 
+    }
```
fedeci, line 258: These are mutually `exclusiveOn`, can you add that property please?

fedeci (changes requested):

src/exa.ts
```diff
@@ -252,6 +252,10 @@ const completionSpec: Fig.Spec = {
       description: "Shows icons before each folder or file",
       name: "--icons",
```
fedeci, line 253:
```suggestion
      name: "--icons",
      exclusiveOn: ["--no-icons"],
```

src/exa.ts
```diff
@@ -252,6 +252,10 @@ const completionSpec: Fig.Spec = {
...
     },
+    {
+      description: "Does not display icons (This overwrites -icons)",
```
fedeci, line 256:
```suggestion
      description: "Does not display icons",
      exclusiveOn: ["--icons"],
```

fedeci (changes requested):
Sorry I should have pointed you in the right direction, this is the best way to do it since we can then only suggest an option when it is really relevant!

ben5vik:
Ya, I also noticed the `--no-icons` is not that important because u could just not use `--icons` in order to also have `--no-icons`

fedeci (approved):
Thanks! It looks good to me!

src/exa.ts
```diff
@@ -251,6 +251,12 @@ const completionSpec: Fig.Spec = {
     {
       description: "Shows icons before each folder or file",
       name: "--icons",
+      exclusiveOn: ["--no-icons"],
+    },
+    {
+      description: "Does not display icons",
+      exclusiveOn: ["--icons"],
+      name: "--no-icons",
```
fedeci, line 259:
```suggestion
      name: "--no-icons",
```

PR #1436, opened 2022-08-20
feat: add `coda` spec

arsm100:
Feature: Add [Coda SDK](https://coda.io/packs/build/latest/guides/development/cli/#create-pack-definition) autocomplete spec.

No Coda SDK spec

New Coda SDK spec added

**Additional info:**
[Coda](https://coda.io/about) is a new kind of doc that brings words, data, and teams together. It comes with a set of building blocks that anyone can combine to create a doc as powerful as an app.
With a growing dev community, Coda SDK is becoming more popular than ever. This new spec will make Fig an easy choice for Coda Pack developers!

### DIFF
--- a/src/coda.ts
+++ b/src/coda.ts
@@ -1,0 +1,196 @@
+import { filepaths } from "@fig/autocomplete-generators";
+
+const formulaNames: Fig.Generator = {
+  script: `grep -A5 --include=\*.ts --exclude-dir=node_modules -r 'addFormula\\|addSyncTable\\|makeFormula\\|makeSyncTable' . | grep -A3 -i formula | grep name: | grep -oE "['\\"]\\w*['\\"]"`,
+  postProcess: (output) => {
+    if (output.trim().length === 0) {
+      return [];
+    }
+    return output.split("\n").map((formulaName) => {
+      return {
+        name: formulaName.replaceAll(/['"]/g, "").trim(),
+        description: `Execute ${formulaName}`,
+      };
+    });
+  },
+};
+
+const pathToPackArg: Fig.Arg = {
+  name: "path/to/pack.ts",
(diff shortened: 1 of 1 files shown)


src/coda.ts
```diff
@@ -0,0 +1,204 @@
...
+    name: "--help",
+    description: "Show help",
+  },
+];
+
+const pathToPackArg: Fig.Arg = {
+  name: "path/to/pack.ts",
+  description: "The path to the pack.ts file. E.g. src/pack.ts",
+  template: "filepaths",
+};
```
fedeci, line 31: This can use the `filepaths` function from the `@fig/autocomplete-generators` package to filter over .ts file extensions

src/coda.ts
```diff
@@ -0,0 +1,204 @@
...
+        description: `Execute ${formulaName}`,
+      };
+    });
+  },
+};
+
+const commonOptions: Fig.Option[] = [
+  {
+    name: "--version",
+    description: "Show version number",
```
fedeci, line 19: This allow the option to be in all the subcommands too without having to add it multiple times
```suggestion
    description: "Show version number",
    isPersistent: true,
```

fedeci: In this way you can avoid having `commonOptions` and we can just add it once to the root command.

fedeci (changes requested):
Thanks for this first spec! This spec looks great, there are just a couple nitpicks to solve before landing it!

src/coda.ts
arsm100: Addressed.

src/coda.ts
arsm100: Thanks for the suggestion. The file is much shorter and cleaner this way! Addressed.

fedeci (approved):
Thanks!

PR #1437, opened 2022-08-20
feat(brew): add `--prefix` subcommand

bengiv:
Adding the --prefix subcommand

### DIFF
--- a/src/brew.ts
+++ b/src/brew.ts
@@ -1204,6 +1204,15 @@ const completionSpec: Fig.Spec = {
       },
     },
     {
+      name: "--prefix",
+      description: "Prefix of <formula>",
+      args: {
+        isVariadic: true,
+        name: "formula",
+        generators: formulaeGenerator,
+      },
+    },
+    {
       name: "cask",
       description:
         "Homebrew Cask provides a friendly CLI workflow for the administration of macOS applications distributed as binaries",


src/brew.ts
```diff
@@ -1203,6 +1203,15 @@ const completionSpec: Fig.Spec = {
...
     },
+    {
+      name: "--prefix",
+      description: "Prefix of <formula>",
+      args: {
+        isVariadic: true,
+        name: "formula",
+        generators: formulaeGenerator,
+      },
+    },
```
mschrage, line 1214: This looks like it should be an option rather than a subcommand?

ben5vik: Hmm

ben5vik: Well we know it wasn't in autocompletion and this adds it 😅

SeparateRecords: It looks like an option but @ben5vik is right, it is actually a subcommand 🤨 even has its own non-global options!

@ben5vik: Can you add `type: "option"` and `icon: "fig://icon?type=option` to this, just to make it look more like an option in the list?

SeparateRecords: Ah, and also a comment explaining that _yes_ it looks like a subcommand but _no_ it actually isn't one

SeparateRecords, line 1214:
```suggestion
    {
      // This looks like an option, but it's actually a subcommand with its own options
      // See: brew help --prefix
      name: "--prefix",
      description: "Prefix of <formula>",
      icon: "fig://icon?type=option",
      args: {
        isVariadic: true,
        name: "formula",
        generators: formulaeGenerator,
      },
      options: [
        {
          name: "--unbrewed",
          description: "List files in Homebrew's prefix not installed by Homebrew",
        },
        {
          name: "--installed",
          description:
            "Outputs nothing and returns a failing status code if formula is not installed"
        },
      ],
    },
```

fedeci:
recheck

src/brew.ts
fedeci, line 1214:
```suggestion
    {
      // NOTE: this is actually a command even if it has the double dash in the front
      name: "--prefix",
      description: "Prefix of <formula>",
      type: "option",
      args: {
        isVariadic: true,
        name: "formula",
        generators: formulaeGenerator,
        },
      options: [
        {
          name: "--unbrewed",
          description: "List files in Homebrew's prefix not installed by Homebrew",
        },
        {
          name: "--installed",
          description:
            "Outputs nothing and returns a failing status code if formula is not installed",
        },
      ],
    },
```

PR #1440, opened 2022-08-21
feat: add lsblk spec

SebDanielsson:
Feature

Autocompletion for lsblk is currently not available. Closes https://github.com/withfig/autocomplete/issues/1405 

Adds autocompletion for lsblk

**Additional info:**
This is my first contribution so feel free to point out any errors☺️

### DIFF
--- a/src/lsblk.ts
+++ b/src/lsblk.ts
@@ -85,6 +85,7 @@ const completionSpec: Fig.Spec = {
       description: "Specify which output columns to print",
       args: {
         name: "list",
+        isVariadic: true,
       },
     },
     {
--- a/src/lsblk.ts
+++ b/src/lsblk.ts
@@ -1,0 +1,151 @@
+const completionSpec: Fig.Spec = {
+  name: "lsblk",
+  description: "List block devices",
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show help for lsblk",
+    },
+    {
+      name: ["--version", "-V"],
+      description: "Show version for lsblk",
+    },
+    {
+      name: ["--all", "-a"],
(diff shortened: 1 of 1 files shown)


src/lsblk.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+    {
+      name: ["--noheadings", "-n"],
+      description: "Do not print a header line",
+    },
+    {
+      name: ["--output", "-o"],
+      description: "Specify which output columns to print",
+      args: {
+        name: "list",
```
fedeci, line 87:
```suggestion
        name: "list",
        isVariadic: true,
```

src/lsblk.ts
```diff
@@ -0,0 +1,151 @@
...
+      name: ["--scsi", "-S"],
+      description: "Output info about SCSI devices only",
+    },
+    {
+      name: ["--inverse", "-s"],
+      description: "Print dependencies in inverse order",
+    },
+    {
+      name: ["--tree", "-T"],
+      description: "Force tree-like output format",
```
fedeci, line 112: Apparently this requires an argument
```suggestion
      description: "Force tree-like output format",
      args: {
        name: "column",
      },
```

fedeci (changes requested):
Thanks @SebDanielsson, this spec looks great! I just left two comments to be solved before landing this!

src/lsblk.ts
SebDanielsson: Yeah I wasn't sure about that one because there's no space between the option and the argument.

SeparateRecords (approved):
LGTM! Thank you @SebDanielsson 😊

PR #1444, opened 2022-08-23
feat: add `firefox` spec

mrhappyma:
I was messing around a bit and accidentally discovered that there's a Firefox CLI. Huh. Here's a spec for it.

Only a draft because of one thing: how should I do options for the `--start-debugger-server`? The options are `[ <port> | <path> ]`, but I don't see any way to use something like `exclusiveOn` for arguments. This has got to have come up before, right?

### DIFF
--- a/src/firefox.ts
+++ b/src/firefox.ts
@@ -1,0 +1,252 @@
+const completionSpec: Fig.Spec = {
+  name: "firefox",
+  description: "Free open-source web browser developer by Mozilla",
+  args: {
+    name: "URL",
+    description: "URL to open",
+  },
+  options: [
+    {
+      name: "--display",
+      description: "Specify an X display to use",
+      args: {
+        name: "display",
+        description: "X display to use",
+      },
+    },
+    {
+      name: "--sync",
+      description: "Make X calls synchronous",
+    },
+    {
+      name: "--g-fatal-warnings",
+      description: "Make all warnings fatal",
+    },
+    {
+      name: ["-h", "--help"],
+      description: "Print help message and exit",
+    },
+    {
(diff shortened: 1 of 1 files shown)


SeparateRecords:
> The options are [ <port> | <path> ], but I don't see any way to use something like exclusiveOn for arguments

When suggesting two mutually exclusive arguments you can give suggestions for both! No need for exclusiveOn there.

```ts
args: {
  name: "port or path",
  template: "filepaths",
  generators: {
    // Needs to trigger on every keystroke
    trigger: () => true,

    // Suggest current token if it's a port number
    custom: (tokens) => {
      const finalToken = tokens[tokens.length-1];
      const port = Number(finalToken);
      if (Number.isNaN(port)) return [];
      if (!Number.isInteger(port)) return [];
      if (port < 0 || port > 65535) return [];
      return [{ name: finalToken, description: "Port number" }];
    },
  },
},

```

mschrage (approved):
Looks good to me!

mschrage:
@SeparateRecords love this port generator! We could add this to [@fig/autocomplete-generators](https://github.com/withfig/autocomplete-tools/tree/main/generators/src) package?

SeparateRecords:
> We could add this to [@fig/autocomplete-generators](https://github.com/withfig/autocomplete-tools/tree/main/generators/src) package?

If there are more specs that need it, I think it would make sense to do this! Reason I wrote the generator for it because the arg is also suggesting filepaths, so if you're writing a port number we don't want that getting in the way. For most port args I don't think that's necessary, but if we want to include it as a little visual sugar I wouldn't be against it :)

PR #1445, opened 2022-08-23
feat: add `mount` spec

grant0417:

**Additional info:**

### DIFF
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -10,6 +10,7 @@ const completionSpec: Fig.Spec = {
           script: "cat /proc/partitions", // this way we don't depend on lsblk
           postProcess: (out) => {
             return out
+              .trim()
               .split("\n")
               .splice(2, out.length)
               .map((line) => "/dev/" + line.split(" ").pop())
@@ -20,9 +21,10 @@ const completionSpec: Fig.Spec = {
           },
         },
         {
-          script: "ls /dev/mapper", // usually LUKS encrypted partitions are here
+          script: "\\command ls -1 /dev/mapper", // usually LUKS encrypted partitions are here
           postProcess: (out) => {
             return out
+              .trim()
               .split("\n")
               .filter((x) => x.length != 0)
               .map((blk) => {
--- a/src/mount.ts
+++ b/src/mount.ts
@@ -1,0 +1,217 @@
+const completionSpec: Fig.Spec = {
+  name: "mount",
+  description: "Mount disks and manage subtrees",
+  args: [
+    {
+      name: "Disk/loopfile",
+      template: "filepaths",
+      generators: [
+        {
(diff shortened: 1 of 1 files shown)


grant0417:
Another spec from @Alonely0

src/mount.ts
```diff
@@ -0,0 +1,217 @@
...
+  args: [
+    {
+      name: "Disk/loopfile",
+      template: "filepaths",
+      generators: [
+        {
+          script: "cat /proc/partitions", // this way we don't depend on lsblk
+          postProcess: (out) => {
+            return out
+              .split("\n")
```
SeparateRecords, line 13:
```suggestion
            return out
              .trim()
              .split("\n")
```

src/mount.ts
```diff
@@ -0,0 +1,217 @@
...
+              .map((blk) => {
+                return { name: blk, description: "Block device" };
+              });
+          },
+        },
+        {
+          script: "ls /dev/mapper", // usually LUKS encrypted partitions are here
+          postProcess: (out) => {
+            return out
+              .split("\n")
```
SeparateRecords, line 26:
```suggestion
            return out
              .trim()
              .split("\n")
```

src/mount.ts
```diff
@@ -0,0 +1,217 @@
...
+              .splice(2, out.length)
+              .map((line) => "/dev/" + line.split(" ").pop())
+              .filter((x) => x != "/dev/")
+              .map((blk) => {
+                return { name: blk, description: "Block device" };
+              });
+          },
+        },
+        {
+          script: "ls /dev/mapper", // usually LUKS encrypted partitions are here
```
SeparateRecords, line 23: Guarantees using system ls instead of bash builtin
```suggestion
          script: "\\command ls -1 /dev/mapper", // usually LUKS encrypted partitions are here
```

SeparateRecords (approved):
LGTM! Some quick fixes but approving for when those changes are made. Nice job @Alonely0

src/mount.ts
grant0417: Unsure about this one. I'd prefer we tried to keep script blocks as vanilla as possible. At least under the hood we are not even running aliases or the rc files at all.

grant0417: Ah I see. Avoiding the bash builtin, maybe we should rethink this API.

Alonely0: is `-1` part of POSIX ls? GNU has added a lot of stuff and it's difficult to make things portable when using coreutils in scripts.

SeparateRecords: Not sure if it's posix (pretty sure it is) but `-1` is definitely on gnu ls

SeparateRecords: Yup I'm like 99% sure `-1` is a posix thing

PR #1448, opened 2022-08-23
feat(ssh): suggest hosts from `Include`d config files

### LINKED ISSUE #1187: [ssh] Autocompletition for ssh is not working with Includes
### Sanity checks

ssh

### Issue Details

When you have config file in .ssh/ its being taken in account.
But when you put the below into your config
```
Host home*
  Include ~/.ssh/home.config
```
which contains other hosts, those hosts are not being taken in account to the autocomplete.

Could you please have a look at this? 

Thanks in advance. :-)

### LINKED ISSUE #1450: [ssh] SSH does not suggest hosts valid hosts from ~/.ssh/config (if `Host` keyword is spelled in other casing)
### Sanity checks

ssh

### Issue Details

According to the [open ssh documentation](https://www.ssh.com/academy/ssh/config#listing-of-client-configuration-options) - the `~/.ssh/config` spec should allow case insensitive keywords

Currently, if the config file contains a host definition in any force other than `Host` - it is being ignored

For example:
If `~/.ssh/config` contains:

```
Host server1
    HostName www1.example.com
    User my

host server2
   hostname www2.example.com

HOST server3
  HOSTNAME www3.example.com
```
Only `server1` will be suggested by fig, although `server2` and `server3` are also valid hosts for the `ssh` command

yaronuliel:
Spec update for `ssh`

If `~/.ssh/config` file includes `Include` statement - fig doesn't provide suggestion from the included file [Bug: closes #1187]

The `ssh` spec now suggests hosts that are defined in config file included from the `~/.ssh/config` file

**Additional info:**
Solves this case (`customer-main-site` will be suggested as a host for `ssh ` command

`~/.ssh/config`:
```
Host my-personal-site
     HostName 172.0.0.1
     User siteuser

Include config.d/*
```

`~/.ssh/config.d/customer`:

```
Host customer-main-site
   HostName ...
   User blabla
```

     

This PR also resolves #1450 issue with the suggestions when `~/.ssh/config` contains lowercase `Host` configurations

### DIFF
--- a/src/ssh.ts
+++ b/src/ssh.ts
@@ -1,9 +1,5 @@
 const knownHostRegex = /(?:[a-zA-Z0-9-]+\.)+[a-zA-Z0-9]+/; // will match numerical IPs as well as domains/subdomains
 
-type ExecuteShellCommandFunction = (
-  commandToExecute: string
-) => Promise<string>;
-
 const resolveAbsolutePath = (path: string, basePath: string): string => {
   if (path.startsWith("/") || path.startsWith("~/") || path === "~") {
     return path;
@@ -14,7 +10,7 @@ const resolveAbsolutePath = (path: string, basePath: string): string => {
 
 const getConfigLines = async (
   file: string,
-  executeShellCommand: ExecuteShellCommandFunction,
+  executeShellCommand: Fig.ExecuteShellCommandFunction,
   basePath
 ) => {
   const absolutePath = resolveAbsolutePath(file, basePath);
--- a/src/ssh.ts
+++ b/src/ssh.ts
@@ -25,7 +25,10 @@ export const configHosts: Fig.Generator = {
   postProcess: function (out) {
     return out
       .split("\n")
-      .filter((line) => line.trim().startsWith("Host ") && !line.includes("*"))
+      .filter(
+        (line) =>
+          line.trim().toLowerCase().startsWith("host ") && !line.includes("*")
+      )
       .map((host) => ({
         name: host.split(" ")[1],
         description: "SSH host",
--- a/src/ssh.ts
+++ b/src/ssh.ts
@@ -1,5 +1,41 @@
 const knownHostRegex = /(?:[a-zA-Z0-9-]+\.)+[a-zA-Z0-9]+/; // will match numerical IPs as well as domains/subdomains
 
+type ExecuteShellCommandFunction = (
+  commandToExecute: string
+) => Promise<string>;
+
+const resolveAbsolutePath = (path: string, basePath: string): string => {
+  if (path.startsWith("/") || path.startsWith("~/") || path === "~") {
+    return path;
+  }
+
+  return basePath.replace(/\/$/, "") + "/" + path;
+};
+
+const getConfigLines = async (
+  file: string,
+  executeShellCommand: ExecuteShellCommandFunction,
+  basePath
+) => {
+  const absolutePath = resolveAbsolutePath(file, basePath);
+
+  const out = await executeShellCommand(`cat ${absolutePath}`);
+  const configLines = out.split("\n").map((line) => line.trim());
+
+  // Get list of includes in the config file
+  const includes = configLines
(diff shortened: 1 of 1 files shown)


yaronuliel (commented):
Resolves #1450

SeparateRecords (approved):
Looks really good, thank you so much @yaronuliel!

@fedeci could you have a look as well? If you don't see anything to change either it's an easy merge 🙂

src/ssh.ts
```diff
@@ -1,5 +1,41 @@
...
+  const configLines = out.split("\n").map((line) => line.trim());
+
+  // Get list of includes in the config file
+  const includes = configLines
+    .filter((line) => line.toLowerCase().startsWith("include "))
+    .map((line) => line.split(" ")[1]);
+
+  // Get the lines of every include file
+  const includeLines = await Promise.all(
+    includes.map((file) => getConfigLines(file, executeShellCommand, basePath))
```
fedeci, line 32: When you recurse just one directory deep it is ok to keep the original basePath, but what if another include is added and it is way more nested?
e.g.
```
# ~/.ssh/config
include foo/config-2
```

```
# ~/.ssh/foo/config-2
include bar/config-3
```

```
# ~/.ssh/foo/bar/config-3
...
```
 
While recursing, basePath will always be `~/.ssh` while file will be "foo/config-2" and "bar/config-3". In the first case the file gets correctly loaded since we resolve it to `~/.ssh/foo/config-2` but the second one is not since it is `~/.ssh/bar/config-3`.

src/ssh.ts
```diff
@@ -1,5 +1,41 @@
...
+type ExecuteShellCommandFunction = (
+  commandToExecute: string
+) => Promise<string>;
+
+const resolveAbsolutePath = (path: string, basePath: string): string => {
+  if (path.startsWith("/") || path.startsWith("~/") || path === "~") {
+    return path;
+  }
+
+  return basePath.replace(/\/$/, "") + "/" + path;
```
fedeci, line 12: Is this replacing the last slash and then adding it again?

fedeci (changes requested):
Also, this does not handle * imports, right?

src/ssh.ts
yaronuliel: @fedeci You are correct, this is the way openssh works, ALL includes are relative to the `~/.ssh` (this was counterintuitive to me as well)
See https://man.openbsd.org/ssh_config#Include

> Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file or /etc/ssh if included from the system configuration file

src/ssh.ts
```diff
@@ -1,5 +1,41 @@
...
+};
+
+const getConfigLines = async (
+  file: string,
+  executeShellCommand: ExecuteShellCommandFunction,
```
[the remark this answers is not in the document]
yaronuliel: Thanks! I didn't realized it's available on the `Fig` namespace.. my bad

Fixed

src/ssh.ts
```diff
@@ -1,5 +1,41 @@
 const knownHostRegex = /(?:[a-zA-Z0-9-]+\.)+[a-zA-Z0-9]+/; // will match numerical IPs as well as domains/subdomains
 
+type ExecuteShellCommandFunction = (
+  commandToExecute: string
+) => Promise<string>;
```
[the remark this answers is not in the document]
yaronuliel: Fixed

src/ssh.ts
yaronuliel: @fedeci Yes, I do this to support both `dir/` and `dir` as parameters for the function

src/ssh.ts
fedeci: Oops sorry!

src/ssh.ts
fedeci: nitpicky but I prefer this syntax for future maintainers
```suggestion
  return basePath + (basePath.endsWith("/") ? "" : "/") + path;
```

fedeci (approved):
Thanks!

PR #1452, opened 2022-08-24
feat: add pacman

clo4:
This is still WIP. `exclusiveOn` and `dependsOn` don't quite work for pacman. We need full "options with options".

### DIFF
--- a/src/pacman.ts
+++ b/src/pacman.ts
@@ -1,0 +1,391 @@
+const operations = [
+  {
+    name: ["-D", "--database"],
+    description: `Operate on the package database. This operation allows you to modify certain attributes of the installed packages in pacman’s database. It also allows you to check the databases for internal consistency. See Database Options below`,
+  },
+  {
+    name: ["-Q", "--query"],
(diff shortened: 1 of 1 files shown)


fedeci:
@SeparateRecords using subcommands to describe this kind of options isn't a feasible solution?

grant0417:
@fedeci does saying that is a subcommand currently work, otherwise are you able to impl this?

The cases this would have to cover are:

```
$ pacman -Ss
$ pacman -S --search
$ pacman --sync -s
$ pacman --sync --search
```

fedeci:
It works in the sense that they are correctly suggested and treated like subcommands, but for example `-Ss` wouldn't work

grant0417:
What would we need to do to support the `-Ss` case? That is the primary way `pacman` users would expect the command to work and some other arch utils work that way as well.

grant0417:
Would this be an additional flag on the subcommand or impl be default?

SeparateRecords:
Might be interesting to have a parser directive like `parseSubcommandsAsOptions` which uses the option-parsing logic for subcommands?

fedeci:
I think for the moment `additionalSuggestions` may come to our help to achieve that result easily.
@SeparateRecords `parserDirectives` would apply globally but in some cases we only want it for some options.

mschrage:
@grant0417 worthwhile keeping this open? Less relevant now, since we are not focused on Linux support.

brendanfalk:
Going to close this draft for the moment. Can always re-open. @clo4 Any interest in getting this up and running again?

PR #1455, opened 2022-08-24
fix(bun): more args, generators, etc

SeparateRecords:


src/bun.ts
```diff
@@ -92,10 +92,23 @@ const spec: Fig.Spec = {
   name: "bun",
   description:
     "A fast bundler, transpiler, JavaScript Runtime and package manager for web software",
+  args: [
+    {
+      name: "file",
+      generators: [
+        // js/jsx/mjs/cjs/mjsx/cjsx etc (and also for typescript)
```
fedeci, line 99: Are cjsx and mjsx valid extensions?
Ref https://github.com/microsoft/TypeScript/issues/44442

SeparateRecords: Yup you're totally right, although I think it's not worth the extra complexity to change the regex to `/([mc]?[tj]s|[tj]sx?)/` - imo it's harder to read and doesn't really give much benefit either

fedeci: It makes sense.

fedeci (approved):

PR #1456, opened 2022-08-25
feat(deno): update to 1.25.0

SeparateRecords:
Can't update automatically because there's a fair amount that has to be cleaned up manually in this spec, but _wow_ it is much easier to scan the Fig spec diff than diff the deno CLI's source code!! Using the generated spec helped so much.

### DIFF
--- a/src/deno.ts
+++ b/src/deno.ts
@@ -1,4 +1,3 @@
-import { filepaths } from "@fig/autocomplete-generators";
 import {
   generateDocs,
   generateInstalledDenoScripts,
--- a/src/deno.ts
+++ b/src/deno.ts
@@ -280,6 +280,7 @@ const completion: Fig.Spec = {
           name: "files",
           isVariadic: true,
           isOptional: true,
+          template: "filepaths",
         },
         {
           name: "script_arg",
--- a/src/deno.ts
+++ b/src/deno.ts
@@ -9,7 +9,7 @@ import {
 
 const completion: Fig.Spec = {
   name: "deno",
-  description: "A secure JavaScript and TypeScript runtime",
+  description: "A modern JavaScript and TypeScript runtime",
   subcommands: [
     {
       name: "bench",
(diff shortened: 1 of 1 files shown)


fedeci:
Can we use diff-versioned specs for deno?

src/deno.ts
```diff
@@ -9,7 +8,7 @@ import {
 
 const completion: Fig.Spec = {
   name: "deno",
-  description: "A secure JavaScript and TypeScript runtime",
+  description: "A modern JavaScript and TypeScript runtime",
```
fedeci, line 11: Isn't it secure anymore? 😆

fedeci (commented):
I imagine this is not yet ready to be merged, ping me when it is!

src/deno.ts
SeparateRecords: I know!! I thought the same thing when I saw their tagline had changed 😄

SeparateRecords:
> Can we use diff-versioned specs for deno?

Possibly! I just used `deno completions fig` to generate the updated one then merged them. Not entirely sure how to create a diff-versioned spec without using the github action though - could be missing something but couldn't find it in the docs. Could you point me in the right direction?

SeparateRecords:
Also this is ready to be merged rn @fedeci

fedeci:
> > Can we use diff-versioned specs for deno?
> 
> 
> 
> Possibly! I just used `deno completions fig` to generate the updated one then merged them. Not entirely sure how to create a diff-versioned spec without using the github action though - could be missing something but couldn't find it in the docs. Could you point me in the right direction?

https://fig.io/docs/concepts/advanced-specs that's it!

SeparateRecords:
🤦 right in front of me... thank you!!

src/deno.ts
```diff
@@ -287,7 +280,6 @@ const completion: Fig.Spec = {
           name: "files",
           isVariadic: true,
           isOptional: true,
-          template: "filepaths",
```
fedeci, line 290: Is this intended, then?

SeparateRecords: Not intended, thanks for catching this

PR #1435, opened 2022-08-20
feat: add minectl cli spec

dirien:
Signed-off-by: Engin Diri <engin.diri@mail.schwarz>

**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

Adds support for `minectl`, closes #1434

PR #1458, opened 2022-08-25
feat(minectl): update spec

dirien:

### DIFF
--- a/src/minectl.ts
+++ b/src/minectl.ts
@@ -55,10 +55,7 @@ const completionSpec: Fig.Spec = {
         {
           name: ["--filename", "-f"],
           description: "Contains the configuration for minectl",
-          args: {
-            name: "filename",
-            template: "filepaths",
-          },
+          args: { name: "filename" },
         },
         {
           name: ["--wait", "-w"],
@@ -73,10 +70,7 @@ const completionSpec: Fig.Spec = {
         {
           name: ["--filename", "-f"],
           description: "That contains the configuration for minectl",
-          args: {
-            name: "filename",
-            template: "filepaths",
-          },
+          args: { name: "filename" },
         },
         {
           name: "--id",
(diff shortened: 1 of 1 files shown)


fedeci:
Great! @dirien, what integration are you using?

dirien:
Hi @fedeci,

i tried the [Push To Autocomplete Action](https://fig.io/docs/guides/integrating/push-to-autocomplete-action#push-to-autocomplete-action) and I can see that all the feedback from the PR #1435 got removed.

Is there something I do wrong?

fedeci:
That's why I was asking which framework is minectl using.
To understand what's wrong here I need to explain how the merge tool works:
- Each CLI framework integration is capable of generating a determined set of fig schema properties.
- The merge tool that we use to merge the old spec with the new one needs to know which fig schema properties are generated by the integration and which are not, so we created presets for our first-party integrations. When merging it uses as a base the new spec and as diff the old one. In the resulting spec we will have all properties automatically generated (from the new spec so the old ones gets overridden) and all the properties from the old spec that are not automatically generated.

The problem here is likely that the integration used can generate the `template` property (the merge tool knows it from the preset) but the CLI framework is lacking some config that associates those args to a file.

e.g. for commander we print `template: "filepaths"` only when the argument is `<file>|[file]`.

dirien:
Thanks @fedeci for the explanation! Do you have an example using `cobra`? So I can see how to label the arguments correctly in the code.

I created several PRs for fig support in other cli tools too (goreleaser, cosign) and maybe I have to see what I need to do in my `minectl` cli before continue the work there

fedeci:
@dirien this is the line we use to detect if a filepaths template is used. https://github.com/withfig/autocomplete-tools/blob/94f003f2ede44499f6160b4857167be99a26fe52/integrations/cobra/main.go#L170

dirien:
Close draft for new PR

PR #1461, opened 2022-08-25
feat: add tree completion spec

tmwrnr:
feature

n/a

Adds a new spec for [tree](http://mama.indstate.edu/users/ice/tree/)

**Additional info:**
n/a

### DIFF
--- a/src/tree.ts
+++ b/src/tree.ts
@@ -1,0 +1,252 @@
+const completionSpec: Fig.Spec = {
+  name: "tree",
+  description: "Display directories as trees (with optional color/HTML output)",
+  args: {
+    template: "folders",
+    suggestCurrentToken: true,
+  },
+  options: [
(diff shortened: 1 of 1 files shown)


src/tree.ts
```diff
@@ -0,0 +1,252 @@
+const completionSpec: Fig.Spec = {
+  name: "tree",
+  description: "Display directories as trees (with optional color/HTML output)",
+  args: {
+    template: "folders",
+    suggestCurrentToken: true,
```
fedeci, line 6: When directory do not exist tree returns an error
```suggestion
```

fedeci (changes requested):
Thanks @tmwrnr! I left one comment then we can ship this spec!

fedeci (approved):
LGTM 🚀

PR #1463, opened 2022-08-26
feat: add capacitor completion spec

tmwrnr:
feature

n/a

Adds a new spec for [capacitor](https://capacitorjs.com/docs/cli)

**Additional info:**
I also added a `cap.ts` spec file because the capacitor CLI can be used either with `capacitor` or `cap` ([Reference](https://github.com/ionic-team/capacitor/blob/22c2fb505ee80dfce335c05f841ffafd21b7842c/cli/package.json#L34)). I was not sure if this is ok or if there is a better way of doing this. If you prefer not to have the second spec I can remove it again.

### DIFF
--- a/src/capacitor.ts
+++ b/src/capacitor.ts
@@ -3,6 +3,39 @@ const platforms: Fig.Suggestion[] = [
   { name: "ios", icon: "fig://icon?type=apple" },
 ];
 
+function isPlatform(value: string): value is "ios" | "android" {
+  return value === "ios" || value === "android";
+}
+
+const targetGenerator: Fig.Generator = {
+  cache: {
+    ttl: 1000 * 60, // Only caches targets for one minute, in case a new device is connected
+  },
+  custom: async (tokens, executeShellCommand) => {
+    const [cliName, command, platform] = tokens;
+    if (!isPlatform(platform)) {
+      return [];
+    }
+
+    const out = await executeShellCommand(
+      `npx capacitor run ${platform} --list`
+    );
+
+    return out
+      .trim()
+      .split("\n")
+      .slice(2) // Ignore output header lines
+      .map((s) => {
+        const [name, api, targetId] = s.replace(/\s\s+/g, "|").split("|");
+        return {
+          name: targetId,
+          displayName: `${name} ${api}`,
+          icon: "📱",
+        };
+      });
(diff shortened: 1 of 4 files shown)


src/capacitor.ts
```diff
@@ -0,0 +1,135 @@
+const platforms: Fig.Suggestion[] = [
+  { name: "android", icon: "fig://icon?type=android" },
+  { name: "ios", icon: "fig://icon?type=apple" },
+];
+
+const helpOption: Fig.Option = {
+  name: ["--help", "-h"],
+  description:
+    "Output usage information. Can be used with individual commands too",
+};
```
fedeci, line 10: If you add this property the option become available globally to all subcommands so it no longer needs to be repeated and can be moved directly in the completion object
```suggestion
  name: ["--help", "-h"],
  description:
    "Output usage information. Can be used with individual commands too",
  isPersistent: true,
};
```

src/capacitor.ts
```diff
@@ -0,0 +1,135 @@
...
+        {
+          name: "--list",
+          description:
+            "Print a list of target devices available to the given platform",
+        },
+        {
+          name: "--target",
+          description: "Run on a specific target device",
+          args: {
+            name: "target",
```
fedeci, line 80: Can you add a [generator](https://fig.io/docs/reference/generator) here? Ping me if you need help!

fedeci (changes requested):
Thanks! It seems capacitor can run through yarn and npx, could you add it to those specs too?

src/capacitor.ts
tmwrnr: Thanks that is way better 👍

src/capacitor.ts
```diff
@@ -3,10 +3,37 @@ const platforms: Fig.Suggestion[] = [
...
+      .slice(2) // Ignore output header lines
+      .map((s) => {
+        const [name, api, targetId] = s.replace(/\s\s+/g, "|").split("|");
+        return {
+          name: targetId,
+          displayName: `${name} ${api}`,
+          icon: "📱",
+        };
+      });
+  },
```
fedeci, line 36: Love it!

fedeci (approved):
Thanks!

PR #1468, opened 2022-08-30
feat: add mgnl completion spec

tmwrnr:
feature

n/a

Adds a new spec for [mgnl](https://docs.magnolia-cms.com/magnolia-cli/4.x/magnolia-cli.html)

**Additional info:**
n/a

### DIFF
--- a/src/mgnl.ts
+++ b/src/mgnl.ts
@@ -36,7 +36,8 @@ const lightModuleGenerator: Fig.Generator = {
 
 const completionSpec: Fig.Spec = {
   name: "mgnl",
-  description: "",
+  description:
+    "Magnolia CLI is a command line interface (CLI) tool to set up and facilitate light development with Magnolia",
   subcommands: [
     {
       name: "add-availability",
--- a/src/mgnl.ts
+++ b/src/mgnl.ts
@@ -1,0 +1,317 @@
+import { npmSearchGenerator } from "./npm";
+
+const pathOption = (description: string): Fig.Option => ({
+  name: ["-p", "--path"],
+  description,
+  args: {
+    name: "path",
+    template: "folders",
+  },
+});
+
+const prototypeOption = (description: string): Fig.Option => ({
+  name: ["-P", "--prototype"],
+  description,
+  args: {
+    name: "prototype",
+  },
+});
+
+const lightModuleGenerator: Fig.Generator = {
+  ...npmSearchGenerator,
+  custom: async (context, executeShellCommand, shellContext) => {
+    const searchTerm = context[context.length - 1];
+    if (searchTerm === "") {
+      return [];
+    }
+
+    const queryWithKeyword = searchTerm + "+keywords:magnolia-light-module";
+    return npmSearchGenerator.custom(
+      [queryWithKeyword],
+      executeShellCommand,
+      shellContext
+    );
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "mgnl",
+  description: "",
+  subcommands: [
+    {
+      name: "add-availability",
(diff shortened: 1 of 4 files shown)


src/mgnl.ts
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    const queryWithKeyword = searchTerm + "+keywords:magnolia-light-module";
+    return npmSearchGenerator.custom(
+      [queryWithKeyword],
+      executeShellCommand,
+      shellContext
+    );
+  },
+};
```
fedeci, line 35: This breaks when we query versions. The solution may be to refactor the npm generator logic to a function that accepts both the context and the eventual optional keywords.

src/mgnl.ts
```diff
@@ -0,0 +1,317 @@
...
+      args: {
+        name: "command",
+        suggestions: [{ name: "install" }, { name: "uninstall" }],
+      },
+    },
+    {
+      name: "search",
+      description: "Searches for Magnolia-related packages available from npm",
+      args: {
+        name: "query",
```
fedeci, line 290:
```suggestion
        name: "query",
        generators: lightModuleGenerator,
```

src/mgnl.ts
```diff
@@ -0,0 +1,317 @@
...
+      [queryWithKeyword],
+      executeShellCommand,
+      shellContext
+    );
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "mgnl",
+  description: "",
```
fedeci, line 39: Can you add a description here? It is useful for https://fig.io/manual

fedeci (changes requested):

fedeci:
Hey @tmwrnr, thanks for this completion! I left a couple comments, let me know what do you think.

src/mgnl.ts
tmwrnr: Good catch, I did not think about versions. Are the changes as you imagined?

tmwrnr:
> Hey @tmwrnr, thanks for this completion! I left a couple comments, let me know what do you think.

Thank you for the review! I hope the changes look good to you. I also added support for running the cli with `npx`.

src/npm.ts
```diff
@@ -32,14 +40,22 @@ export const npmSearchGenerator: Fig.Generator = {
   cache: {
     ttl: 1000 * 60 * 60 * 24 * 2, // 2 days
   },
-  custom: async (context, executeShellCommand) => {
+  custom: async (
+    context: string[],
+    executeShellCommand: Fig.ExecuteShellCommandFunction,
+    shellContext: Fig.ShellContext,
+    keywords?: string[]
+  ): Promise<Fig.Suggestion[]> => {
```
fedeci, line 48: We cannot change the signature of a function we call internally in the engine because it may change in the future. By refactoring I meant something like this: 
```ts
export const createNpmSearchHandler = (keywords?: string[]) => 
  async (
    context: string[],
    executeShellCommand: Fig.ExecuteShellCommandFunction,
    shellContext: Fig.ShellContext
  ) => {
    // the logic you wrote for the the generator goes here
  }
```
Then in the `npm` spec you can use it so:
```ts
export const npmSearchGenerator: Fig.Generator = {
  trigger: ...,
  getQueryTerm: ...,
  cache: ...,
  custom: createNpmSearchHandler(),
}
```
And in `@magnolia/cli`:
```ts
const magnoliaPackageGenerator: Fig.Generator = {
  ...npmSearchGenerator,
  custom: createNpmSearchHandler([/* keywords here */])
}
```

fedeci (changes requested):

fedeci (approved):
Thanks!

PR #1475, opened 2022-09-01
feat: add `osqueryi`

SeparateRecords:
CLI for https://github.com/osquery/osquery

This took about 5 mins total because [Helix](https://github.com/helix-editor/helix) is _amazing_ for creating specs out of help text

### DIFF
--- a/src/osqueryi.ts
+++ b/src/osqueryi.ts
@@ -1,0 +1,1013 @@
+import { valueList } from "@fig/autocomplete-generators";
+
+const tables = [
+  "account_policy_data",
+  "acpi_tables",
+  "ad_config",
+  "alf",
+  "alf_exceptions",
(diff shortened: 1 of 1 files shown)


src/osqueryi.ts
```diff
@@ -0,0 +1,1013 @@
...
+      name: "--disable_watchdog",
+      description: "Disable userland watchdog process",
+    },
+    {
+      name: "--enable_extensions_watchdog",
+      description: "Enable userland watchdog for extensions processes",
+    },
+    {
+      name: "--enable_tables",
+      args: { name: "value", generators: valueList({ values: tables }) },
```
SeparateRecords, line 314: This could use a simple generator + query term but `valueList` gives us that for free and soon will also not suggest already-provided values

PR #1480, opened 2022-09-05
feat(trivy): add additional `--format` suggestions

nevotheless:
The spec for trivy seems a bit broken in regards of `sbom` scans. Since i'm not too familiar with the trivy cli i only added the additional `--format`'s for the `trivy fs` command.

`trivy fs --format` is missing some formats.

`trivy fs --format` now includes additional formats.

**Additional info:**
The spec for `trivy sbom` seems to need changes as well.

### DIFF
--- a/src/trivy.ts
+++ b/src/trivy.ts
@@ -166,6 +166,27 @@ const formatOptions: Fig.Option = {
   },
 };
 
+const fsFormatOptions: Fig.Option = {
+  name: ["--format", "-f"],
+  description:
+    'Format (table, json, sarif, template) (default: "table") [$TRIVY_FORMAT]',
+  args: {
+    name: "format",
+    suggestions: [
+      "table",
+      "json",
+      "sarif",
+      "template",
+      "cyclonedx",
+      "spdx",
+      "spdx-json",
+      "github",
+      "cosign-vuln",
+    ],
+    default: "table",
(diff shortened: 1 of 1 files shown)


src/trivy.ts
```diff
@@ -350,6 +371,7 @@ const completionSpec: Fig.Spec = {
         "Scan local filesystem for language-specific dependencies and config files",
       subcommands: [
         ...scanOptions,
+        fsFormatOptions,
```
fedeci, line 374: `--format` is not a subcommand but it is an option instead. All the others too, can you change those? It should be as simple as changing `subcommands: [` to `options: [`
Also can you please check the others subcommands too? Apparently they have the same problem (e.g. the image subcommand)

fedeci (changes requested):
Thanks for this PR!

src/trivy.ts
nevotheless: Thanks for the feedback. Much appreciated :) I'll do that tomorrow and also take a look at the other subcommands as well.

fedeci: Thanks!

nevotheless: Sorry, i haven't forgotten. I have little time right now but i'll try to finish this PR as soon as i can.

PR #1490, opened 2022-09-13
feat(nextflow): Improve nextflow spec

mribeirodantas:

It's a feature. Improve autocomplete with generators.

The `run` command will only suggest downloaded projects, but not local files ending in `.nf` in the current directory.

Suggest also local files ending in `.nf` in the current directory.

### DIFF
--- a/src/nextflow.ts
+++ b/src/nextflow.ts
@@ -12,7 +12,7 @@ const runname: Fig.Generator = {
 
 const projectname: Fig.Generator = {
   script:
-    "find $HOME/.nextflow/assets/* -maxdepth 1 -type d | cut -d/ -f6,7 | grep / | grep -v assets",
+    "{ find * -maxdepth 0 -type f -name '*.nf' 2> /dev/null && find $HOME/.nextflow/assets/* -maxdepth 1 -type d | cut -d/ -f6,7 | grep / | grep -v assets; } 2> /dev/null",
   postProcess: (output) => {
     if (output == "") {
       return [];
--- a/src/nextflow.ts
+++ b/src/nextflow.ts
@@ -11,8 +11,7 @@ const runname: Fig.Generator = {
 };
 
 const projectname: Fig.Generator = {
-  script:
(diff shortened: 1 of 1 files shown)


src/nextflow.ts
```diff
@@ -12,7 +12,7 @@ const runname: Fig.Generator = {
 
 const projectname: Fig.Generator = {
   script:
-    "find $HOME/.nextflow/assets/* -maxdepth 1 -type d | cut -d/ -f6,7 | grep / | grep -v assets",
+    "{ find * -maxdepth 0 -type f -name '*.nf' 2> /dev/null && find $HOME/.nextflow/assets/* -maxdepth 1 -type d | cut -d/ -f6,7 | grep / | grep -v assets; } 2> /dev/null",
```
SeparateRecords, line 15: Hmm we recently changed how scripts are run, now it's run in the user's shell which means this will break for Fish users (`{ ... }` isn't valid fish syntax).

We can get around this by explicitly invoking a posix shell:
```suggestion
    `/bin/sh -c "{ find * -maxdepth 0 -type f -name '*.nf' 2> /dev/null && find $HOME/.nextflow/assets/* -maxdepth 1 -type d | cut -d/ -f6,7 | grep / | grep -v assets; } 2> /dev/null"`,
```
<br>

Or by making sure that stderr is always redirected:
```suggestion
    "find * -maxdepth 0 -type f -name '*.nf' 2>/dev/null && find $HOME/.nextflow/assets/* -maxdepth 1 -type d 2>/dev/null | cut -d/ -f6,7 | grep / | grep -v assets",
```

SeparateRecords (changes requested):

src/nextflow.ts
SeparateRecords: also cc @fedeci, want your thoughts on this

fedeci: Windows hates /dev/null right? I'll opt for the solution that guarantees the support to more OSs.

SeparateRecords: Honestly my experience with Windows is that it hates everything 😄 but yes I'd go for the first one myself. Looks like that's the consensus. @mribeirodantas quick fix when you're ready!

mribeirodantas: It's done!

mribeirodantas:
Thanks @SeparateRecords and @fedeci for the feedback 😄

PR #1491, opened 2022-09-13
feat(deno): update generators

clo4:

**Additional info:**

### DIFF
--- a/src/deno/generators.ts
+++ b/src/deno/generators.ts
@@ -657,7 +657,7 @@ export const generateTasks: Fig.Generator = {
 // --- Generate installed deno scripts
 
 export const generateInstalledDenoScripts: Fig.Generator = {
-  script: "\\find ~/.deno/bin -maxdepth 1 -perm -111 -type f",
+  script: "command find ~/.deno/bin -maxdepth 1 -perm -111 -type f",
   postProcess: (out) =>
     out
       .split("\n")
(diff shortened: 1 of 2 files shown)


src/deno/generators.ts
```diff
@@ -237,25 +237,331 @@ export const generateRunnableFiles = filepaths({
...
+    },
+    {
+      name: "ban-ts-comment",
+      description:
+        "Disallows the use of Typescript directives without a comment",
+    },
+    {
+      name: "ban-types",
+      description:
+        "Bans the use of primitive wrapper objects (e.g. `String` the object is a wrapper\nof `string` the primitive) in addition to the non-explicit `Function` type and\nthe misunderstood `Object` type",
```
SeparateRecords, line 255: Newlines should be replaced, and the script to generate the lint rules should be documented

src/deno/generators.ts
```diff
@@ -237,25 +237,331 @@ export const generateRunnableFiles = filepaths({
...
+      name: "no-extra-boolean-cast",
+      description: "Disallows unnecessary boolean casts",
+    },
+    {
+      name: "no-extra-non-null-assertion",
+      description: "Disallows unnecessary non-null assertions",
+    },
+    {
+      name: "no-extra-semi",
+      description: "<!-- deno-fmt-ignore-file -->",
```
fedeci, line 396: This seems to be a wrong description

SeparateRecords: Damn right that's the wrong description! 😆 Thanks for catching

PR #1492, opened 2022-09-13
feat: add stencil completion spec

tmwrnr:
feature

n/a

Adds a new spec for [stencil](https://stenciljs.com/docs/cli)

**Additional info:**
Is there a way to configure, that at least one options is required? The `test` command need either the `--spec` or `--e2e` option.


fedeci (approved):
Thanks

PR #1494, opened 2022-09-13
feat(man): increase cache TTL

SeparateRecords:
Can do this a couple of different ways. Currently the cache will be cleared when the generator is next used (lazily). We can also use `setTimeout` to proactively clear the cache after an hour, which may have memory benefits but idk if there are any drawbacks. Will experiment


mschrage (approved):

fedeci:
@SeparateRecords I am not sure if the `setTimeout` would actually work because of how we load spec.

PR #1498, opened 2022-09-15
feat(vim): remove suggestCurrentToken

SeparateRecords:
`suggestCurrentToken` for {n,}vim args was added because someone asked for it, but it's actually _much_ more annoying this way. This PR removes it.

Maybe we should have a way to indicate preference for `suggestCurrentToken` in editor args?

```
fig settings spec.vim.suggestCurrentToken true
fig settings spec.nvim.suggestCurrentToken true
```
And in the args:

```
{
  suggestCurrentToken: "spec.nvim.suggestCurrentToken",
}
```

### DIFF
--- a/src/nvim.ts
+++ b/src/nvim.ts
@@ -3,7 +3,7 @@ const completionSpec: Fig.Spec = {
   description: "Hyperextensible Vim-based text editor",
   args: {
     template: ["filepaths", "folders"],
-    suggestCurrentToken: true,
+    // suggestCurrentToken: true,
   },
   parserDirectives: {
     flagsArePosixNoncompliant: true,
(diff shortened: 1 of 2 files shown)


fedeci:
> `suggestCurrentToken` for {n,}vim args was added because someone asked for it, but it's actually _much_ more annoying this way. This PR removes it.
> 
> Maybe we should have a way to indicate preference for `suggestCurrentToken` in editor args?
> 
> ```
> fig settings spec.vim.suggestCurrentToken true
> fig settings spec.nvim.suggestCurrentToken true
> ```
> 
>  
> And in the args:
> 
> ```
> {
>   suggestCurrentToken: "spec.nvim.suggestCurrentToken",
> }
> ```

Love it, would you mind writing an RFC for this?

fedeci (approved):

PR #1502, opened 2022-09-15
feat(dtm): update spec

IronCore864:

### DIFF
--- a/src/dtm.ts
+++ b/src/dtm.ts
@@ -35,11 +35,7 @@ const completionSpec: Fig.Spec = {
         {
           name: ["--plugin-dir", "-d"],
           description: "Plugins directory",
-          args: {
-            name: "plugin-dir",
-            default: ".devstream",
-            template: ["folders"],
-          },
+          args: { name: "plugin-dir", template: ["folders"] },
         },
         {
           name: ["--yes", "-y"],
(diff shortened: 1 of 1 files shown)


src/dtm.ts
```diff
@@ -243,7 +232,7 @@ const completionSpec: Fig.Spec = {
             {
               name: ["--plugin-dir", "-d"],
               description: "Plugins directory",
-              args: { name: "plugin-dir", default: ".devstream" },
+              args: { name: "plugin-dir" },
```
SeparateRecords, line 235: Should this also have `template: "folders"`?

SeparateRecords (commented):
Looks good! Just one question then we should be good to merge 🙂

src/dtm.ts
fedeci: It should be handled in the CLI itself to avoid the same issue in the future

mschrage (approved):
LGTM! Thanks for contributing @IronCore864. How was the process & documentation around setting up autogenerated specs?

src/dtm.ts
fedeci, line 235:
```suggestion
              args: { name: "plugin-dir", template: "folders" },
```

PR #1504, opened 2022-09-15
feat(nextflow): Add runname generator for log subcommand

mribeirodantas:
Fix

Fig won't help the user when s/he wants to use `nextflow log` to a run name
without specifying one of the options: -but/-before/-after

Both the log subcommand and its options -but/-before/-after
accept a run name as argument. If the user provides a run name
both for the log subcommand and its other options, it's fine,
nextflow knows what to do.

Signed-off-by: Marcel Ribeiro-Dantas <mribeirodantas@seqera.io>

### DIFF
--- a/src/nextflow.ts
+++ b/src/nextflow.ts
@@ -539,6 +539,10 @@ const completionSpec: Fig.Spec = {
     {
       name: "log",
       description: "Print executions log and runtime info",
+      args: {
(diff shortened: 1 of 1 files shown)


SeparateRecords (approved):
LGTM, thanks @mribeirodantas!! 😄

PR #1505, opened 2022-09-15
feat(minectl): update spec

dirien:

### DIFF
--- a/src/minectl.ts
+++ b/src/minectl.ts
@@ -54,11 +54,8 @@ const completionSpec: Fig.Spec = {
       options: [
         {
           name: ["--filename", "-f"],
-          description: "Contains the configuration for minectl",
-          args: {
-            name: "filename",
-            template: "filepaths",
-          },
+          description: "Location of the manifest file",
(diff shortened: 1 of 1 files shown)


src/minectl.ts
```diff
@@ -217,9 +196,49 @@ const completionSpec: Fig.Spec = {
...
+        {
+          name: "version",
+          description: "Display the clients version information",
+        },
+        {
+          name: "wizard",
+          description:
+            "Calls the minectl wizard to create interactively a minectl 🗺 config",
+        },
+      ],
```
SeparateRecords, line 241: Given that this is autogenerated I'm not opposed to accepting this as-is, @fedeci wdyt? Should be fixed in upstream in the integration?

SeparateRecords (approved):
LGTM! Thanks @dirien 🙂
Waiting on review from @fedeci too.

src/minectl.ts
fedeci: I agree, it is autogenerated so we do not have to maintain it!

PR #1508, opened 2022-09-16
feat: add m-cli

BalliAsghar:
Adding completion spec for [m-cli](https://github.com/rgcr/m-cli) -  Swiss Army Knife for macOS


src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+                description: "Name of the location",
+              },
+            },
+            {
+              name: "delete",
+              description: "Delete a location",
+              args: {
+                name: "name",
+                description: "Name of the location",
+              },
```
SeparateRecords, line 810: Can make a generator:
```ts
const generateNetworkLocations: Fig.Generator = {
  script: "m network list | grep -e 'Device: ' | cut -d' ' -f2-",
  splitOn: "\n",
};
```

```suggestion
              args: {
                name: "name",
                description: "Name of the location",
                generators: generateNetworkLocations,
              },
```

And you can add this to the `switch` subcommand below

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+          name: "until",
+          description: "No sleep until",
+          subcommands: [
+            {
+              name: "pid",
+              description: "No sleep until process id ends",
+              args: {
+                name: "pid",
+                description: "Process id",
+              },
```
SeparateRecords, line 838: Can reuse the generator from `kill.ts`:
```ts
function getPidIcon(path: string): string {
  const idx = path.indexOf(".app/");
  if (idx === -1) {
    return "fig://icon?type=gear";
  }
  return "fig://" + path.slice(0, idx + 4);
}

const generatePids: Fig.Generator = {
  script: "ps axo pid,comm | sed 1d",
  postProcess: (result) => {
    return result.split("\n").map((line) => {
      const [pid, path] = line.trim().split(/\s+/);
      const name = path.slice(path.lastIndexOf("/") + 1);
      return {
        name: pid,
        description: path,
        displayName: `${pid} (${name})`,
        icon: getPidIcon(path),
      };
    });
  },
};
```

```suggestion
              args: {
                name: "pid",
                description: "Process id",
                generators: generatePids,
              },
```

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+            {
+              name: "time",
+              description: "Time in seconds",
+              isOptional: true,
+            },
+            {
+              name: "script",
+              description: "Path to script",
+              isOptional: true,
+            },
```
SeparateRecords, line 851:
```suggestion
            {
              name: "script",
              description: "Path to script",
              isOptional: true,
              template: "filepaths",
            },
```

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+          description: "Enable and show web interface",
+        },
+      ],
+    },
+    {
+      name: "restart",
+      description: "Restart computer (needs confirmation)",
+      subcommands: [
+        {
+          name: "-f",
```
SeparateRecords, line 929:
```suggestion
          name: ["-f", "--force"],
```

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+            isOptional: true,
+          },
+        },
+        {
+          name: "start",
+          description: "Start a service",
+          args: {
+            name: "service",
+            description: "Service name",
+          },
```
SeparateRecords, line 999: Can make a generator for services:
```ts
const generateServices: Fig.Generator = {
  script: "launchctl list | awk '{ print $3 }'",
  splitOn: "\n",
};
```

And you can use this in the args that take a service

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+          name: "create",
+          description: "Create a new user",
+        },
+        {
+          name: "delete",
+          description: "Delete a user",
+          args: {
+            name: "username",
+            description: "Username",
+          },
```
SeparateRecords, line 1109: The user generator we made earlier can be used again here!

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+    },
+    {
+      name: "wallpaper",
+      description: "Change wallpaper",
+      options: [
+        {
+          name: "",
+          description: "Path to the wallpaper",
+        },
+      ],
```
SeparateRecords, line 1208:
```suggestion
      args: {
        name: "image",
        description: "Path to the wallpaper",
        template: "filepaths",
      },
```

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+          description: "Scan wifi",
+        },
+        {
+          name: "showpassword",
+          description: "Show wifi network password",
+          args: {
+            name: "ESSID",
+            description: "ESSID",
+            isOptional: true,
+          },
```
SeparateRecords, line 1229: We'll reuse the generator from [wifi-password](https://github.com/withfig/autocomplete/blob/68a357150d28094967cb9e963cfbb2640605ca05/src/wifi-password.ts)

```ts
const generateWifiNetworks: Fig.Generator = generators: {
  script:
    "networksetup -listallhardwareports | awk '/Wi-Fi/{getline; print $2}' | xargs networksetup -listpreferredwirelessnetworks | tail -n +2",
  postProcess: (out) =>
    out
      .trim()
      .split("\n")
      .map((line) => ({ name: line.trim() })),
},
```

```suggestion
          args: {
            name: "ESSID",
            description: "ESSID",
            isOptional: true,
            generators: generateWifiNetworks,
          },
```

Can use the same generator for any other place that takes an ssid

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+        },
+        {
+          name: "ls",
+          description: "List known wifi networks",
+        },
+        {
+          name: "list",
+          description: "List known wifi networks",
+        },
```
SeparateRecords, line 1238: Can merge

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+    },
+    {
+      name: "--uninstall",
+      description: "Uninstall m-cli",
+    },
+    {
+      name: "help",
+      description: "Show help for specific command",
+      isPersistent: true,
+    },
```
SeparateRecords, line 1290: There's some parsing weirdness that happens when you have an option that has no leading dashes, you'd need to add `parserDirectives.flagsArePosixNoncompliant` which makes other things behave strangely. I think it's better overall to add the `help` subcommand to each command, honestly 🙂

SeparateRecords (approved):
WOW!! That's a huge spec. Thank you _so much_ for contributing this!

Lots of feedback but it's mostly ways to generate dynamic suggestions for each subcommand 😄

The only two pieces of feedback that apply broadly:
1. The commands have `help` subcommands, which should be added manually instead of with an option named `help`
2. The `list` commands have two names, `ls` and `list`, so that should be fixed on all of them. You can use an array for the name 🙂

(sorry for the delay btw, I've been slowly chipping away at writing the review since you opened the PR)

src/m.ts
```diff
@@ -0,0 +1,1294 @@
+const completionSpec: Fig.Spec = {
+  name: "m-cli",
+  description: " Swiss Army Knife for macOS",
```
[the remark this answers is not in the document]
BalliAsghar: I would remove the icon entirely/

src/m.ts
```diff
@@ -0,0 +1,1294 @@
...
+    },
+    {
+      name: "dns",
+      description: "Manage dns",
+      subcommands: [
+        {
+          name: "flush",
+          description: "Flush local DNS cache",
+        },
+      ],
```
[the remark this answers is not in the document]
BalliAsghar: it shows help command through isPersistent at options!

SeparateRecords (changes requested):
Feedback round 2! Fewer changes this time, should be fine to merge after this. Thanks so much for all the work you're putting into this, it's really paying off 🤩

After it's merged, would you be open to us tweeting about it from [@fig](https://twitter.com/fig)? Happy to give you a shoutout!

BalliAsghar:
> Feedback round 2! Fewer changes this time, should be fine to merge after this. Thanks so much for all the work you're putting into this, it's really paying off 🤩
> 
> After it's merged, would you be open to us tweeting about it from [@fig](https://twitter.com/fig)? Happy to give you a shoutout!

Sure! shoutout at [@BalliAsghar ](https://twitter.com/balli_asghar) would be great. thanks

SeparateRecords (approved):
This is so good. Thank you so much for all the work you put into this, it's awesome!!

PR #1513, opened 2022-09-21
feat: Add ddev spec for DDEV

khalwat:

Spec addition: `ddev` for the [DDEV](https://ddev.com/) local development environment

Currently `ddev` is not available

Adds the `ddev` spec

**Additional info:**

Leverages the Fig-provided Cobra module for generating the spec directly from the `ddev` command: https://github.com/drud/ddev/pull/4216

### DIFF
--- a/src/ddev.ts
+++ b/src/ddev.ts
@@ -3,40 +3,40 @@ const completionSpec: Fig.Spec = {
   description: "DDEV-Local local development environment",
   subcommands: [
     {
-      name: ["auth"],
+      name: "auth",
       description: "A collection of authentication commands",
       subcommands: [
         {
-          name: ["ssh"],
+          name: "ssh",
           description:
             "Add ssh key authentication to the ddev-ssh-auth container",
           options: [
             {
               name: ["--ssh-key-path", "-d"],
-              description: "full path to ssh key directory",
-              args: [{ name: "ssh-key-path" }],
+              description: "Full path to ssh key directory",
+              args: { name: "ssh-key-path" },
             },
           ],
         },
       ],
     },
     {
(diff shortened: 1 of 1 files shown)


SeparateRecords:
Looks good! Can you run `npm run lint:fix` to fix the linting errors?

khalwat:
Done. FWIW, the output here is coming directly from your own Cobra tool:

https://fig.io/docs/guides/integrating/integrations/cobra

...so if what it outputs isn't passing linting, probably it should be fixed as well?

SeparateRecords:
Agreed! It's not a huge deal because the linting errors can be fixed automatically and will be fixed when using our push to autocomplete CI action, but agreed, it should be as close as possible to the intended output by default.

SeparateRecords (approved):
Big PR so thank you for making it! Looks good to me but it's >1K lines so I'd like a second pair of eyes too just to make sure I haven't missed anything obvious. There's a couple of things I'd change but given that it's generated automatically most could (should?) be fixed upstream, so I'm not too concerned about that.

SeparateRecords:
cc @fedeci

khalwat:
If there's anything else you need me to do here, just let me know.

mschrage (approved):
Looks good to me! Thanks for contributing the GitHub action upstream to DDEV as well. 

This means the autocomplete for DDEV will always stay up to date. 🎉

khalwat:
How does it work in terms of an actual release that will be available for people to use?

fedeci:
@khalwat specs are automatically published once they are merged in but, because of caches, (npm registry and our CDN provider) they often need a couple days before being publicly available.

PR #1517, opened 2022-09-23
chore: clarify semver as dependency

### LINKED ISSUE #1516: [infracost] infracost spec depends on `semver`, but it is not specified in `package.json`
### Sanity checks

infracost

### Issue Details

The file `src/infracost/index.ts` imports the `semver` package, but `semver` is **not** listed as a dependency in `package.json`. **This causes errors when installing with pnpm** (which I happen to use), and it's generally considered good practice to list all dependencies in the project. A fix for this would be adding `"semver": "^7.3.7"` to the dependencies list in `package.json`.

doinkythederp:
This fixes a bug with the `infracost` spec.
Currently, the spec imports the `semver` package, but it is not listed as a dependency. This causes errors when using pnpm.
`semver` is added as a dependency in `package.json`.
**Additional info:**
Fixes #1516

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "@types/glob": "^7.2.0",
     "fs": "^0.0.1-security",
     "strip-json-comments": "^4.0.0",
-    "yaml": "^1.10.2"
+    "yaml": "^1.10.2",
+    "semver": "^7.3.7"
   }
 }


fedeci (approved):
Thanks!

PR #1519, opened 2022-09-25
feat: add dotenv-vault

motdotla:

Adding completion spec for [dotenv-vault](https://github.com/dotenv-org/dotenv-vault)

I and others enjoy Fig but have been missing autocompletion for dotenv-vault. Excited to get it into Fig's excellent autocompletion system!

### DIFF
--- a/src/dotenv-vault.ts
+++ b/src/dotenv-vault.ts
@@ -1,3 +1,5 @@
+import { filepaths } from "@fig/autocomplete-generators";
+
 // Common Options
 const yesOption: Fig.Option = {
   name: ["--yes", "-y"],
@@ -39,7 +41,7 @@ const filenameArg = {
   description:
     "Set input filename. Defaults to .env for development and .env.{environment} for other environments",
   isOptional: true,
-  template: "filepaths",
+  generators: filepaths({ matches: /^\.env.*$/ }),
 };
 
 const completionSpec: Fig.Spec = {
--- a/src/dotenv-vault.ts
+++ b/src/dotenv-vault.ts
@@ -5,11 +5,6 @@ const yesOption: Fig.Option = {
     "Automatic yes to prompts. Assume yes to all prompts and run non-interactively",
 };
 
-const helpOption: Fig.Option = {
-  name: "--help",
-  description: "Get documentation for the command",
-};
-
 const dotenvMeOption: Fig.Option = {
   name: ["-m", "--dotenvMe"],
   description:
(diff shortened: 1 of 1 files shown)


src/dotenv-vault.ts
```diff
@@ -0,0 +1,116 @@
...
+const yesOption: Fig.Option = {
+  name: ["--yes", "-y"],
+  description:
+    "Automatic yes to prompts. Assume yes to all prompts and run non-interactively",
+};
+
+const helpOption: Fig.Option = {
+  name: "--help",
+  description: "Get documentation for the command",
+};
```
SeparateRecords, line 11: Little simplification, you can actually use `persistent: true` on an option, put it on the root command, then it will show up on each subcommand! Means you can just define `--help` inline in the spec and you don't need to reference it in each subcommand.

```ts
  name: "dotenv-vault",
  description: "...",
  options: [
    {
      name: "--help",
      description: "Get documentation for the command",
      persistent: true,
    },
  ],
  // ...
}
```

src/dotenv-vault.ts
```diff
@@ -0,0 +1,116 @@
...
+  description: "Set environment to open to. Defaults to development",
+  isOptional: true,
+};
+
+const filenameArg = {
+  name: "FILENAME",
+  description:
+    "Set input filename. Defaults to .env for development and .env.{environment} for other environments",
+  isOptional: true,
+};
```
SeparateRecords, line 47: Adding `template: "filepaths"` to an arg gives you `cd`-like suggestions!

```suggestion
const filenameArg = {
  name: "FILENAME",
  description:
    "Set input filename. Defaults to .env for development and .env.{environment} for other environments",
  isOptional: true,
  template: "filepaths",
};
```

src/dotenv-vault.ts
```diff
@@ -0,0 +1,116 @@
...
+  description:
+    "Pass .env.me (DOTENV_ME) credential directly (rather than reading from .env.me file)",
+  args: {
+    name: "DOTENV_ME",
+    description: "Pass .env.me (DOTENV_ME) credential directly",
+  },
+};
+
+// Common Arguments
+const dotenvVaultArg = {
```
SeparateRecords, line 24: Absolutely not required but if you add the types you'll get nicer suggestions for the arg properties 🙂
```suggestion
const dotenvVaultArg: Fig.Arg = {
```

src/dotenv-vault.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const dotenvMeOption: Fig.Option = {
+  name: ["-m", "--dotenvMe"],
+  description:
+    "Pass .env.me (DOTENV_ME) credential directly (rather than reading from .env.me file)",
+  args: {
+    name: "DOTENV_ME",
+    description: "Pass .env.me (DOTENV_ME) credential directly",
+  },
```
SeparateRecords, line 20:
```suggestion
  args: {
    name: "DOTENV_ME",
    description: "Pass .env.me (DOTENV_ME) credential directly",
    template: "filepaths",
  },
```

SeparateRecords (changes requested):
Looks good @motdotla ! Thank you so much. Just left a couple of review comments, should be quick to resolve then we can merge.

src/dotenv-vault.ts
motdotla: Handy. Thank you.

src/dotenv-vault.ts
motdotla: This is really great!

Any plans to let me filter the type of files in the filepath? Ideally I would only want to display `.env*` files.

So a glob pattern here would be really wonderful. Something like `template: "filepaths:.env*"`

src/dotenv-vault.ts
motdotla: Thank you, but this one actually takes a string credential representing a `me_1234` token.

src/dotenv-vault.ts
motdotla: I'll skip this for now.

src/dotenv-vault.ts
SeparateRecords: Ahhh yes this is possible! We have `@fig/autocomplete-generators`, you can use a regex.

```ts
import { filepaths } from "@fig/autocomplete-generators";
```

Pretty sure this regex should work for you:
```ts
const filenameArg = {
  name: "FILENAME",
  description:
    "Set input filename. Defaults to .env for development and .env.{environment} for other environments",
  isOptional: true,
  generators: filepaths({ matches: /^\.env\..+$/ }),
};
```

Example from bun:
https://github.com/withfig/autocomplete/blob/ed9580a7e411f9a679a537eaa419e84d8fb7f09c/src/bun.ts#L100

SeparateRecords (approved):
Amazing, thank you @motdotla ! 😄

PR #1524, opened 2022-09-28
feat(gh): Add completion for aliases

SeparateRecords:
CleanShot 2022-09-28 at 11 02 10@2x

Adds support for completions for gh aliases


Alainmma (commented):
https://github.com/withfig/autocomplete/pull/1524#issue-1388572694

PR #1559, opened 2022-10-01
feat(typeorm): make spec diff-versioned

fedeci:


fedeci:
This makes it really verbose but at the same time we should start to expand and try to bring as many open source projects as possible to be directly integrated with autocomplete!

fedeci:
Closes #1158

loosheng:
> This makes it really verbose by duplicating the spec

I agree with you, I thought of this when I added autocomplete to `pnpm`.

I have an idea, maybe we should add a dependsOnVersion for `subcommand`
```ts
interface Subcommand {
	dependsOnVersion: (string | (executeShellCommand) => boolean)[]
}
```

When it's a string, you can refer directly to the `package.json` version control, which is a very mature solution.

Version range | Matching version
-- | -- 
`* ` | You can use `subcommand` regardless of the version, and this will be the default
`^2.1.0` | 2.1.0 <= v < 3  Versions available
`~2.1.0` | 2.1.0 <= v < 2.2 Versions available
`2.1.0 - 3.2.7` | 2.1.0 <= v <= 3.2.7 Versions available
`<2.2.0` | ...
`<=2.0.0` | ...
`\>4.2.0` | ...
`\>=4.2.0` | ...
`=4.3.0` | ...
`^0.7 \|\| ~2` | ...

When it's a function, we need to take into account that some commands depend not only on their own version but also on the version of the external environment, e.g. `pnpm` v7 requires `nodejs` >= 14, at which point we can do some custom comparisons in the function and return whether the version is available in the current environment.

eg:
```ts
{
// pnpm v7 
// pnpm patch
  {
    name: "patch",
    description: `This command will cause a package to be extracted in a temporary directory intended to be editable at will`,
	dependsOnVersion: [
		'\>=7.14.0',
		async ( executeShellCommand ){
			const nodeVersion = await executeShellCommand('node -v');
			return compareVersions(nodeVersion,'>=14')
		},
	// Maybe there are multiple environmental dependencies?
	],
    args: {
      name: "package",
    },
    options: [
      {
        name: "--edit-dir",
        description: `The package that needs to be patched will be extracted to this directory`,
      },
    ],
  },

}
```

fedeci:
Hey @loosheng, we already have something like for subcommands this and it is called `generateSpec`. However since these solutions are really difficult to automate (they use functions to determine CLI tool version) we created diff-versioned specs like this one.

PR #1563, opened 2022-10-03
feat(dscl): Add completion specs

### LINKED ISSUE #1545: [dscl] missing spec

plibither8:

Adds missing completion specs.

Missing completion specs for the `dscl` command.

Adds the `dscl` command specs.

**Additional info:**

Closes #1545

### DIFF
--- a/src/dscl.ts
+++ b/src/dscl.ts
@@ -4,6 +4,59 @@ const satisfies =
   <U extends T>(u: U) =>
     u;
 
+const dsclOptions: Fig.Option[] = [
+  {
+    name: "-p",
+    description: "Prompt for password",
+  },
+  {
+    name: "-u",
+    description: "Authenticate as user",
+    args: {
+      name: "user",
+      description: "User to authenticate as",
+    },
+  },
+  {
+    name: "-P",
+    description: "Authenticate with password",
+    args: {
+      name: "password",
+      description: "Password to authenticate with",
+    },
+  },
+  {
+    name: "-f",
+    description: "Targeted local node database file path",
+    args: {
+      name: "file",
+      description: "File path",
+    },
+  },
+  {
+    name: "-raw",
+    description: "Don't strip off prefix from DirectoryService API constants",
+  },
+  {
+    name: "-plist",
+    description: "Print out record(s) or attribute(s) in XML plist format",
+  },
+  {
+    name: "-url",
+    description: "Print record attribute values in URL-style encoding",
+  },
+  {
+    name: "-q",
+    description: "Quiet - no interactive prompt",
+  },
+];
+
+const dsclOptionsWithArgs = new Set<string>(
+  dsclOptions
+    .filter((option) => option.args !== undefined)
+    .flatMap((option) => option.name)
+);
+
 const generateDsclPath: Fig.Generator = {
   trigger: "/",
   getQueryTerm: "/",
@@ -14,7 +67,7 @@ const generateDsclPath: Fig.Generator = {
     }
     // can't guarantee that dscl is the first token. If it's not found,
     // this is -1, so searching for `datasource` will start from index 0
-    const dsclIndex = tokens.findIndex((token) => token === "dscl");
+    const dsclIndex = tokens.indexOf("dscl");
     let datasource: string;
     for (let i = dsclIndex + 1; i < tokens.length; i++) {
       const token = tokens[i];
@@ -24,11 +77,7 @@ const generateDsclPath: Fig.Generator = {
       }
       // skip if the previous token is a flag that takes an argument
       const prev = tokens[i - 1];
-      if (
-        dsclOptions.find(
-          (option) => [option.name].flat().includes(prev) && !!option.args
-        )
-      ) {
+      if (dsclOptionsWithArgs.has(prev)) {
         continue;
       }
       datasource = token;
@@ -44,9 +93,7 @@ const generateDsclPath: Fig.Generator = {
       .map((line) => ({
         name: line,
         icon: "📁",
-        priority: line.slice(line.lastIndexOf("/") + 1).startsWith("_")
-          ? 49
-          : 50,
+        priority: line[line.lastIndexOf("/") + 1] === "_" ? 49 : 50,
       }));
   },
 };
@@ -252,53 +299,6 @@ const dsclSubcommands: Fig.Subcommand = {
     },
   ],
 };
-
-const dsclOptions: Fig.Option[] = [
-  {
-    name: "-p",
-    description: "Prompt for password",
-  },
-  {
-    name: "-u",
-    description: "Authenticate as user",
-    args: {
-      name: "user",
-      description: "User to authenticate as",
-    },
-  },
-  {
-    name: "-P",
-    description: "Authenticate with password",
-    args: {
-      name: "password",
-      description: "Password to authenticate with",
-    },
-  },
-  {
-    name: "-f",
-    description: "Targeted local node database file path",
-    args: {
-      name: "file",
-      description: "File path",
-    },
-  },
-  {
-    name: "-raw",
-    description: "Don't strip off prefix from DirectoryService API constants",
-  },
-  {
-    name: "-plist",
-    description: "Print out record(s) or attribute(s) in XML plist format",
-  },
-  {
-    name: "-url",
-    description: "Print record attribute values in URL-style encoding",
-  },
-  {
-    name: "-q",
-    description: "Quiet - no interactive prompt",
-  },
-];
 
 const completionSpec: Fig.Spec = {
   name: "dscl",
--- a/src/dscl.ts
+++ b/src/dscl.ts
@@ -1,0 +1,225 @@
+const args: Record<string, Fig.Arg> = {
+  path: {
+    name: "path",
+    description: "Path to the record",
+  },
+  key: {
+    name: "key",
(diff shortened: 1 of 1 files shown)


src/dscl.ts
```diff
@@ -0,0 +1,225 @@
...
+    {
+      name: "-q",
+      description: "Quiet - no interactive prompt",
+    },
+  ],
+  args: {
+    name: "datasource",
+    suggestions: [".", "..", "localhost", "localonly"],
+    isOptional: true,
+    loadSpec: "dscl",
```
fedeci, line 221: Is `loadSped` here being used to enforce argument before commands?

plibither8: Sorta... yes. `dscl` has an awkward format, it requires the datasource argument before the actual subcommands. This was the best available workaround I could find, can this be done in a better way?

src/dscl.ts
```diff
@@ -0,0 +1,225 @@
...
+  },
+};
+
+// Add "optional" arguments
+for (const [name, arg] of Object.entries(args)) {
+  args[`${name}Optional`] = {
+    ...arg,
+    isOptional: true,
+  };
+}
```
SeparateRecords, line 48: We can make this a bit more type-safe to prevent accidental `undefined`s from future editors:

```suggestion
const args = {
  path: {
    name: "path",
    description: "Path to the record",
  } as Fig.Arg,
  key: {
    name: "key",
    description: "Directory Service record attribute type",
  } as Fig.Arg,
  keys: {
    name: "keys",
    description: "Directory Service record attribute type",
    isVariadic: true,
  } as Fig.Arg,
  plistPath: {
    name: "plist path",
    description: "Path to the plist file",
  } as Fig.Arg,
  valueIndex: {
    name: "value index",
    description: "Value index of the key",
  } as Fig.Arg,
  val: {
    name: "val",
    description: "Value of the key",
  } as Fig.Arg,
  vals: {
    name: "vals",
    description: "Value of the key",
    isVariadic: true,
  } as Fig.Arg,
  oldVal: {
    name: "old val",
    description: "Old value of the key",
  } as Fig.Arg,
  newVal: {
    name: "new val",
    description: "New value of the key",
  } as Fig.Arg,
} as const;

const arg = (name: keyof typeof args, merge?: Partial<Fig.Arg>): Fig.Arg => {
  return { ...args[name], ...merge };
}
```

You can use it like so:
```ts
arg("path", { isOptional: true });
```

src/dscl.ts
```diff
@@ -0,0 +1,225 @@
+const args: Record<string, Fig.Arg> = {
+  path: {
+    name: "path",
+    description: "Path to the record",
+  },
```
SeparateRecords, line 5: A generator for DSCL paths:
```ts
const generateDsclPath: Fig.Generator = {
  trigger: "/",
  getQueryTerm: "/",
  custom: async (tokens, executeShellCommand) => {
    const lastToken = tokens[tokens.length - 1];
    if (lastToken === "") {
      return [];
    }
    const dsclIndex = tokens.findIndex((token) => token === "dscl");
    let datasource: string;
    for (let i = dsclIndex + 1; i < tokens.length; i++) {
      const token = tokens[i];
      // skip if the current token is a flag
      if (token.startsWith("-")) {
        continue;
      }
      // skip if the previous token is a flag that takes an argument
      const prev = tokens[i - 1];
      if (prev === "-u" || prev === "-P" || prev === "-f") {
        continue;
      }
      datasource = token;
      break;
    }
    const search = lastToken.slice(0, lastToken.lastIndexOf("/"));
    let command = `dscl ${datasource} -list ${search}`;
    if (search === "") {
      command = `dscl ${datasource} -list /`;
    }
    const lines = await executeShellCommand(command);
    return lines.trim().split("\n").map((line) => ({
      name: line,
      icon: "📁",
      priority: line.slice(line.lastIndexOf("/") + 1).startsWith("_") ? 49 : 50,
    }));
  },
};
```

```suggestion
  path: {
    name: "path",
    description: "Path to the record",
    generators: generateDsclPath,
  },
```

src/dscl.ts
SeparateRecords: We can move the subcommands into their own command like this:

```ts
const dsclSubcommands: Fig.Subcommand = {
  name: "dscl",
  subcommands: [
    // ...
  ],
};
```

Then after defining that, reference it in the main spec:

```ts
  args: {
    name: "datasource",
    suggestions: [".", "..", "localhost", "localonly"],
    isOptional: true,
    loadSpec: dsclSubcommands,
  },
```

This is currently broken by the autocomplete engine, will fix this.

src/dscl.ts
```diff
@@ -0,0 +1,225 @@
...
+      args: [args.path, args.key, args.vals],
+    },
+    {
+      name: ["merge", "-merge"],
+      description:
+        "Appends one or more values to a property in a given directory if the property does not already have those values",
+      args: [args.path, args.key, args.vals],
+    },
+    {
+      name: ["delete", "-delete", "rm"],
```
SeparateRecords, line 120:
```suggestion
      name: ["delete", "-delete", "rm", "-rm"],
```

src/dscl.ts
```diff
@@ -0,0 +1,225 @@
...
+      description: "Lists the subdirectories of the given directory",
+      args: [args.path, args.keyOptional],
+    },
+    {
+      name: ["search", "-search"],
+      description: "Searches for records that match a pattern",
+      args: [args.path, args.key, args.val],
+    },
+    {
+      name: ["create", "-create", "mk"],
```
SeparateRecords, line 86:
```suggestion
      name: ["create", "-create", "mk", "-mk"],
```

src/dscl.ts
```diff
@@ -0,0 +1,225 @@
...
+      args: [args.path, args.key, args.plistPath],
+    },
+    {
+      name: ["readpli", "-readpli"],
+      description:
+        "Prints the contents of plist_path for the plist at value_index of the key",
+      args: [args.path, args.key, args.valueIndex, args.plistPath],
+    },
+    {
+      name: ["list", "-list", "ls"],
```
SeparateRecords, line 76:
```suggestion
      name: ["list", "-list", "ls", "-ls"],
```

SeparateRecords (changes requested):

src/dscl.ts
plibither8: Add `as const` doesn't work here as expected, resorted to a `satisfies` utility function that extracts the exact types.

src/dscl.ts
plibither8: This is cool! Made some minor changes to the function to make it future-proof :)

src/dscl.ts
plibither8: Implemented, thanks!

src/dscl.ts
```diff
@@ -0,0 +1,316 @@
...
+    const path = lastToken.slice(0, lastSlashIndex < 0 ? 0 : lastSlashIndex);
+    const command = `dscl ${datasource} -list ${path || "/"}`;
+    const lines = await executeShellCommand(command);
+    return lines
+      .trim()
+      .split("\n")
+      .map((line) => ({
+        name: line,
+        icon: "📁",
+        priority: line.slice(line.lastIndexOf("/") + 1).startsWith("_")
```
SeparateRecords, line 47: Just realized the cleaner+faster way of doing this 😄
Ultimately doesn't _really_ matter but it's still perf increase
```suggestion
        priority: line[line.lastIndexOf("/") + 1] === "_"
```

src/dscl.ts
```diff
@@ -0,0 +1,316 @@
...
+  trigger: "/",
+  getQueryTerm: "/",
+  custom: async (tokens, executeShellCommand) => {
+    const lastToken = tokens[tokens.length - 1];
+    if (lastToken === "") {
+      return [];
+    }
+    // can't guarantee that dscl is the first token. If it's not found,
+    // this is -1, so searching for `datasource` will start from index 0
+    const dsclIndex = tokens.findIndex((token) => token === "dscl");
```
SeparateRecords, line 17: Not sure why I went straight to `findIndex` the first time...
```suggestion
    const dsclIndex = tokens.indexOf("dscl");
```

src/dscl.ts
```diff
@@ -0,0 +1,316 @@
...
+      if (token.startsWith("-")) {
+        continue;
+      }
+      // skip if the previous token is a flag that takes an argument
+      const prev = tokens[i - 1];
+      if (
+        dsclOptions.find(
+          (option) => [option.name].flat().includes(prev) && !!option.args
+        )
+      ) {
```
SeparateRecords, line 31: This is really clever!! We could extract this so we don't have to check it every loop, computed on load instead:
```ts

const dsclOptionsWithArgs = new Set<string>(
  dsclOptions
    .filter((option) => option.args !== undefined)
    .flatMap((option) => option.name),
);

const generateDsclPath: Fig.Generator = {
  // ...
```

Then we can just check the Set:

```suggestion
      if (dsclOptionsWithArgs.has(prev)) {
```

SeparateRecords (changes requested):
Fantastic! Just a couple of things to change - two of which were mine 😅

src/dscl.ts
plibither8: Good suggestion! 😄

SeparateRecords (approved):
LGTM! Thank you so much!!

PR #1555, opened 2022-09-30
feat: add gem subcommands

AlanSyue:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
feature

**What is the current behavior? (You can also link to an open issue here)**
N/A

**What is the new behavior (if this is a feature change)?**
Add more gem subcommands

**Additional info:**
According to the [official document](https://guides.rubygems.org/command-reference/), GEM has the following commands:

- [x] [gem build](https://guides.rubygems.org/command-reference/#gem-build)
- [x] [gem cert](https://guides.rubygems.org/command-reference/#gem-cert)
- [x] [gem check](https://guides.rubygems.org/command-reference/#gem-check)
- [x] [gem cleanup](https://guides.rubygems.org/command-reference/#gem-cleanup)
- [x] [gem contents](https://guides.rubygems.org/command-reference/#gem-contents)
- [x] [gem dependency](https://guides.rubygems.org/command-reference/#gem-dependency)
- [x] [gem environment](https://guides.rubygems.org/command-reference/#gem-environment)
- [x] [gem fetch](https://guides.rubygems.org/command-reference/#gem-fetch)
- [x] [gem generate_index](https://guides.rubygems.org/command-reference/#gem-generate_index)
- [x] [gem help](https://guides.rubygems.org/command-reference/#gem-help)
- [x] [gem info](https://guides.rubygems.org/command-reference/#gem-info)
- [x] [gem install](https://guides.rubygems.org/command-reference/#gem-install)
- [x] [gem list](https://guides.rubygems.org/command-reference/#gem-list)
- [x] [gem lock](https://guides.rubygems.org/command-reference/#gem-lock)
- [x] [gem mirror](https://guides.rubygems.org/command-reference/#gem-mirror)
- [x] [gem open](https://guides.rubygems.org/command-reference/#gem-open)
- [x] [gem outdated](https://guides.rubygems.org/command-reference/#gem-outdated)
- [x] [gem owner](https://guides.rubygems.org/command-reference/#gem-owner)
- [x] [gem pristine](https://guides.rubygems.org/command-reference/#gem-pristine)
- [x] [gem push](https://guides.rubygems.org/command-reference/#gem-push)
- [x] [gem query](https://guides.rubygems.org/command-reference/#gem-query)
- [x] [gem rdoc](https://guides.rubygems.org/command-reference/#gem-rdoc)
- [x] [gem search](https://guides.rubygems.org/command-reference/#gem-search)
[...]

PR #1565, opened 2022-10-03
feat(gem): add update & outdated subcommands

cadenkriese:
Adds completion spec for two subcommands of `gem`.

Missing completion spec.

**What is the new behavior**
Implemented completion spec.

### DIFF
--- a/src/gem.ts
+++ b/src/gem.ts
@@ -24,9 +24,23 @@ const completionSpec: Fig.Spec = {
       name: ["install", "i"],
       description: "Install a gem into the local repository",
       args: {
-        name: "GEMNAME",
+        name: "gem",
         generators: gems,
         debounce: true,
+      },
+    },
+    {
+      name: "outdated",
+      description: "Display all gems that need updates",
+    },
+    {
+      name: "update",
+      description: "Update installed gems to the latest version",
+      args: {
+        name: "gems",
+        description: "The gem(s) you want to update",
+        isVariadic: true,
+        isOptional: true,
       },
     },
   ],
--- a/src/gem.ts
+++ b/src/gem.ts
@@ -77,6 +77,7 @@ const completionSpec: Fig.Spec = {
           description: "Gem repository directory to get installed gems",
           args: {
             name: "DIR",
+            template: "folders",
           },
         },
         {
@@ -85,6 +86,7 @@ const completionSpec: Fig.Spec = {
             "Directory where executables will be placed when the gem is installed",
           args: {
             name: "DIR",
+            template: "folders",
           },
         },
         {
@@ -92,6 +94,7 @@ const completionSpec: Fig.Spec = {
           description: "Generate documentation for installed gems",
           args: {
             name: "TYPES",
+            suggestions: ["rdoc", "ri"],
           },
         },
         {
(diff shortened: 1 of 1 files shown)


SeparateRecords:
Hey @cadenkriese! Thanks for the PR 🙂

So this just adds the `outdated` and `update` subcommands but no options -- this is totally fine! Was wondering if you'd also be able to add their flags? No worries if you don't, we can absolutely merge as-is.

SeparateRecords:
(also, https://github.com/withfig/autocomplete/pull/1555 is open, but this has a much smaller scope so can be merged sooner)

cadenkriese:
No problem, I added the options. However, everything after `--prerelease` is duplicated between `gem install` and `gem update`. Should I put them in their own variable to share? Worth noting that I'm not familiar with typescript syntax, so pseudocode is much appreciated.

One other question, a lot of these options have an optional 'no-' prefix, ie. `--no-prerelease` to explicitly disable them, does fig have a convention for this?

src/gem.ts
```diff
@@ -29,6 +29,180 @@ const completionSpec: Fig.Spec = {
...
+          args: {
+            name: "DIR",
+          },
+        },
+        {
+          name: "--document",
+          description: "Generate documentation for installed gems",
+          args: {
+            name: "TYPES",
+          },
```
SeparateRecords, line 95: Do we know ahead of time what types are available? If we do, let's add `suggestions`! It's an array of strings

src/gem.ts
```diff
@@ -29,6 +29,180 @@ const completionSpec: Fig.Spec = {
...
+          name: "--post-install-message",
+          description: "Print post install message",
+        },
+        {
+          name: ["--file", "-g"],
+          description:
+            "Read from a gem dependencies API file and install the listed gems",
+          args: {
+            name: "FILE",
+          },
```
SeparateRecords, line 176: Can add `template: "filepaths"` for file suggestions 🙂

SeparateRecords (changes requested):
> everything after --prerelease is duplicated between gem install and gem update. Should I put them in their own variable to share?

No need! I'm a big fan of WET (write everything twice), no need to simplify it yet. Can be refactored later to clean up 🙂

> Worth noting that I'm not familiar with typescript syntax, so pseudocode is much appreciated.

That's okay! Only thing you really need to know for Fig:
```
const [varname]: [type] = [essentially a json object]
```

For a TypeScript array, just add `[]` after the type name, so a list of Fig options is

```ts
const options: Fig.Option[] = [
  // objects go here
];
```

We've designed the API surface to be as minimal as possible so there's not much to learn, unless you want to get into some _really_ wild and wacky generators like [`deno doc`](https://github.com/withfig/autocomplete/blob/589d8d2d80fb301fb249573483179429941d5a39/src/deno/generators.ts#L7-L184) (provides real-time symbol suggestions)

https://user-images.githubusercontent.com/52195359/194490777-7f814728-59a0-4a61-b516-5f238d0dad45.mp4

> One other question, a lot of these options have an optional 'no-' prefix, ie. --no-prerelease to explicitly disable them, does fig have a convention for this?

There's no convention for this, you've done it perfectly!

src/gem.ts
cadenkriese: I could only find the types 'rdoc' and 'ri' so I've added those. 👍

SeparateRecords (approved):
Looks good to me! Thanks @cadenkriese  🙂

PR #1568, opened 2022-10-04
feat(echo): add environment variable generator

duncanlutz:
Feature: Generate args for environment variables on the users system.

Fig currently only returns -n, -e and -E as options for the echo command.

Displays an arg for each environment currently contained on the system

**Additional info:**


fedeci (approved):
Thanks! Looks great and I can't wait to have this available!

PR #1573, opened 2022-10-04
feat(z): add z cache results to completion

sghsri:
Feature

The current behavior is that z only shows the paths in the current folder. 

The PR changes the z completion to take into account the directories cached by z and the weighting that z provides them. Now you will be able to see all the possible paths that z can take you.

image

**Additional info:**

### DIFF
--- a/src/z.ts
+++ b/src/z.ts
@@ -3,10 +3,53 @@ const completionSpec: Fig.Spec = {
   name: "z",
   description: "CLI tool to jump around directories",
   args: {
-    template: ["folders"],
-    name: "regex",
+    name: "directory",
     isVariadic: true,
     isOptional: true,
+    generators: {
+      script: "cat ${${ZSHZ_DATA:-${_Z_DATA:-${HOME}/.z}}:A}",
+      postProcess: (out, ctx) => {
+        const lines = out.split("\n").map((line) => {
+          const [path, weight, time] = line.split("|");
+          return {
+            path,
+            weight: Number(weight),
+            time: Number(time),
+          };
+        });
+
+        // filter for valid "directory" entries within the arguments
+        const args = ctx.filter(
+          (arg) => arg && arg !== "z" && !arg.startsWith("-")
+        );
+
+        // directory arg is variadic with each subsequent arg being an
+        // additional filter. filtered will filter directories by all args.
+        const filtered = lines.filter(({ path }) =>
+          args.every((arg) => path.includes(arg))
+        );
+
+        return filtered.map(({ path, weight, time }) => {
+          const splitPath = path.split("/");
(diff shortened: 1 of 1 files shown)


src/z.ts
```diff
@@ -3,10 +3,53 @@ const completionSpec: Fig.Spec = {
...
+          // that's already been entered. This 'if' prevents redundant
+          // suggestions.
+          if (!args.includes(name)) {
+            return {
+              name,
+              description: path,
+              // Docs state max weight is 100 but this seems
+              // to work regardless of any amount over that limit.
+              // Fig should defer assigning priority to z.
+              // 75 added to keep args above options.
```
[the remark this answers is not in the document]
sghsri: ok, good call! just pushed up a commit doing just that

src/z.ts
```diff
@@ -44,7 +44,7 @@ const completionSpec: Fig.Spec = {
               // to work regardless of any amount over that limit.
               // Fig should defer assigning priority to z.
               // 75 added to keep args above options.
-              priority: 75 + weight,
+              priority: 75 + (weight >= 100 ? 100 : weight),
```
fedeci, line 47: This isn't really what I was thinking. My idea was more like this: the max sum of all the frecencies of z is either a custom value or 9000 https://github.com/rupa/z/blob/b82ac78a2d4457d2ca09973332638f123f065fd1/z.sh#L72
We should assume that all the items are going to be in the range `[0, MAX || 9000]`.
This should do the tweak:
```suggestion
              priority: 75 + (weight * 25 / (maxFrecency || 9000)), // where maxFrecency is read executing a shell command: `echo "$_Z_MAX_SCORE"`
```
Ping me for help!

sghsri: oooh that makes sense! how would we go about executing another shell command here?

sghsri: also, when echoing that shell variable even though I have z active, its empty 
.

sghsri: I'm using https://github.com/agkozak/zsh-z , which is said to be more stable and more performant

fedeci: Updated!

sghsri: Ok awesome, pushed up the commit with the suggestion!

fedeci (approved):
Thanks again for contributing!

fedeci:
@sghsri linter is failing :( Can you run `npm run lint:fix` and commit the changes?

sghsri:
> @sghsri linter is failing :( Can you run `npm run lint:fix` and commit the changes?
commit after running `npm run lint:fix` pushed up!

sghsri:
@fedeci thanks for approving! What is the next step in the process for getting this merged in / available in the production fig build?

fedeci:
Merged!

Issue #1527, opened 2022-09-30
[which] doesn't suggest programs


SeparateRecords:
`which` could suggest programs using a generator

I'm happy to help with this! Feel free to ask questions on our discord https://fig.io/community

davidalpert:
addressed in #1574

PR #1574, opened 2022-10-04
feat(which): add generator to pull executables from $PATH

duncanlutz:
Feature

Currently which only suggests the -s and -a options

Suggests names of programs located in $PATH variable when running the "which" command ([issue that suggested this feature](https://github.com/withfig/autocomplete/issues/1527))

**Additional info:**

### DIFF
--- a/src/which.ts
+++ b/src/which.ts
@@ -1,9 +1,19 @@
+const programGenerator: Fig.Generator = {
+  script: `for i in $(echo $PATH | tr ":" "\n"); do find $i -maxdepth 1 -perm -111 -type f; done`,
+  postProcess: (out) =>
+    out
+      .split("\n")
+      .map((path) => path.split("/")[path.split("/").length - 1])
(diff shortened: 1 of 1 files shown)


src/which.ts
```diff
@@ -1,9 +1,19 @@
...
+      .map((pr) => ({ name: pr, description: "Executable file", type: "arg" })),
+};
+
 const completionSpec: Fig.Spec = {
   name: "which",
   description: "Locate a program in the user's PATH",
   args: {
     name: "names",
     isVariadic: true,
+    generators: programGenerator,
```
fedeci, line 16:
```suggestion
    generators: programGenerator,
    filterStrategy: "fuzzy",
```

fedeci (changes requested):
Thanks for the PR! The suggestion I left improves the UX allowing to search from the list in a fuzzy way!

fedeci (approved):
Thanks!

PR #1581, opened 2022-10-04
feat(login): Add completion spec

### LINKED ISSUE #1537: [login] missing spec

mrhappyma:
Add completion spec for `login` command
Based off [Ubuntu manpage](https://manpages.ubuntu.com/manpages/jammy/en/man1/login.1.html), potentially other OS have additional syntax and this won't work?

Closes #1537

### DIFF
--- a/src/login.ts
+++ b/src/login.ts
@@ -1,0 +1,41 @@
+const completionSpec: Fig.Spec = {
+  name: "login",
+  description: "Begin session on the system",
+  options: [
+    {
+      name: "-p",
+      description: "Preserve environment",
+    },
+    {
+      name: "-r",
+      description: "Perform autologin protocol for rlogin",
+    },
+    {
+      name: "-h",
+      description: "Specify host",
+      args: {
+        name: "host",
+      },
+    },
+    {
+      name: "-f",
+      description: "Don't authenticate user, user is preauthenticated",
+    },
+  ],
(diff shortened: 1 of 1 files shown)


SeparateRecords:
Interesting, this is maybe the first spec where the options are _quite_ different between Linux and macOS. Here's from the Mac man-page:

CleanShot 2022-10-05 at 17 58 59@2x

@fedeci do we have an RFC for platform-dependent specs?

My thoughts: let's merge this for now, because it _is_ correct on Linux, then we can use it as a test-bed for platform-dependent specs

SeparateRecords:
Also, good time for platform-dependent generators too. /etc/passwd is probably a good way to get users on linux, on macOS the only reliable way is dscl

fedeci:
> Interesting, this is maybe the first spec where the options are _quite_ different between Linux and macOS. Here's from the Mac man-page:
> 
> 
> 
> CleanShot 2022-10-05 at 17 58 59@2x
> 
> 
> 
> @fedeci do we have an RFC for platform-dependent specs?
> 
> 
> 
> My thoughts: let's merge this for now, because it _is_ correct on Linux, then we can use it as a test-bed for platform-dependent specs

We did a call and throw some ideas on paper a while ago. I prefer to have the engine refactored before implementing new big stuff in it :)

PR #1591, opened 2022-10-06
feat: add spec for pkgutil

### LINKED ISSUE #1542: [pkgutil] missing spec

austinchen12:
Closes #1542 

Not implemented

Kapture 2022-10-05 at 21 18 35

**Additional info:**
* Question: is it possible to show 2 different descriptions depending on if an keyword argument is being passed? Reference to --pkgs subcommand with REGEXP?
* Please add hacktoberfest label

### DIFF
--- a/src/pkgutil.ts
+++ b/src/pkgutil.ts
@@ -1,6 +1,6 @@
 const completionSpec: Fig.Spec = {
   name: "pkgutil",
-  description: "Package manager for macOS",
+  description: "Query and manipulate for macOS Installer packages and receipts",
   subcommands: [
     {
       name: "--packages",
@@ -149,7 +149,7 @@ const completionSpec: Fig.Spec = {
     {
       name: "--bom",
       description:
-        "Extract any BOM files from the flat pkg at path into /tml and return the filename(s)",
+        "Extract any BOM files from the flat pkg at path into /tmp and return the filename(s)",
       args: {
         name: "path",
         description: "The path to the flat package to extract the BOM from",
--- a/src/pkgutil.ts
+++ b/src/pkgutil.ts
@@ -3,23 +3,13 @@ const completionSpec: Fig.Spec = {
   description: "Query and manipulate for macOS Installer packages and receipts",
   subcommands: [
     {
-      name: "--packages",
+      name: ["--packages", "--pkgs"],
       description: "List all installed package IDs on the specified --volume",
     },
     {
       name: "--pkgs-plist",
       description:
         "List all installed package IDs, in Mac OS X plist(5) format, on the specified --volume",
-    },
-    {
-      name: "--pkgs",
-      description:
-        "List all installed package IDs, optionally matching kwarg REGEXP, on the specified --volume",
-      args: {
-        name: "key=value",
-        description: "Regular expression",
-        isOptional: true,
-      },
     },
     {
       name: "--files",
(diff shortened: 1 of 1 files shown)


SeparateRecords:
> is it possible to show 2 different descriptions depending on if an keyword argument is being passed? Reference to --pkgs subcommand with REGEXP?

Kinda! Although I think for flags such as `--pkgs=...` it might not work super well because the `=` tricks it into thinking it's in the value, not the key. Can probably fix this in the generator, but would be best to fix in the tokenizer cc @fedeci 

You'd use it like this:

```ts
import { keyValue } from "@fig/autocomplete-generators";

// then, in your spec ...


  args: {
    name: "whatever your arg name is",
    generators: keyValue({
      keys: [
        // list of keys to suggest
      ],
      values: [
        // list of values to suggest
      ]
    });
  },
```

> Please add hacktoberfest label

Repo already has this, PR should be automatically counted for hacktoberfest!

austinchen12:
> 
seems like values are for suggestions in `keyValue` is for suggestions, I was thinking description. Also, REGEXP is user input, it wouldn't seem like I can hardcode values for those right? Or did I misunderstand what you meant

SeparateRecords:
> seems like values are for suggestions in keyValue is for suggestions, I was thinking description

Thanks for clarifying, I misunderstood. You're right, no way to change the description right now outside of some weird hacks (try writing `rich --padding 4,3,2,1` in your terminal to see one of said hacks)

austinchen12:
> > seems like values are for suggestions in keyValue is for suggestions, I was thinking description
> 
> Thanks for clarifying, I misunderstood. You're right, no way to change the description right now outside of some weird hacks (try writing `rich --padding 4,3,2,1` in your terminal and watch the description to see one of said hacks - [source](https://github.com/withfig/autocomplete/blob/53833c9276a7964e1503e90968e650be788b6c23/src/rich.ts#L463-L496))
@SeparateRecords any blockers?

src/pkgutil.ts
```diff
@@ -0,0 +1,225 @@
...
+        {
+          name: "pkg-path",
+          description: "The path to the flat package to create",
+        },
+      ],
+    },
+    {
+      name: "--bom",
+      description:
+        "Extract any BOM files from the flat pkg at path into /tml and return the filename(s)",
```
SeparateRecords, line 152: Quick typo fix
```suggestion
        "Extract any BOM files from the flat pkg at path into /tmp and return the filename(s)",
```

src/pkgutil.ts
```diff
@@ -0,0 +1,225 @@
...
+    },
+    {
+      name: "--check-signature",
+      description:
+        "Check the validity and trust of the signature on the package at pkg-path",
+      args: {
+        name: "pkg-path",
+        description: "The path to the flat package to check the signature of",
+      },
+    },
```
SeparateRecords, line 176: Same for these, both of their args can have a template

src/pkgutil.ts
```diff
@@ -0,0 +1,225 @@
...
+    {
+      name: "--pkgs",
+      description:
+        "List all installed package IDs, optionally matching kwarg REGEXP, on the specified --volume",
+      args: {
+        name: "key=value",
+        description: "Regular expression",
+        isOptional: true,
+      },
+    },
```
SeparateRecords, line 23: If you add `requiresSeparator: true` to the option, the arg can only be provided with `=` 🙂

SeparateRecords (changes requested):
Thanks so much for working on this!

One more suggestion: let's add a generator for package-id args!

```ts
const generatePackageIds: Fig.Generator = {
  script: "pkgutil --packages",
  splitOn: "\n",
};
```

Then any arg that takes a `package-id`, add `generators: generatePackageIds` to get a list of all available IDs.

SeparateRecords:
Hey @austinchen12, looks like you've marked the feedback as resolved, but haven't pushed any commits 🙂 Have you finished the PR locally?

SeparateRecords (approved):
This is fantastic! Just tried it out locally and it feels great. Thank you so much for working on this 🙂

PR #1592, opened 2022-10-06
feat(trap): add trap spec

### LINKED ISSUE #1531: [trap] missing spec

Fall1ngStar:

New spec

Closes #1531

Add trap spec

**Additional info:**

### DIFF
--- a/src/trap.ts
+++ b/src/trap.ts
@@ -1,0 +1,38 @@
+/*
+ * Generators
+ */
+
+const availableSignalsGenerator = (
+  suggestOptions?: Partial<Fig.Suggestion>
+): Fig.Generator => ({
+  script: "trap -l",
+  postProcess: (output) =>
+    output.split(" ").map((signal) => ({
+      name: signal,
+    })),
+});
+
+const completionSpec: Fig.Spec = {
+  name: "trap",
+  description:
+    "Automatically execute commands after receiving signals by processes or the operating system",
+  options: [
+    {
+      name: ["--list-signals", "-l"],
+      description: "Prints a list of signal names",
+    },
+    {
+      name: ["--print", "-p"],
+      description: "Prints all defined signal handlers",
+    },
+    {
+      name: ["--help", "-h"],
+      description: "Displays help about using this command",
+    },
+  ],
+  args: [
+    { name: "arg", isOptional: true },
+    { name: "reason", generators: availableSignalsGenerator() },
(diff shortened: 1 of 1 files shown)


SeparateRecords (changes requested):
`trap` is one of those fun shell-builtin commands where you can never guarantee that the spec is correct - this works for me because I'm also a fish user, but for zsh and bash it's not right 😭

Here's what I'm thinking: let's use `command kill -l` to list the signals (this is the _exact_ same output as fish's `trap -l`), and omit the options entirely. So it'll be like

```ts
args: [
  { name: "function name" },
  { name: "signals", isVariadic: true, generators: /* generator here */ },
]
```

src/trap.ts
```diff
@@ -0,0 +1,38 @@
+/*
+ * Generators
+ */
+
+const availableSignalsGenerator = (
+  suggestOptions?: Partial<Fig.Suggestion>
+): Fig.Generator => ({
+  script: "trap -l",
+  postProcess: (output) =>
+    output.split(" ").map((signal) => ({
```
SeparateRecords, line 10: Also, I think this might be wrong - there can be a newline in the output. More reliable:
```suggestion
    output.split(/\s+/).map((signal) => ({
```

Fall1ngStar: No longer needed

Fall1ngStar:
Howdy fellow fish user 👋 

Looks like output of `command kill -l` has a different format depending on the shell 🙃 

```
bash -c "command kill -l"
 1) SIGHUP	 2) SIGINT	 3) SIGQUIT	 4) SIGILL	 5) SIGTRAP
 6) SIGABRT	 7) SIGEMT	 8) SIGFPE	 9) SIGKILL	10) SIGBUS
```

```
fish -c "command kill -l"
HUP INT QUIT ILL TRAP ABRT EMT FPE KILL BUS SEGV SYS PIPE ALRM TERM URG
STOP TSTP CONT CHLD TTIN TTOU IO XCPU XFSZ VTALRM PROF WINCH INFO USR1 USR2
```
`zsh` output looks like `bash` and `csh` more like `fish`. I added a regex with an optional group for the number before the signal name, should work on most common shells.

SeparateRecords (approved):
LGTM! Didn't know kill's output differs between shells even when you explicitly use /bin/kill... that's wild. Learned something new 😄

PR #1594, opened 2022-10-06
feat(ping): add Apple-only options

cHaLkdusT:


SeparateRecords:
@fedeci thoughts on merging? These options exist on macOS but obv not linux

fedeci:
> @fedeci thoughts on merging? These options exist on macOS but obv not linux. I'm landing more on the side of not merging until we have OS-specific options but I don't have a good reason for that... it's not like Linux users would try to use an option marked with `--apple` and go "why doesn't this work??"

I still think TODO comments are the way.
@cHaLkdusT could you add this comment to the relevant options? `// TODO(platform): apple only option`

cHaLkdusT:
I have now added // TODO comments to the relevant options

fedeci (approved):
Thanks!

PR #1595, opened 2022-10-06
feat: add `passwd` spec

cHaLkdusT:


SeparateRecords (changes requested):
Looks good! I'd love to add a generator for users. This will only work on macOS at the moment but that's where most Fig users are right now. I plan to make a cross-platform version of this generator later:

```
const generateUsers: Fig.Generator = {
  script: "dscl . -list /Users | grep -E -v '^_'",
  postProcess: (out) => out.trim().map((name) => ({ name, icon: "👤" })),
};
```

cHaLkdusT:
Thanks @SeparateRecords! I have now added the generator for users

fedeci (approved):
Thanks!

PR #1598, opened 2022-10-06
feat: dscacheutil spec

### LINKED ISSUE #1553: [dscacheutil] missing spec

austinchen12:
closes #1553 
dscacheutil spec
Not implemented
Kapture 2022-10-06 at 13 35 41
Kapture 2022-10-06 at 20 25 01

**Additional info:**
1. For the -q subcommand, we can improve -a arg suggestions if we can pass in the value passed to -q. Is this possible right now?
2. There seems to be a bug with the -buckets and -entries options in -cachedump since the name of those options are more than 1 character long (excluding the dash), chaining multiple don't work. See the gif below
Kapture 2022-10-06 at 13 33 30

### DIFF
--- a/src/dscacheutil.ts
+++ b/src/dscacheutil.ts
@@ -10,7 +10,52 @@ const categoryKeysMapping = {
   user: ["name", "uid"],
 };
 const categories = Array.from(Object.keys(categoryKeysMapping));
-const associatedKeys = Array.from(Object.values(categoryKeysMapping));
+
+interface PostProcessQuery {
+  attributeKey: string;
+}
+
+const postProcessQuery =
+  (options: PostProcessQuery): Fig.Generator["postProcess"] =>
+  (out) => {
+    const { attributeKey } = options;
+    const values = new Set<string>();
+    out.split("\n\n").map((entry) => {
+      entry.split("\n").forEach((line) => {
+        const [key, value] = line.trim().split(": ");
+        if (key === attributeKey) {
+          values.add(value);
+        }
+      });
+    });
+
+    const suggestions = [];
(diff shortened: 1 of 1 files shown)


austinchen12:
In the second gif, for some reason after typing `dscacheutil -q user -a ` the suggestions come up when you type the `-a` but goes away right when you press space. If someone can take a look and see if I'm doing anything wrong, it seems like adding arguments to `-a` should make it work properly

src/dscacheutil.ts
```diff
@@ -0,0 +1,128 @@
...
+        tokens
+      );
+    },
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "dscacheutil",
+  description: "Utility for managing the Directory Service cache",
+  subcommands: [
```
SeparateRecords, line 63: Single-letter options are behaving strangely because of the chaining rules, if you make this change it should be a lot more intuitive!
```suggestion
const completionSpec: Fig.Spec = {
  name: "dscacheutil",
  description: "Utility for managing the Directory Service cache",
  parserDirectives: {
    flagsArePosixNoncompliant: true,
  },
  subcommands: [
```

SeparateRecords (changes requested):

src/dscacheutil.ts
austinchen12: Including this change seems to make the `dscacheutil -q user -a uid` command break. The `dscacheutilGenerators.keys` doesn't seem to be called

austinchen12: @SeparateRecords ^

SeparateRecords (approved):
LGTM! Thank you @austinchen12 😄

Issue #1547, opened 2022-09-30
[gum] missing spec


SeparateRecords:

Adedunmol:
Is anyone working on this? I would like to work on this.

SeparateRecords:
This one just got completed, but there's lots more still to get through!

PR #1599, opened 2022-10-07
feat(gum): add gum spec

Fall1ngStar:
New spec for `gum`

Missing spec for `gum` (#1547)

Add spec for `gum`

**Additional info:**

### DIFF
--- a/src/gum.ts
+++ b/src/gum.ts
@@ -1,8 +1,16 @@
 const alignValues = ["left", "center", "right", "bottom", "middle", "top"];
 
 const styleOptions: Fig.Option[] = [
-  { name: "--background", description: "Background Color", args: {} },
-  { name: "--foreground", description: "Foreground Color", args: {} },
+  {
+    name: "--background",
+    description: "Background Color",
+    args: { name: "color" },
+  },
+  {
+    name: "--foreground",
+    description: "Foreground Color",
+    args: { name: "color" },
+  },
   {
     name: "--border",
     description: "Border Style",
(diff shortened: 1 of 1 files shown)


src/gum.ts
```diff
@@ -0,0 +1,387 @@
...
+];
+
+const styleOptionGenerator = (element: string) =>
+  styleOptions.map(
+    (style): Fig.Option => ({
+      ...style,
+      name: (style.name as string).replace("--", `--${element}.`),
+      description: `${style.description} for ${element}`,
+    })
+  );
```
SeparateRecords, line 46: This is really smart. Nice work 🤩

src/gum.ts
```diff
@@ -0,0 +1,387 @@
...
+      args: {
+        name: "options",
+        isVariadic: true,
+        description: "Options to choose from",
+      },
+      options: [
+        {
+          name: "--height",
+          description: "Height of the list",
+          args: {},
```
SeparateRecords, line 64: Could you add a `name` property to the arg object? Same comment for all the args below. If no obvious name for the value jumps out, don't worry about it, but for example with this you could do `{ name: "lines" }`

src/gum.ts
```diff
@@ -0,0 +1,387 @@
...
+    {
+      name: "confirm",
+      description: "Ask a user to confirm an action",
+      args: {
+        name: "prompt",
+        description: "Prompt to display",
+      },
+      options: [
+        {
+          name: "--afirmative",
```
SeparateRecords, line 116: Quick typo fix
```suggestion
          name: "--affirmative",
```

src/gum.ts
```diff
@@ -0,0 +1,387 @@
...
+  ],
+  options: [
+    {
+      name: ["--help", "-h"],
+      description: "Show context-sensitive help",
+    },
+    {
+      name: ["--version", "-v"],
+      description: "Print the version number",
+    },
```
SeparateRecords, line 382: These options are available on all child commands, you can actually add `isPersistent: true` to make them inherited!

src/gum.ts
```diff
@@ -0,0 +1,387 @@
...
+  ],
+  // Only uncomment if gum takes an argument
+  // args: {}
```
SeparateRecords, line 385: Can be deleted

SeparateRecords (changes requested):
Awesome. Thank you so much for working on this! Just a couple of (hopefully) easy changes then it's ready to merge 🙂

src/gum.ts
Fall1ngStar: 😊 thank !

src/gum.ts
Fall1ngStar: Done, hope the naming I gave is matching the Fig style

src/gum.ts
Fall1ngStar: Fixed

src/gum.ts
Fall1ngStar: Fixed

SeparateRecords (approved):
Amazing, thank you so much @Fall1ngStar!

PR #1602, opened 2022-10-08
feat(code): add missing flags

mrienstra:
Motivated by missing `-` option.

See also:

https://github.com/microsoft/vscode/blob/main/src/vs/platform/environment/node/argv.ts
Complete.

https://github.com/microsoft/vscode/blob/main/cli/src/commands/args.rs
Missing `profile`, maybe more.

https://github.com/microsoft/vscode/blob/main/resources/completions/zsh/_code
Missing `profile`, maybe more.

_(I'll open a PR on `microsoft/vscode` about those later)_

```
>code -v
1.71.2
```

### DIFF
--- a/src/code.ts
+++ b/src/code.ts
@@ -1,7 +1,7 @@
 const commonOptions: Fig.Option[] = [
   {
     name: ["-"],
-    description: "Read from stdin",
+    description: "Read from stdin (e.g. 'ps aux | grep code | code -')",
   },
   {
     name: ["-d", "--diff"],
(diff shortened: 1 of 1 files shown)


mrienstra:
Whoops, missed the "conventional commits" requirement, my bad. Should I rebase & force push?

mrienstra:
I tested all new and changed options using `npm run dev`, they all looked fine.

fedeci:
@mrienstra do not worry about conventional commits when you work on your PR. We squash the commits so that part of history gets "lost" (it gets lost from the git history only. It remains in the github PRs history). 
The only thing that should follow semantic commits is the PR title, cause that's the commit message that we use when merging the PR.

src/code.ts
```diff
@@ -1,4 +1,8 @@
 const commonOptions: Fig.Option[] = [
+  {
+    name: ["-"],
```
fedeci, line 3:
```suggestion
    name: "-",
```

fedeci (changes requested):
Thanks!

fedeci (approved):
Thanks again!

mrienstra:
Thanks for your help!

PR #1603, opened 2022-10-08
feat: Add spec for `seq`

### LINKED ISSUE #1464: Add spec for `seq`
### Sanity checks

seq

### Issue Details

How do we not have a spec for `seq` !

Cool-Runningz:
- New spec

- Resolves #1464 

- Add new `seq` spec

**Additional info:**
- [seq man page](https://man.openbsd.org/seq.1)
- `seq` doesn't take any subcommands and only takes in arguments and flags. If no flags are passed in, the only required argument is the `first` number. After that, it will optionally accept either a `step` or `last` argument and those arguments differ depending on whether two or all three arguments are passed in. 
   - The order of operations with flags and arguments affects the output and I was a little unsure of how to handle all the permutations so just wanted to mention that ahead of time in case there are any improvements that you think I need to make around that area.

### DIFF
--- a/src/seq.ts
+++ b/src/seq.ts
@@ -1,0 +1,46 @@
+const completionSpec: Fig.Spec = {
+  name: "seq",
+  description: "Print sequences of numbers. (Defaults to increments of 1)",
+  args: [
+    {
+      name: "first",
+      description: "Starting number in sequence",
+    },
+    {
+      name: "step",
+      description: "Increment interval",
+      isOptional: true,
+    },
+    {
+      name: "last",
+      description: "Last number in sequence",
+      isOptional: true,
+    },
+  ],
+  options: [
+    {
+      name: "-w",
+      description:
+        "Equalize the widths of all numbers by padding with zeros as necessary",
+    },
+    {
+      name: "-s",
+      description: "String separator between numbers. Default is newline",
+      insertValue: `-s "{cursor}"`,
+      args: {
+        name: "string",
+        description: "Separator",
+      },
+    },
+    {
+      name: "-f",
+      description: "Use a printf(3) style format to print each number",
+      insertValue: `-f %{cursor}`,
+      args: {
+        name: "format",
+        description: "Print all numbers using format",
+      },
+    },
+  ],
+};
+export default completionSpec;


src/seq.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+    {
+      name: "-f",
+      description: "Use a printf(3) style format to print each number",
+      insertValue: `-f %{cursor}`,
+      args: {
+        name: "format",
+        description: "Print all numbers using format",
+      },
+    },
```
fedeci, line 43: Can you add the `--help` and `--version` options please?

src/seq.ts
```diff
@@ -0,0 +1,46 @@
...
+      name: "-s",
+      description: "String separator between numbers. Default is newline",
+      insertValue: `-s "{cursor}"`,
+      args: {
+        name: "string",
+        description: "Separator",
+      },
+    },
+    {
+      name: "-f",
```
fedeci, line 36:
```suggestion
      name: ["-f", "--format"],
```

src/seq.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+    {
+      name: "last",
+      description: "Last number in sequence",
+      isOptional: true,
+    },
+  ],
+  options: [
+    {
+      name: "-w",
```
fedeci, line 22:
```suggestion
      name: ["-w", "--fixed-width"],
```

src/seq.ts
```diff
@@ -0,0 +1,46 @@
...
+      description:
+        "Equalize the widths of all numbers by padding with zeros as necessary",
+    },
+    {
+      name: "-s",
```
fedeci, line 27:
```suggestion
      name: ["-s", "--separator"],
```

fedeci (changes requested):
Thanks for the PR @Cool-Runningz!
This is the man page from the same macos tool. I think it is an extended version and as you can see there is extra option there (`--terminator`).
```
NAME
     seq – print sequences of numbers

SYNOPSIS
     seq [-w] [-f format] [-s string] [-t string] [first [incr]] last

DESCRIPTION
     The seq utility prints a sequence of numbers, one per line (default), from first (default 1), to near
     last as possible, in increments of incr (default 1).  When first is larger than last, the default incr
     is -1.

     All numbers are interpreted as floating point.

     Normally integer values are printed as decimal integers.

     The seq utility accepts the following options:

     -f format, --format format
                   Use a printf(3) style format to print each number.  Only the A, a, E, e, F, f, G, g, and
                   % conversion characters are valid, along with any optional flags and an optional numeric
                   minimum field width or precision.  The format can contain character escape sequences in
                   backslash notation as defined in ANSI X3.159-1989 (“ANSI C89”).  The default is %g.

     -s string, --separator string
                   Use string to separate numbers.  The string can contain character escape sequences in
                   backslash notation as defined in ANSI X3.159-1989 (“ANSI C89”).  The default is \n.

     -t string, --terminator string
                   Use string to terminate sequence of numbers.  The string can contain character escape
                   sequences in backslash notation as defined in ANSI X3.159-1989 (“ANSI C89”).  This option
                   is useful when the default separator does not contain a \n.

     -w, --fixed-width
                   Equalize the widths of all numbers by padding with zeros as necessary.  This option has
                   no effect with the -f option.  If any sequence numbers will be printed in exponential
                   notation, the default conversion is changed to %e.
```

Can you add it along with this comment on top of it `TODO(platform): macos only option`?

src/seq.ts
Cool-Runningz: Apologies for not adding the `--*` flags. When I tried using them on my Mac it didn't seem to work which is why I had initially left them off. I will make sure to add them in for all the ones that you mentioned.

src/seq.ts
Cool-Runningz: Same here as my comment below, I couldn't get the `--help` or `--version` options to work on my Mac without throwing an error but I will add those options in.

src/seq.ts
SeparateRecords: @Cool-Runningz this might be because of your macOS version? Not sure why that would be otherwise

src/seq.ts
SeparateRecords: @fedeci on macOS this CLI doesn't have --help and --version

src/seq.ts
Cool-Runningz: Yea I think you might be right because I tested it on my work computer, which is a newer Mac, and it seems to be working there. `--fixed-width` is the only one I couldn't get to work but it does seem to be machine related. Sorry for the false alarm.

src/seq.ts
Cool-Runningz: If it turns out `--help` and `--version` are not supported then I can push up a commit to remove those options since I just added them

fedeci (approved):
Thanks!

Issue #1285, opened 2022-06-13
[yarn] version missing commands


ShawnCockburn:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [X] I would like to work on this.

### What CLI tool does this relate to?

yarn

### Which statement makes the most sense?

This completion spec isn't complete

### Issue Details

(Note to self, I will fix this: @ShawnCockburn )

yarn version missing auto complete commands

also check for other missing yarn commands

fedeci:
Since yarn 1 is very different from yarn 2 and 3

ShawnCockburn:
> Actually we should start using diff-versioned specs for yarn

I was referring to version yarn `1.22.15`.

I just had a look at the documentation on the website and I couldn't find much info about diff-versioned specs and how to use/make them, could you point me in the right direction?

Thank you mate!

ShawnCockburn:
@fedeci @SeparateRecords ^^^ bump ^^^

fedeci:
Thanks for pinging me (always ping me as I tend to be overwhelmed by notifications and often forget something)!
I believe we need to document diff-versioned specs, for the moment just update the current spec which is actually yarn@1 with some commands from later versions!

ShawnCockburn:
will do, thanks!

joeheyming:
I can take this, but I'm not sure how many arguments need to be added to the spec

joeheyming:
https://github.com/withfig/autocomplete/pull/1606

brendanfalk:
This was fixed with the above merges. Closing
<img width="874" alt="CleanShot 2022-11-23 at 10 45 10@2x" src="https://user-images.githubusercontent.com/4949076/203624406-650c7469-2ade-4f26-a076-73348fb29cdf.png">

PR #1606, opened 2022-10-09
feat: update yarn completion spec

joeheyming:
This addresses the changes in https://github.com/withfig/autocomplete/issues/1285
the yarn command has more options for the `yarn version` command so here I did:
`yarn version -h` then added all the options that were present that had to do with version.

The current behavior is that it only shows `--major/minor/patch`

Now we show in addition to major/minor/patch:
message, premajor, preminor, prepatch, prerelease, preid, no-git-tag-version, no-commit-hooks, access, tag
Also, I added commonOptions, because I saw those were present as well with the `yarn version -h` command

### DIFF
--- a/src/yarn.ts
+++ b/src/yarn.ts
@@ -1328,6 +1328,13 @@ const completionSpec: Fig.Spec = {
       name: "version",
       description: "Update version of your package",
       options: [
+        ...commonOptions,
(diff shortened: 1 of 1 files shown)


SeparateRecords (approved):
Looks good to me, thanks @joeheyming!!

SeparateRecords:
-_- lint failure isn't your fault, merging anyway

PR #1613, opened 2022-10-11
fix: suggest current token for grep argument

mschrage:

**Additional info:**


SeparateRecords (approved):
Great! We should add this to ripgrep too actually

PR #1617, opened 2022-10-11
feat(nuxi): improvements & updates

cpreston321:

Following new changes to commands here: https://v3.nuxtjs.org/api/commands/add

This PR includes updated to more options that were added to each argument. 

## New Command 

- Added `cleanup` command

## Command Options 

- Added `--dotenv` option to
  - `preview` command
  - `build` command
  - `dev` command
- Added `--prerender` option
  - `build` command  
- Added `--no-clear` option
  - `dev`
- Added `--force`, `--prefer-offline`, `--shell`
  - `init|create` command 
  
 ## Command Icons
 
- Modified some of the `Icons` that made more since for the command itself

> **Note**
> The current behavior is somewhat outdated since the CLI is evolving over time.

No new behavior just one more command some more options added.

**Additional info:**

## Question 

I want to add `modifier` flag options as if you type `npx nuxi add component Name --mode server` it will generate a new file called `Name.server.vue` but I want to implement within this PR but I tried looking for some docs but I couldn't find anything that I want to do. Maybe it doesn't exists. 

I need to trigger a flag when `add component|plugin` then display flag the other options I don't need to show the user in fig autocomplete.

Examples: https://v3.nuxtjs.org/api/commands/add#nuxi-add-plugin, https://v3.nuxtjs.org/api/commands/add#nuxi-add-api

I would love feedback about what I described above to see if that is possible!

Thanks!

### DIFF
--- a/src/nuxi.ts
+++ b/src/nuxi.ts
@@ -52,6 +52,18 @@ const mainNuxiSubcommands: Fig.Subcommand[] = [
         name: ["--template", "-t"],
         description: "Specify a Git repository to use as a template",
       },
+      {
+        name: "--force",
+        description: "Force clone to any existing directory",
+      },
+      {
+        name: "--prefer-offline",
+        description: "Try local cache first to download templates",
+      },
+      {
+        name: "--shell",
+        description: "Open shell in cloned directory (experimental)",
+      },
     ],
   },
   {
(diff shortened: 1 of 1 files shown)


SeparateRecords (approved):
LGTM! Thank you @cpreston321 🙂

PR #1621, opened 2022-10-12
feat: Add `pytest` spec

sri-go:
Add spec file for pytest cli

Additional info:
The [pytest](https://docs.pytest.org/en/7.1.x/) framework makes it easy to write small, readable tests, and can scale to support complex functional testing for applications and libraries.

### DIFF
--- a/src/pytest.ts
+++ b/src/pytest.ts
@@ -63,7 +63,7 @@ const completionSpec: Fig.Spec = {
       name: "--code-highlight",
       description:
         "Whether code should be highlighted (only if --color is also enabled)",
-      dependsOn: "--color",
+      dependsOn: ["--color"],
       args: {
         name: "Highlight",
         suggestions: ["yes", "no"],
--- a/src/pytest.ts
+++ b/src/pytest.ts
@@ -214,6 +214,7 @@ const completionSpec: Fig.Spec = {
       args: {
         name: "Mode",
         default: "prepend",
+        suggestions: ["prepend", "append", "importlib"],
       },
     },
     {
--- a/src/pytest.ts
+++ b/src/pytest.ts
@@ -10,6 +10,14 @@ const completionSpec: Fig.Spec = {
   },
   options: [
     {
+      name: "-c",
+      description:
+        "Load configuration from `file` instead of trying to locate one of the implicit configuration files",
+      args: {
+        name: "File",
+      },
+    },
+    {
       name: "--cache-clear",
       description: "Remove all cache contents at start of test run",
     },
@@ -41,12 +49,32 @@ const completionSpec: Fig.Spec = {
       },
     },
     {
+      name: ["--co", "--collect-only"],
+      description: "Only collect tests, don't execute them",
+    },
+    {
+      name: "--collect-in-virtualenv",
+      description: "Don't ignore tests in a local virtualenv directory",
+    },
+    {
       name: "--color",
       description: "Color terminal output",
       args: {
         name: "Color",
         description: "(yes/no/auto)",
       },
+    },
+    {
+      name: "--confcutdir",
+      description: "Only load conftest.py's relative to specified dir",
+      args: {
+        name: "Dir",
+      },
+    },
+
+    {
+      name: "--continue-on-collection-errors",
+      description: "Force test execution even if collection errors occur",
     },
     {
       name: "--durations",
@@ -65,8 +93,44 @@ const completionSpec: Fig.Spec = {
       },
     },
     {
+      name: "--deselect",
+      description:
+        "Deselect item (via node id prefix) during collection (multi-allowed)",
+      args: {
+        name: "nodeid_prefix",
+      },
+    },
+    {
       name: ["--disable-warnings", "--disable-pytest-warnings"],
       description: "Disable warnings summary",
+    },
+    {
+      name: "--doctest-continue-on-failure",
+      description:
+        "For a given doctest, continue to run after the first failure",
+    },
+    {
+      name: "--doctest-ignore-import-errors",
+      description: "Ignore doctest ImportErrors",
+    },
+    {
+      name: "--doctest-modules",
+      description: "Run doctests in all .py modules",
+    },
+    {
+      name: "--doctest-report",
+      description: "Choose another output format for diffs on doctest failure",
+      args: {
+        name: "Output format",
+        description: "None,cdiff,ndiff,udiff,only_first_failure",
+      },
+    },
+    {
+      name: "--doctest-glob",
+      description: "Doctests file matching pattern, default: test*.txt",
+      args: {
+        name: "Pattern",
+      },
     },
     {
       name: ["--exitfirst", "-x"],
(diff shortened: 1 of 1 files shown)


src/pytest.ts
```diff
@@ -0,0 +1,540 @@
+const completionSpec: Fig.Spec = {
+  name: "pytest",
+  args: {
+    name: "File or Directory",
+    description:
+      "The test file or directory you want to run pytest on. If omitted, pytest will run all \
+      of the files of the form test_*.py or *_test.py in the current directory \
+      and its subdirectories",
+    isOptional: true,
+  },
```
SeparateRecords, line 10: Let's add `template: ["filepaths", "folders"]` so this suggests files and folders!

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+  options: [
+    {
+      name: "--assert",
+      description:
+        "Control assertion debugging tools. \
+                    'plain' performs no assertion debugging. \
+                    'rewrite' (the default) rewrites assert statements in \
+                    test modules on import to provide assert expression information",
```
SeparateRecords, line 18: This could probably be a single line description, it's more readable _as code_ over multiple lines but I'd favor simplicity instead of line continuations.

Plus for tools that don't directly render the description to DOM, such as https://fig.io/manual or https://shell.how, this could cause problems:
CleanShot 2022-10-12 at 14 13 47@2x

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+      args: {
+        name: "Mode",
+      },
```
SeparateRecords, line 21: There are known modes (plain, rewrite), let's add `suggestions: ["plain", "rewrite"]` so it suggests them!

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+        name: "Directory",
+      },
+    },
+    {
+      name: "-c",
+      description:
+        "Load configuration from `file` instead of trying to locate one of the implicit configuration files",
+      args: {
+        name: "File",
+      },
```
SeparateRecords, line 37: Filepaths template to suggest files

```suggestion
      args: {
        name: "File",
        template: "filepaths",
      },
```

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+    },
+    {
+      name: "--cache-clear",
+      description: "Remove all cache contents at start of test run",
+    },
+    {
+      name: "--cache-show",
+      description: "",
```
SeparateRecords, line 45: This option is missing its description

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+      args: {
+        name: "Glob",
+        description: "",
```
SeparateRecords, line 48: Can probably be omitted

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+        default: "*",
+      },
+    },
+    {
+      name: "--capture",
+      description: "Per-test capturing method",
+      args: {
+        name: "Method",
+        description: "One of fd|sys|no|tee-sys",
+      },
```
SeparateRecords, line 59: Same here, can you add `suggestions`?

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+    },
+    {
+      name: "--code-highlight",
+      description:
+        "Whether code should be highlighted (only if --color is also enabled)",
+      args: {
+        name: "",
```
SeparateRecords, line 66: Was this supposed to have a name? Empty name can be omitted, but I think it's clearer if it has one

src/pytest.ts
```diff
@@ -0,0 +1,540 @@
...
+        description: "{yes,no}",
```
SeparateRecords, line 67: Same thing as above, let's do `suggestions`!

SeparateRecords (changes requested):
Thanks for the PR @sri-go!

Left some feedback, stopped part way through the file (but I did read all of it). The feedback applies to the whole thing.

Hopefully it's simple changes! Summarizing here:
1. Add `template` property to args that take a file or folder
2. Replace multiline descriptions with single lines to remove possible whitespace issues
3. For args with known values (eg. yes/no), add `suggestions` property to show them!

You've done a great job on this spec. Thank you so much for working on it!!

src/pytest.ts
SeparateRecords: Oh also, for this option, you might want to add `dependsOn: ["--color"]` (I'm guessing based on the description). It'll have its priority boosted if `--color` is provided

src/pytest.ts
```diff
@@ -0,0 +1,539 @@
...
+        template: "filepaths"
+      },
+    },
+    {
+      name: "--cache-clear",
+      description: "Remove all cache contents at start of test run",
+    },
+    {
+      name: "--cache-show",
+      description: "Show cache contents, don't perform collection or tests. Optional argument: glob (default: '*').",
```
fedeci, line 43:
```suggestion
      description: "Show cache contents, don't perform collection or tests. Optional argument: glob (default: '*')",
```

src/pytest.ts
```diff
@@ -0,0 +1,539 @@
...
+        "Exit on test failure and continue from last failing test next time",
+    },
+    {
+      name: ["--stepwise-skip", "--sw-skip"],
+      description:
+        "Ignore the first failing test but stop on the next failing test",
+    },
+    {
+      name: "--strict",
+      description: "(deprecated) alias to --strict-markers",
```
fedeci, line 484: This can use the deprecated property: https://fig.io/docs/reference/option#deprecated

src/pytest.ts
```diff
@@ -0,0 +1,539 @@
...
+      name: ["--quiet", "-q"],
+      description: "Decrease verbosity",
+    },
+    {
+      name: "-r",
+      description:
+        "Show extra test summary info as specified by chars: (f)ailed, (E)rror, (s)kipped, (x)failed, (X)passed, (p)assed, (P)assed with output, (a)ll except passed (p/P), or (A)ll. (w)arnings are enabled by default (see --disable-warnings), 'N' can be used to reset the list. (default: 'fE')",
+      args: {
+        name: "chars",
+        suggestions: ["a", "A", "E", "f", "N", "p", "P", "s", "w", "x", "X"],
```
fedeci, line 427: This needs a custom list generator. I am writing it rn. I will update it when it is available!

src/pytest.ts
```diff
@@ -0,0 +1,539 @@
...
+      args: {
+        name: "Log Format",
+      },
+    },
+    {
+      name: "--log-file",
+      description: "Path to a file where logging will be written to",
+      args: {
+        name: "Log File Path",
+        template: "filepaths"
```
fedeci, line 309:
```suggestion
        template: "filepaths",
        suggestCurrentToken: true,
```

src/pytest.ts
```diff
@@ -0,0 +1,539 @@
...
+      name: "--trace",
+      description: "Immediately break when running each test",
+    },
+    {
+      name: " --trace-config",
+      description: "Trace considerations of conftest.py files",
+    },
+    {
+      name: ["--verbose", "-v"],
+      description: "Increase verbosity",
```
fedeci, line 514:
```suggestion
      description: "Increase verbosity",
      isRepeatable: true,
```

fedeci: I use this all the time :)

fedeci (changes requested):
Thanks for this PR! I left some comments to solve before merging too.

SeparateRecords (changes requested):
Awesome, this is looking really good. Thanks again for taking this on, pytest has a _huge_ number of options, this is getting super close to being ready to merge!!

SeparateRecords (approved):
Fantastic! No complaints 🙂 Thank you so much for all the effort @sri-go, this will be hugely appreciated by every python user!!

PR #1622, opened 2022-10-12
chore: add dependabot for npm and github-actions

lsaudon:

Add dependabot for automates dependency updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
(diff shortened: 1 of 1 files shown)


SeparateRecords:
@fedeci I don't have any particular disagreement with adding dependabot. Doesn't strike me as _necessary_ per se, but it couldn't hurt to have our dependencies kept up to date automatically. Thoughts?

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
```
fedeci, line 10: Having it daily is quite annoying when you have a lot of deps.
```suggestion
      interval: 'weekly'
```

fedeci:
@SeparateRecords I agree with you. We just need to check the changelogs when we bump the version of packages. It happens that even patches break stuff.

SeparateRecords:
Sounds good. Will test changes locally before merging too, can check for build errors!

SeparateRecords (approved):
Thanks @lsaudon !

PR #1647, opened 2022-10-14
feat: add spec for `mdls`

### LINKED ISSUE #1548: [mdls] missing spec

Cool-Runningz:
- New spec

- Resolves #1548 

- Add new `mdls` spec

**Additional info:**
```
MDLS(1)                   BSD General Commands Manual                  MDLS(1)
NAME
     mdls -- lists the metadata attributes for the specified file

SYNOPSIS
     mdls [-name attributeName] [-raw [-nullMarker markerString]] file ...

DESCRIPTION
     The mdls command prints the values of all the metadata attributes associated with the files pro-
     vided as an argument.

     The following options are available:

     -name        Print only the matching metadata attribute value.  Can be used multiple times.

     -raw         Print raw attribute data in the order that was requested.  Fields will be separated
                  with a ASCII NUL character, suitable for piping to xargs(1) -0.

     -nullMarker  Sets a marker string to be used when a requested attribute is null.  Only used in
                  -raw mode.  Default is "(null)".
```

### DIFF
--- a/src/mdls.ts
+++ b/src/mdls.ts
@@ -1,6 +1,9 @@
 const completionSpec: Fig.Spec = {
   name: "mdls",
   description: "Lists the metadata attributes for the specified file",
+  parserDirectives: {
+    flagsArePosixNoncompliant: true,
+  },
   args: {
     name: "file",
     template: "filepaths",
--- a/src/mdls.ts
+++ b/src/mdls.ts
@@ -64,6 +64,7 @@ const completionSpec: Fig.Spec = {
       insertValue: `--nullMarker "{cursor}"`,
       dependsOn: ["--raw", "-raw"],
     },
+    // TODO(platform): Add --plist macos only option
   ],
 };
 export default completionSpec;
--- a/src/mdls.ts
+++ b/src/mdls.ts
@@ -1,0 +1,66 @@
+const completionSpec: Fig.Spec = {
+  name: "mdls",
+  description: "Lists the metadata attributes for the specified file",
+  args: {
+    name: "file",
+    template: "filepaths",
+    isVariadic: true,
+  },
+  options: [
+    {
+      name: ["--name", "-name"],
+      description:
+        "Print only the matching metadata attribute value.  Can be used multiple times",
+      isRepeatable: true,
+      args: {
+        name: "attributeName",
+        description: "Metadata attribute name",
+        suggestions: [
+          "_kMDItemDisplayNameWithExtensions",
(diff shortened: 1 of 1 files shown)


Cool-Runningz:
`mdls` was unique in the sense that it doesn't require the `--` for the verbose flags. This means both `--name` and `-name` will produce the same output. That being said, I noticed that the `suggestions` would not populate if the double dashes weren't used, i.e. (`-name`), which is why I added it as the second option in the `name` arrays. Not sure if you want to include the `-*` flags but I wanted to mention this since it was a bit unorthodox.

SeparateRecords (dismissed):
This is perfect! You handled the `--name/-name` thing exactly how I would've too.

One thing to change, because of the options starting with a single dash, you'll need to add `parserDirectives` to the top-level spec:
```ts
  parserDirectives: {
    flagsArePosixNoncompliant: true,
  },
```

This is just because of some internal parser stuff. Single dashes make the parser think you're doing short-option chaining.

Once that's done, ready to merge!!

SeparateRecords:
Also wanted to find out, are you in our discord? Will give you contributor role. Thank you so much for the contributions!

Cool-Runningz:
@SeparateRecords I joined the Discord group a few days ago so that's perfect timing. (Discord info: **alyssa_codes#5441**)
Thanks for offering to give me a contributor role and I hope to try and contribute more specs in the future. 

Also thanks for taking the time to review the PR's so promptly, it's much appreciated and it's great to have the opportunity to contribute during Hacktotberfest 🙂

fedeci (approved):
The mac version also supports the `--plist` flag. Can you leave a TODO comment in the code to remind to add it once we have spec variants feature available?

SeparateRecords:
AFAIK `mdls` is a Mac-specific tool... for some reason the man-page doesn't list `--plist`. Weird

Cool-Runningz:
I also thought it was odd that the man page didn't list out `--plist` as an option. The only way I was able to get some kind of information on it was by running the command incorrectly 😅  and it gave me the following output:
```
mdls: option `--plist' requires an argument

usage: mdls [-name attr][-sdb] [-raw [-nullMarker markerString]] [-plist file] path
list the values of one or all the attributes of the specified file
  -plist:       output attributes in XML format to file. Use - to write to stdout
                option -plist is incompatible with options -raw, -nullMarker, and -name
```

Based on that, I have updated the spec to now include the `--plist` option.

Issue #1541, opened 2022-09-30
[plutil] missing spec


clo4:
Property-list (plist) utility

This will require `parserDirectives.flagsArePosixNoncompliant` because the options are wack

PR #1652, opened 2022-10-15
feat(plutil): Added completion spec for plutil

duncanlutz:
Feature

No completion spec for plutil currently exists ([issue that requested this feature](https://github.com/withfig/autocomplete/issues/1541))

Added completion spec for plutil

### DIFF
--- a/src/plutil.ts
+++ b/src/plutil.ts
@@ -1,0 +1,81 @@
+const completionSpec: Fig.Spec = {
+  name: "plutil",
+  description:
+    "Check the syntax of property list files, or convert a plist file from one format to another. Specifying - as an input file reads from stdin",
+  parserDirectives: {
+    flagsArePosixNoncompliant: true,
+    optionsMustPrecedeArguments: true,
+  },
+  args: {
+    name: "file",
+    description: "File to check the syntax of or convert",
+  },
+  options: [
+    {
+      name: "-help",
+      description: "Show the usage information for the command and exit",
+    },
+    {
+      name: "-p",
+      description: "Print the property list in a human-readable fashion",
+    },
+    {
+      name: "-lint",
+      description:
+        "Check the named property list files for syntax errors. This is the default command option if none is specified",
+    },
+    {
+      name: "-convert",
+      description:
+        "Convert the named file to the indicated format and write back to the file system",
+      args: {
+        name: "format",
+        description: "Format to convert the named file to",
+        suggestions: [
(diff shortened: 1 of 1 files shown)


src/plutil.ts
```diff
@@ -0,0 +1,81 @@
...
+  name: "plutil",
+  description:
+    "Check the syntax of property list files, or convert a plist file from one format to another. Specifying - as an input file reads from stdin",
+  parserDirectives: {
+    flagsArePosixNoncompliant: true,
+    optionsMustPrecedeArguments: true,
+  },
+  args: {
+    name: "file",
+    description: "File to check the syntax of or convert",
```
fedeci, line 11:
```suggestion
    description: "File to check the syntax of or convert",
    template: "filepaths",
```

src/plutil.ts
```diff
@@ -0,0 +1,81 @@
...
+      name: "-r",
+      description:
+        "For JSON, add whitespace and indentation to make the output more human-readable",
+    },
+    {
+      name: "-o",
+      description:
+        "Specify an alternate path name for the result of the -convert operation; this option is only useful with a single file to be converted. Specifying - as the path outputs to stdout",
+      args: {
+        name: "path",
```
fedeci, line 68:
```suggestion
        name: "path",
        template: "filepaths",
        suggestCurrentToken: true,
```

fedeci (changes requested):
Thanks! Just two comments about some improvements that can be done to the spec!

SeparateRecords (changes requested):
Thanks @duncanlutz! Got some comments on the general structure

Some of `plutil`'s options aren't _really_ options. It takes some "command options" which we can model as subcommands.

Some subcommands are `-convert`, `-extract`, etc.

This also allows us to model the behavior more closely too

```ts
  subcommands: [
    {
      name: "-extract",
      icon: "fig://icon?type=option", // makes it look like an option
      args: [
        { name: "keypath" },
        { name: "format", suggestions: ["xml1", "binary1", "json", "swift", "objc", "raw"] },
        { name: "file", template: "filepaths", suggestions: [{ name: "-", hidden: true }] },
      ],
      options: [
        {
          name: "-expect",
          args: {
            name: "type",
            suggestions: ["-bool", "-integer", "-float", /* ... */ ],
          },
        },
      ],
    }
  ]
```

mschrage (approved):

PR #1670, opened 2022-10-23
feat(aws): add `--profile` option support

### LINKED ISSUE #542: [aws] - Suggest AWS profiles
### Sanity checks

aws

### Issue Details

All AWS commands can be suffixed with `--profile PROFILE`, in which `PROFILE` is one of the configured profile and can be fetched using `aws configure list-profiles`.

I suggest Fig autocompletes the profile list on **all** aws commands

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

joeheyming:
In this commit, we want to make it so aws --profile auto completes
with the list of profiles you have available.

Currently the aws command doesn't know about --profiles

Generally these profiles live inside a credentials file
but instead of directly looking at the credentials file,
we ask aws to give us a list of the profiles using
`aws configure list-profiles`

**Additional info:**

closes #542

### DIFF
--- a/src/aws.ts
+++ b/src/aws.ts
@@ -1,3 +1,32 @@
+export const awsProfileGenerator: Fig.Generator = {
+  cache: {
+    strategy: "stale-while-revalidate",
+    cacheByDirectory: true,
+  },
+  script: "aws configure list-profiles",
+  postProcess: function (out, [awsClient]) {
+    if (out.trim() == "") {
+      return [];
+    }
+
+    try {
+      const lines = out.split("\n");
+
+      if (lines) {
+        return lines.map((line) => {
+          return {
+            name: line,
+          };
+        });
+      }
+    } catch (e) {
+      console.error(e);
+    }
+
+    return [];
+  },
+};
+
 const completionSpec: Fig.Spec = {
   name: "aws",
   async generateSpec(_, executeShellCommand) {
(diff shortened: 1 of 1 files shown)


src/aws.ts
```diff
@@ -1,3 +1,32 @@
+export const awsProfileGenerator: Fig.Generator = {
+  cache: {
+    strategy: "stale-while-revalidate",
+    cacheByDirectory: true,
+  },
+  script: "aws configure list-profiles",
+  postProcess: function (out, [awsClient]) {
```
fedeci, line 7: nit:
```suggestion
  postProcess: function (out) {
```

src/aws.ts
```diff
@@ -1,3 +1,32 @@
...
+          return {
+            name: line,
+          };
+        });
+      }
+    } catch (e) {
+      console.error(e);
+    }
+
+    return [];
```
fedeci, line 26: `trycatch` does nothing in this case and `String.prototype.split` always returns an array of at least one value if the string exists:
```suggestion
    const lines = out.split("\n");
    return lines.map((line) => {
      return {
        name: line,
      };
    });
```

fedeci (changes requested):
Thanks for the improvement to this spec @joeheyming!

joeheyming:
Screen Shot 2022-10-23 at 11 20 25 AM

fedeci (approved):
Thanks!

PR #1663, opened 2022-10-21
feat(nextflow): add secrets subcommand

mribeirodantas:
- feat(nextflow): Add secrets subcommand
- feat(nextflow): Add generator for secret names

**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
It adds autocomplete to the secrets subcommand in the nextflow CLI and a generator
for autocompleting the argument of the `nextflow secrets get` subcommand.

PR #1677, opened 2022-10-25
feat(nextflow): add secrets command and generator

mribeirodantas:
- feat(nextflow): Add secrets command and generator

It adds autocomplete for the new secrets command, and a generator for its args.

There is no autocomplete for the secrets command.

### DIFF
--- a/src/nextflow.ts
+++ b/src/nextflow.ts
@@ -40,6 +40,21 @@ const dockerimage: Fig.Generator = {
   },
 };
 
+const secretname: Fig.Generator = {
+  script: `grep -o '"name": *"[^"]*"' $HOME/.nextflow/secrets/store.json | grep -o '"[^"]*"$' | tr -d \\"`,
+  postProcess: (output) => {
+    if (output == "") {
+      return [];
+    }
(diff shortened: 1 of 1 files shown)


src/nextflow.ts
```diff
@@ -40,6 +40,21 @@ const dockerimage: Fig.Generator = {
   },
 };
 
+const secretname: Fig.Generator = {
+  script: `grep -o '"name": *"[^"]*"' $HOME/.nextflow/secrets/store.json | grep -o '"[^"]*"$' | tr -d \\"`,
```
fedeci, line 44: See https://github.com/withfig/autocomplete/pull/1663#discussion_r1002768582

mribeirodantas: Oh, I hadn't seen that. Thanks for bringing this up. It does work, but Fig takes a while to load the secrets which is not ideal, in my opinion.

With the original script line (the one comitted here):
grep

Calling `nextflow secrets list` in the script line (the suggestion from the other PR):
nextflow

fedeci (approved):
Looks good then! Thanks!

PR #1686, opened 2022-10-29
feat: add lsof spec

joeheyming:
In this commit, we want to add the lsof spec.

I basically cross referenced the `lsof -h` command with `man lsof` and did my best to make it easy to understand most of the args.

Some of the args I had to do some word-smithing to make it short enough to fit inside the autocomplete window.

I tried to make the -i option more dynamic, and added fancy ways of grabbing ip address or service, etc...

**Additional info:**

### DIFF
--- a/src/lsof.ts
+++ b/src/lsof.ts
@@ -1,0 +1,479 @@
+const completionSpec: Fig.Spec = {
+  name: "lsof",
+  description: "List open files",
+  args: {
+    name: "names",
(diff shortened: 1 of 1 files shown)


joeheyming:
can I get a hacktoberfest-accepted label on this so that I can qualify

joeheyming:
Any feedback? Is this in a good enough state to submit?

PR #1687, opened 2022-10-30
feat: add launchctl completion spec.

joeheyming:
This is a basic implementation of launchctl.
We don't want to attempt to list the launchctl labels or service targets.  That can be a future exercise.

Feature

Currently autocomplete doesn't support launchctl

Now launchctl is supported

**Additional info:**

### DIFF
--- a/src/launchctl.ts
+++ b/src/launchctl.ts
@@ -52,6 +52,27 @@ const limitArgs = [
   },
 ];
 
+const listGenerator: Fig.Generator = {
+  script: "launchctl list",
+  postProcess: function (out) {
+    return out
+      .split("\n")
+      .slice(1)
+      .map((line) => {
+        const parts = line.split(/\s/);
+        return {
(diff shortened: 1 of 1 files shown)


joeheyming:
Please add the `hacktoberfest-accepted` or `hacktoberfest` label to this review so it can count towards my contributions.

grant0417:
Thanks for the contrib! Have been wanting this one, could you use `launchctl list` as a generator for some of the arguments?

joeheyming:
Originally, I added the list generator, but it didn't work for all commands.  I felt like it was too much work , so I removed it.  But since you asked, I added it back for commands that use a "label" arg.  But I would have liked to also make it work for service targets and stuff like that, but I think that will have to be future work.

grant0417:
Seems reasonable, LGTM, will let @SeparateRecords merge though!

mschrage (approved):

PR #1690, opened 2022-11-02
feat(cargo): add more generators

grant0417:
Adds better generators based on `cargo metadata`, adds generators for packages, bins, example, test, and bench. Also adds a generator for versions when using search on the `@` trigger.


SeparateRecords:
LGTM! Thanks for continuing to improve Cargo so much 🦀🦀🦀🦀🦀🦀🦀

SeparateRecords (approved):

PR #1694, opened 2022-11-07
feat(brew): add `home` subcommand

FoRRestDp:
Feature that adds `brew home` command autocomplete.

As for now there is no autocomplete for such command.

### DIFF
--- a/src/brew.ts
+++ b/src/brew.ts
@@ -1558,7 +1558,7 @@ const completionSpec: Fig.Spec = {
       ],
     },
     {
-      name: "home",
+      name: ["home", "homepage"],
       description:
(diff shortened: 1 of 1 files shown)


src/brew.ts
```diff
@@ -1557,6 +1557,29 @@ const completionSpec: Fig.Spec = {
         },
       ],
     },
+    {
+      name: "home",
```
fedeci, line 1561:
```suggestion
      name: ["home", homepage],
```

fedeci (changes requested):
Thanks @FoRRestD ! Just an addition and it is ready to be merged!

fedeci (approved):
Thanks!

PR #1695, opened 2022-11-07
feat(fig): add spec for 2.0.0

mschrage:

### DIFF
--- a/src/fig/2.0.0.ts
+++ b/src/fig/2.0.0.ts
@@ -1,3 +1,18 @@
+import {
+  settingsSpecGenerator,
+  subsystemsGenerator,
+  themesGenerator,
+  pluginsGenerator,
+  tokensGenerators,
+  invitationsGenerators,
+  membersGenerators,
+  teamsGenerators,
+  workflowsSpecGenerator,
+  sshHostsGenerator,
+  sshIdentityGenerator,
+  userGenerator,
+} from "./shared";
+
 const completion: Fig.Subcommand = {
   name: "fig_cli",
   description: "Top level cli commands",
@@ -454,6 +469,7 @@ const completion: Fig.Subcommand = {
             name: "files",
             isVariadic: true,
             isOptional: true,
+            generators: subsystemsGenerator,
           },
         },
         {
@@ -847,6 +863,7 @@ const completion: Fig.Subcommand = {
           isOptional: true,
         },
       ],
+      generateSpec: settingsSpecGenerator,
     },
     {
       name: "tips",
@@ -977,6 +994,7 @@ const completion: Fig.Subcommand = {
           args: {
             name: "auth",
             isOptional: true,
+            generators: sshIdentityGenerator,
           },
         },
         {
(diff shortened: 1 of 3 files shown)


SeparateRecords:
oh you're kidding, it removed all the generators 😭😭😭😭

SeparateRecords:
am fixing generators now, one-time effort until cli 3.0 😉

Will merge when done

SeparateRecords:
@grant0417 any generators obviously missing?

fedeci:
@SeparateRecords generators removed are a known issue not solvable with the current approach 😕, we need to use ASTs modifications in order to keep them.

src/fig/2.0.0.ts
```diff
@@ -0,0 +1,4146 @@
...
+  membersGenerators,
+  teamsGenerators,
+  workflowsSpecGenerator,
+  sshHostsGenerator,
+  sshIdentityGenerator,
+  userGenerator,
+} from "./shared";
+
+const completion: Fig.Subcommand = {
+  name: "fig_cli",
```
fedeci, line 17:
```suggestion
  name: "fig",
```

src/fig/2.0.0.ts
```diff
@@ -0,0 +1,4146 @@
...
+            },
+          ],
+        },
+        {
+          name: "diagnostics",
+          description: "Watches diagnostics",
+          options: [
+            {
+              name: "--rate",
+              isRepeatable: true,
```
fedeci, line 546: This does not sound right to me, is this option really repeatable?
Same for a bunch of other options like the ones from the `query-index` subcommand.
```suggestion
              isRepeatable: true,
```

fedeci (changes requested):
@grant0417 can you check the two comments I left, please?

fedeci:
@mschrage I used filemerge to get diffs between the old and new specs. 
Should I build a web tool that calculates diffs from automatic PRs bumping majors and that publishes those diffs to an html accessible by everyone?

mschrage:
@fedeci that would be useful, but I don't think that many people besides us use this tooling - so no need to over optimize it right now!

src/fig/2.0.0.ts
grant0417: Yeah, that doesnt look right, hmm, an issue with the clap generator perhaps

SeparateRecords: Let's fix this in an update to the spec - imo for now we should just land this PR. If it's an error in the clap generator, we'll remove it later, if it's in the CLI itself then we'll fix it with semver

fedeci:
> @fedeci that would be useful, but I don't think that many people besides us use this tooling - so no need to over optimize it right now!

Sounds good, eventually I'll implement it in my free time!

src/fig/2.0.0.ts
fedeci: Great! @SeparateRecords if everything looks good to you feel free to merge.

fedeci (approved):

grant0417:
Going to merge the changes!

PR #1712, opened 2022-11-18
feat: add gt (graphite.dev CLI) completion spec

laurenz-glueck:
Adds completion spec for `gt` cli (CLI of [graphite.dev](https://graphite.dev/))

Currently there is no autocomplete for the `gt` cli 

Now there is autocomplete for the `gt` cli 😉

### DIFF
--- a/src/gt.ts
+++ b/src/gt.ts
@@ -1,46 +1,3 @@
-const defaultOptions = [
-  {
-    name: "--version",
-    description: "Show Graphite CLI version",
-    priority: 0,
-  },
-  {
-    name: ["--help", "-h"],
-    description: "Show help for Graphite CLI",
-    priority: 20,
-  },
-  {
-    name: "--interactive",
-    description: "Prompt the user. Disable with --no-interactive",
-    priority: 10,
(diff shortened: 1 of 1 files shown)


src/gt.ts
```diff
@@ -0,0 +1,1016 @@
+const defaultOptions = [
+  {
+    name: "--version",
+    description: "Show Graphite CLI version",
+    priority: 0,
```
fedeci, line 5: You can add this property to "global" options and just add the default options to the `completionSpec.options` array removing them from subcommands. In this way all options with this property set to true will be inherited by subcommands.
```suggestion
    priority: 0,
    isPersistent: true,
```

fedeci (changes requested):
Thanks @laurenz-glueck! There is a small nitpick to fix that may increase maintainability of the spec in the future.

src/gt.ts
laurenz-glueck: Alright - I didn't know about the `isPersistent` flag 👌🏼 
Changes are implemented ✅

PR #1715, opened 2022-11-22
fix: remove html escapes

grant0417:
AE no longer needs these to be escaped, an in fact looks bad if they are!


src/just.ts
```diff
@@ -142,9 +142,9 @@ function getRecipeUsage(recipe: Recipe): string {
   for (const parameter of recipe.parameters) {
     // Fig sanitizes things like "<NAME>", so this has to be encoded
```
grant0417, line 143:
```suggestion
```

mschrage (approved):

PR #1716, opened 2022-11-23
fix: add missing flags to ls

mschrage:
Add `--color` and `-,` to ls

### DIFF
--- a/src/ls.ts
+++ b/src/ls.ts
@@ -192,13 +192,13 @@ const completionSpec: Fig.Spec = {
     {
       name: "-,",
       description: `When the -l option is set, print file sizes grouped and separated by thousands using the non-monetary separator returned
-        by localeconv(3), typically a comma or period.  If no locale is set, or the locale does not have a non-monetary separator, this
-        option has no effect.  This option is not defined in IEEE Std 1003.1-2001 (“POSIX.1”).`,
+by localeconv(3), typically a comma or period.  If no locale is set, or the locale does not have a non-monetary separator, this
+option has no effect.  This option is not defined in IEEE Std 1003.1-2001 (“POSIX.1”)`,
       dependsOn: ["-l"],
     },
     {
       name: "--color",
-      description: `Output colored escape sequences based on when, which may be set to either always, auto, or never.`,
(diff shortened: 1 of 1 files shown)


src/ls.ts
```diff
@@ -189,6 +189,36 @@ const completionSpec: Fig.Spec = {
...
+    {
+      name: "-,",
+      description: `When the -l option is set, print file sizes grouped and separated by thousands using the non-monetary separator returned
+        by localeconv(3), typically a comma or period.  If no locale is set, or the locale does not have a non-monetary separator, this
+        option has no effect.  This option is not defined in IEEE Std 1003.1-2001 (“POSIX.1”).`,
+      dependsOn: ["-l"],
+    },
+    {
+      name: "--color",
+      description: `Output colored escape sequences based on when, which may be set to either always, auto, or never.`,
```
fedeci, line 201:
```suggestion
      description: `Output colored escape sequences based on when, which may be set to either always, auto, or never`,
```

src/ls.ts
```diff
@@ -189,6 +189,36 @@ const completionSpec: Fig.Spec = {
       description:
         "Distinguish dataless files and directories with a '%' character in long (-l) output, and don't materialize dataless directories when listing them",
     },
+    {
+      name: "-,",
+      description: `When the -l option is set, print file sizes grouped and separated by thousands using the non-monetary separator returned
+        by localeconv(3), typically a comma or period.  If no locale is set, or the locale does not have a non-monetary separator, this
+        option has no effect.  This option is not defined in IEEE Std 1003.1-2001 (“POSIX.1”).`,
+      dependsOn: ["-l"],
```
fedeci, line 197:
```suggestion
      description: `When the -l option is set, print file sizes grouped and separated by thousands using the non-monetary separator returned
by localeconv(3), typically a comma or period.  If no locale is set, or the locale does not have a non-monetary separator, this
option has no effect.  This option is not defined in IEEE Std 1003.1-2001 (“POSIX.1”)`,
      dependsOn: ["-l"],
```

fedeci (approved):

PR #1730, opened 2022-12-03
feat: add a lot of stuff to `cargo` and `rustup`

grant0417:

**Additional info:**

### DIFF
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -1,3 +1,92 @@
+type ToolchainLocalGeneratorOptions = {
+  excludeShort?: boolean;
+};
+
+// A generator of local toolchains including the short names
+const toolchainLocalGenertor: (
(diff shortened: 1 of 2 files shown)


src/rustup.ts
```diff
@@ -1,3 +1,92 @@
...
+
+    return shortToolchains.concat(toolchains).map((name) => ({
+      name,
+    }));
+  },
+});
+
+const toolchainAllGenerator: Fig.Generator = {
+  // Grab the latest versions of rust from github, try the gh cli first as it has a higher rate limit
+  script:
```
grant0417, line 32: Dont love this, might switch to just doing it in a custom generator

grant0417: Ah, the issues is I can not get the exit code of a command in a custom generator without doing something like `(command -v github >/dev/null && echo "0") || echo "1"`

PR #1745, opened 2022-12-09
feat: add spec for spring cli

### LINKED ISSUE #1341: Spring CLI Integration
### Sanity checks

- [X] I have searched [github.com/withfig/fig/issues](https://github.com/withfig/fig/issues?q=) and there are no duplicates of my feature

### Feature Details

for example auto-complete available commands with spring
to show available commands run: spring 

Captura de Pantalla 2022-07-11 a la(s) 22 00 25

Thank you for developing Fig, is amazing and I would love to contribute!

tmwrnr:
feature

n/a

Adds a new spec for [spring](https://docs.spring.io/spring-boot/docs/current/reference/html/cli.html)

**Additional info:**
Closes #1341


target/npmlist.json
```diff
@@ -0,0 +1,13 @@
+{
```
grant0417, line 1: Why did you add this file?

src/spring.ts
```diff
@@ -0,0 +1,312 @@
...
+      displayName: version.name,
+    }));
+  },
+};
+
+const completionSpec: Fig.Spec = {
+  name: "spring",
+  description:
+    "The Spring Boot CLI is a command line tool that you can use to bootstrap a new project from start.spring.io or encode a password",
+  icon: "https://spring.io/images/favicon-9d25009f65637a49ac8d91eb1cf7b75e.ico",
```
grant0417, line 104: Is there a `.jpg` or `.png` version? This is fine just prefer those formats.

src/spring.ts
```diff
@@ -0,0 +1,312 @@
...
+  custom: async (_, executeShellCommand) => {
+    const data = await fetchData(executeShellCommand);
+    if (!data) return [];
+
+    return data.dependencies.values
+      .map(({ values }) => values)
+      .flat()
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .map((dependency) => ({
+        name: dependency.id + ",",
```
grant0417, line 67: Swapping this to just `dependency.name` and adding `insertValue` might be a better way to do this.

grant0417 (changes requested):
Looks pretty good! Just a few comments

target/npmlist.json
tmwrnr: I actually don't know where this file came from. Is removed in the next commit

src/spring.ts
tmwrnr: I need to use `dependency.id`, because this is the actual name. `dependency.name` is a display name:

```json
{
  "description": "Java library to generate text output (HTML web pages, e-mails, configuration files, source code, etc.) based on templates and changing data.",
  "id": "freemarker",
  "name": "Apache Freemarker"
}
```
I tried to find a good solution for the `","` problem. My aim was that the command looks like that:

```sh
spring init -d=freemarker,lombok,datadog
```
But I ran into multiple problems. When I used `requiresSeparator` the `=` was printed, but the suggestions for the dependencies only triggered after typing a space after the `=`. I noticed that the CLI can also handle the argument without a separator so I removed it. 

The next problem was, that the dependencies need to be separated with a `,`. I could not find a right parameter for that use case. I tried `replaceValue` but this only added an `,` after the first dependency.  I am unsure how  `inserValue` can help me with this problem.

Could you give a me an example how I can improve this?

tmwrnr: @grant0417 Any update on this?

grant0417: Ah, didn't see it before. @fedeci might ahve a better idea on how to use `requiresSeparator` with a list.

fedeci: @tmwrnr we provide the `valueList` generator from the `@fig/autocomplete-generators` package but it only supports static values at the moment. I would suggest you to copy its implementation and integrate in it the methods required to get the package list.
```ts
// implementation
export function valueList({
  delimiter = ",",
  values = [],
  cache = false,
  insertDelimiter = false,
  allowRepeatedValues = false,
}: ValueListInit): Fig.Generator {
  return {
    trigger: (newToken, oldToken) =>
      newToken.lastIndexOf(delimiter) !== oldToken.lastIndexOf(delimiter),

    getQueryTerm: (token) => token.slice(token.lastIndexOf(delimiter) + delimiter.length),

    custom: async (...init) => {
      const out = await getSuggestions(values, insertDelimiter ? delimiter : "", cache, init);
      if (allowRepeatedValues) {
        return out;
      }
      const [tokens] = init;
      const valuesInList = tokens[tokens.length - 1]?.split(delimiter);
      return removeRepeatSuggestions(valuesInList, out);
    },
  };
}
```

tmwrnr: @grant0417 @fedeci I improved the generator. Is this how you imagined it?

tmwrnr: @grant0417 @fedeci Anything new for this?

grant0417: I think this is fine, going to merge and sorry for the delay on this, thanks for the contribution!

PR #1728, opened 2022-12-01
fix(npm): improve package suggestions

loosheng:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
pref 
**What is the current behavior? (You can also link to an open issue here)**
close #275
close #974
**What is the new behavior (if this is a feature change)?**

better-npm

**Additional info:**
use https://api.npms.io/v2/search/suggestions?q=%q&size=20

PR #1754, opened 2022-12-17
fix(npm): search does not work

loosheng:
https://github.com/withfig/autocomplete/pull/1728#discussion_r1038179446

Wrong code suggestion causes search to not work.😅

@fedeci


fedeci (approved):
Thanks! My fault not catching this!

PR #1760, opened 2022-12-18
feat: add ubuntu `do-release-upgrade` spec

mrhappyma:
Info taken from https://manpages.ubuntu.com/manpages/bionic/en/man8/do-release-upgrade.8.html

### DIFF
--- a/src/do-release-upgrade.ts
+++ b/src/do-release-upgrade.ts
@@ -1,0 +1,54 @@
+const modeArg: Fig.Arg = {
+  name: "mode",
+  suggestions: ["desktop", "server"],
+};
+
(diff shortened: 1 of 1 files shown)


mrhappyma:
According to the man page, the `-m` and `--mode` options are the same except that only *`--mode`* requires an equals sign for the argument. Same with `-f` and `--frontend`.

My solution to this was to add each option separately and hide the shorter ones. But perhaps there's a more elegant way of doing this.

grant0417:
Thanks for the contrib! I think that is probably the best way to do it if they take arguments in different ways even thought they are the same flag

PR #1215, opened 2022-05-08
feat(yarn): show more executable packages

fedeci:
Closes #1211

Issue #1211, opened 2022-05-07
[yarn] [npx] Suggest npm packages which install into `node_modules/.bin`


tpluscode:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [ ] I would like to work on this.

### What CLI tool does this relate to?

yarn, npx, npm

### Which statement makes the most sense?

This completion spec does not exist

### Issue Details

Many npm packages provide a CLI tool, defined by the [bin package.json field](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bin)

It would be great if fig suggested at least the names of said binaries installed into local `node_modules`. This is supported by running `yarn [bin]` and `npx [bin]`.

I suppose this could work in one of two possible ways:

1. By introspecting the entire contents of closest `node_modules/.bin`
2. By analysing `package.json` only of packages explicitly listed in the local dependencies (preferred?)

As a stretch goal, it would be an absolute killer feature if individual packages would offer even more suggestion. (I hear that packages based on commander can be auto-completed too?)

fedeci:
`npx` is already suggesting those executables located under .bin, I am adding that feature to yarn too!

PR #1762, opened 2022-12-20
fix(yarn): Suggest subcommands with bin only 

simoncrypta:
A member of RedwoodJS community notice that Fig don't autocomplete redwood CLI anymore. 
After looking into recent changes, I see that suggestions are now base by the binaries. #1211 #1215 
Is also take packages from package.json into consideration for suggestion, problem is package name is not always the same as binary name. 

My suggestion is to only go with the binaries, since is more accurate of what is available. 
Whatever, I think we should only suggest subcommand from the list, otherwise this creates a lot a noise by irrelevant suggestion.

cc @fedeci

### DIFF
--- a/src/yarn.ts
+++ b/src/yarn.ts
@@ -28,21 +28,14 @@ export const nodeClis = new Set([
   "typeorm",
   "babel",
   "remotion",
-  "@withfig/autocomplete-tools",
-  "@redwoodjs/core",
+  "autocomplete-tools",
+  "redwood",
+  "rw",
   "create-completion-spec",
-  "@fig/publish-spec-to-team",
+  "publish-spec-to-team",
   "capacitor",
   "cap",
 ]);
-
-type SearchResult = {
-  package: {
-    name: string;
-    description: string;
-  };
-  searchScore: number;
-};
 
 // generate global package list from global package.json file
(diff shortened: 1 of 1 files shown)


src/yarn.ts
```diff
@@ -28,22 +28,15 @@ export const nodeClis = new Set([
   "typeorm",
   "babel",
   "remotion",
-  "@withfig/autocomplete-tools",
-  "@redwoodjs/core",
+  "autocomplete-tools",
+  "redwood",
+  "rw",
   "create-completion-spec",
-  "@fig/publish-spec-to-team",
```
simoncrypta, line 34: I take a guess that this wouldn't have @fig/ name in binaries

src/yarn.ts
```diff
@@ -28,22 +28,15 @@ export const nodeClis = new Set([
...
+  "publish-spec-to-team",
   "capacitor",
   "cap",
 ]);
 
-type SearchResult = {
```
simoncrypta, line 39: This was unused

simoncrypta:

fedeci (approved):
Thanks @simoncrypta!

fedeci:
recheck

PR #1766, opened 2022-12-22
fix(figterm): wrong version flag

mrhappyma:


mrhappyma:
dang it, commit signing didnt work again

fedeci (approved):
Thanks!

PR #1767, opened 2022-12-22
feat: add `chsh` spec

mrhappyma:
https://man.archlinux.org/man/chsh.1.en

### DIFF
--- a/src/chsh.ts
+++ b/src/chsh.ts
@@ -1,0 +1,43 @@
+const shells: Fig.Generator = {
+  script: "chsh -l",
+  postProcess: (output) => {
+    if (output.startsWith("fatal:")) {
+      return [];
+    }
+    return output.split("\n").map((shell) => {
+      return { name: shell.replace("*", "").trim() };
+    });
+  },
+};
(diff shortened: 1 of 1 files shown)


src/chsh.ts
```diff
@@ -0,0 +1,43 @@
+const shells: Fig.Generator = {
```
fedeci, line 1:
```suggestion
// TODO: this does not work on macos
const shells: Fig.Generator = {
```

fedeci (approved):
Thanks

src/chsh.ts
```diff
@@ -0,0 +1,43 @@
...
+      description: "Print help message and exit",
+    },
+    {
+      name: ["-v", "--version"],
+      description: "Print version and exit",
+    },
+  ],
+  args: {
+    name: "username",
+    description: "Target user",
```
fedeci, line 40:
```suggestion
    description: "Target user",
    isOptional: true,
```

mrhappyma: good catch

PR #1776, opened 2023-01-01
feat: add `asciinema` spec

mrhappyma:


grant0417:
Thanks for the contrib!

PR #1797, opened 2023-01-12
feat(fig): update spec

mschrage:


grant0417:
@fedeci, how do we stop the bot from overriding the `generateSpec` for `fig cli`

fedeci:
@grant0417 by updating the generator. Diff versioned spec run on js objects and not on ast representations which are more complex. The plan is to update it, and it can possibly be prioritised after AE fixes.

PR #1821, opened 2023-01-24
feat(fig): update spec

mschrage:

### DIFF
--- a/src/fig/2.0.0.ts
+++ b/src/fig/2.0.0.ts
@@ -5022,5 +5022,1645 @@ versions["2.11.0"] = {
   ],
 };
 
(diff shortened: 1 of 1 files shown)


src/fig/2.0.0.ts
```diff
@@ -5022,5 +5022,1645 @@ versions["2.11.0"] = {
...
+    },
+    {
+      name: "cli",
+      options: [
+        {
+          name: ["-h", "--help"],
+          description: "Print help",
+        },
+      ],
+      args: {
```
grant0417, line 6619: This is still being overriden @fedeci

PR #1825, opened 2023-01-27
feat: add hatch spec

juftin:
## Summary

Added a spec for [hatch](https://github.com/pypa/hatch) via `click_complete_fig`

image


juftin:

src/hatch.ts
```diff
@@ -0,0 +1,731 @@
...
...nv-run) command.\n\n    If the first argument contains a colon, then the preceding component will be\n    interpreted as the name of the environment to target, overriding the `-e`/`--env`\n    [root option](#hatch) and the `HATCH_ENV` environment variable.\n\n    If the environment provides matrices, then you may also provide leading arguments\n    starting with a `+` or `-` to select or exclude certain variables, optionally\n    followed by specific comma-separated values. For example, if you have the\n    following configuration:\n\n    === ":octicons-file-code-16: pyproject.toml"\n\n        \b\n        ```toml\n        [[tool.hatch.envs.test.matrix]]\n        python = ["3.9", "3.10"]\n        version = ["42", "3.14", "9000"]\n        ```\n\n    === ":octicons-file-code-16: hatch.toml"\n\n        \b\n        ```toml\n        [[envs.test.matrix]]\n        python = ["3.9", "3.10"]\n        version = ["42", "3.14", "9000"]\n        ```\n\n    then running:\n\n    \b\n    ```\n    hatch run +py=3.10 -version=9000 test:pytest\n    ```\n\n    would execute `pytest` in the environments `test.py3.10-42` and `test.py3.10-3.14`.\n    Note that `py` may be used as an alias for `python`.\n',
```
fedeci, line 671: These kind of formatted descriptions are not the best to be rendered in JS, can you make them simpler?

fedeci (changes requested):

src/hatch.ts
juftin: Yep, I can do that. This is what's coming over via the `click_complete_fig` tool

fedeci: Yeah, we do not transform fields in our integrations, so if the string in the hatch codebase is like that it will be the same for the generated spec.

PR #1827, opened 2023-01-27
feat(tldr): update to latest version

### LINKED ISSUE #1826: tldr - Oudated spec.
### Sanity checks

tldr

### Issue Details

The `tldr` spec is outdated.

1. The local cache folder needs to be corrected. It is moved from `~/.tldr/cache` to `~/.tldrc/tldr`.
2. `-p, --platform` option is missing.
3. The following options are no longer supported.
   * `-a, --list-all`
   * `--random`
   * `--random-example`
   * `-m, --markdown`

seenukarthi:
This PR fixes the completion spec `tldr`
 
Closes #1826

Sgned-off-by: Karthikeyan Vaithilingam <seenukarthi@gmail.com>

### DIFF
--- a/src/tldr.ts
+++ b/src/tldr.ts
@@ -134,7 +134,7 @@ const completionSpec: Fig.Spec = {
     {
       name: "--platform",
       description: "Select platform",
-      insertValue: "--platform={cursor}",
+      requiresSeparator: true,
       args: {
         name: "platform",
         suggestions: ["linux", "osx", "sunos", "windows", "common"],
--- a/src/tldr.ts
+++ b/src/tldr.ts
@@ -1,15 +1,16 @@
-const android = `~/.tldr/cache/pages/android/`;
-const common = `~/.tldr/cache/pages/common/`;
-const linux = `~/.tldr/cache/pages/linux/`;
-const osx = `~/.tldr/cache/pages/osx/`;
-const sunos = `~/.tldr/cache/pages/sunos/`;
-const windows = `~/.tldr/cache/pages/windows/`;
+const tldrRc = `~/.tldrc/tldr`;
+const android = `${tldrRc}/pages/android/`;
+const common = `${tldrRc}/pages/common/`;
+const linux = `${tldrRc}/pages/linux/`;
+const osx = `${tldrRc}/pages/osx/`;
+const sunos = `${tldrRc}/pages/sunos/`;
+const windows = `${tldrRc}/pages/windows/`;
 
 const isMarkDownRegex = new RegExp(/^.*\.md$/);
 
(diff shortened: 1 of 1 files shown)


src/tldr.ts
```diff
@@ -139,6 +123,23 @@ const completionSpec: Fig.Spec = {
...
+      description: "Select platform",
+      args: {
+        name: "platform",
+        suggestions: ["linux", "osx", "sunos", "windows", "common"],
+      },
+    },
+    {
+      name: "--platform",
+      description: "Select platform",
+      insertValue: "--platform={cursor}",
```
AlanSyue, line 137: Maybe we can use `requiresSeparator` instead.
```suggestion
      requiresSeparator: true,
```
https://github.com/withfig/autocomplete-tools/blob/main/types/index.d.ts#L713-L725

src/tldr.ts
```diff
@@ -139,6 +123,23 @@ const completionSpec: Fig.Spec = {
       name: ["-c", "--clear-cache"],
       description: "Delete the entire local cache",
     },
+    {
+      name: "-p",
+      description: "Select platform",
+      args: {
+        name: "platform",
+        suggestions: ["linux", "osx", "sunos", "windows", "common"],
```
AlanSyue, line 131: [nit]
Maybe we should declare a variable to gather these keywords?

https://github.com/withfig/autocomplete/pull/1827/files#diff-50bd02ab6f158b81b333c17f0cc4c4058f28dbdb7e710d9d0acb4cca77a99a13R140

AlanSyue (approved):

src/tldr.ts
seenukarthi: @AlanSyue Yes have made the suggestions constant.

AlanSyue (approved):

fedeci (changes requested):
Thanks @seenukarthi! I left a comment to solve before merging this.

src/tldr.ts
```diff
@@ -139,6 +125,23 @@ const completionSpec: Fig.Spec = {
...
+    },
+    {
+      name: "--platform",
+      description: "Select platform",
+      requiresSeparator: true,
+      args: {
+        name: "platform",
+        suggestions: platformSuggestions,
+      },
+    },
```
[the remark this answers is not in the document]
seenukarthi: @fedeci the option `-p` expects a white space between `-p` and the argument, on the other hand `--platform` need a `=` between option and the argument that's the reason I made it as desperate options.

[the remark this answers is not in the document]
fedeci: `--platform` is not requiring the equals nor `-p` requires a space.

PR #1837, opened 2023-02-03
feat: add `firewall-cmd` spec

mrhappyma:
https://firewalld.org/documentation/man-pages/firewall-cmd.html
i really hate these sorts of options-based commands...


mschrage:
Are you still working on this @mrhappyma ?

PR #1840, opened 2023-02-07
feat: add `klist` completion spec

seenukarthi:
This PR adds completion spec for `klist`

`klist` reads and displays the current tickets in the credential cache (also known as the ticket file).


fedeci (approved):
Thanks!

Issue #1755, opened 2022-12-17
[k9s] allow `k9s -n` to list all namespaces


AlanSyue:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [X] I would like to work on this.

### What CLI tool does this relate to?

k9s

### Which statement makes the most sense?

This completion spec isn't complete

### Issue Details

When I use `k9s -n` commands, I found we should support it to list all namespaces to make users can choose the namespace they want to use easily.

Maybe we can use `kubectl` command to find all namespaces.
reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#viewing-namespaces
```
kubectl get namespace
```

PR #1848, opened 2023-02-14
feat: search namespace for k9s

AlanSyue:
# Summary
It is convenient for users to list all namespaces when using the `--namespace` option in the `k9s` command.
截圖 2023-02-14 上午11 44 49

# Issue
Closes https://github.com/withfig/autocomplete/issues/1755

### DIFF
--- a/src/k9s.ts
+++ b/src/k9s.ts
@@ -1,3 +1,20 @@
+const namespaces: Fig.Generator = {
+  script: () => {
+    return `kubectl get namespaces`;
+  },
+  postProcess: (out) => {
+    return out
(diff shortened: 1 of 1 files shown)


src/k9s.ts
```diff
@@ -1,3 +1,20 @@
+const namespaces: Fig.Generator = {
+  script: () => {
+    return `kubectl get namespaces`;
+  },
```
fedeci, line 4:
```suggestion
  script: "kubectl get namespaces",
```

fedeci (approved):
Thanks!

PR #1849, opened 2023-02-14
style: use lint:fix to format the file

AlanSyue:
Since there are two code style issues that let the CI pipline fail, I run the command `yarn lint:fix` to fix them
image


fedeci:
Thanks!

PR #1860, opened 2023-02-19
fix(@usermn/sdc): remove v in version 0.0.7

mrhappyma:
Looks like I forgot to update the workflow to remove the v. Oops.


fedeci (approved):
My fault merging it without checking that🤦. Thanks!

PR #1871, opened 2023-02-28
fix(jenv): make the argument of `jenv exec` as command

seenukarthi:
Made the argument of `jenv exec` as a command so fig will autocomplete the command properly.


fedeci (approved):
Thanks!

PR #1874, opened 2023-03-01
feat: add tkn spec

BalliAsghar:
Added completion spec for [tkn](https://tekton.dev/docs/cli/) - A CLI for interacting with Tekton!


fedeci (approved):
Thanks @BalliAsghar! Just for curiosity, was this written manually or autogenerated? It is a gigantic spec!

PR #1876, opened 2023-03-03
feat: add `xxd` completion spec

seenukarthi:
This PR adds completion spec for `xxd`

`xxd` creates a hex dump of a given file or standard input.  It can also convert a hex dump back to its original binary form.  Like uuencode(1) and uudecode(1), it allows the transmission of binary data in a 'mail-safe' ASCII representation but has the advantage of decoding to standard output.

### DIFF
--- a/src/xxd.ts
+++ b/src/xxd.ts
@@ -56,6 +56,9 @@ const completionSpec: Fig.Spec = {
     {
       name: ["-len", "-l"],
       description: "Stop after writing <len> octets",
+      args: {
+        name: "len",
+      },
     },
     {
       name: ["-name", "-n"],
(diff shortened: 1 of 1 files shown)


src/xxd.ts
```diff
@@ -0,0 +1,112 @@
...
+      },
+    },
+    {
+      name: ["-include", "-i"],
+      description: "Output in C include file style",
+      exclusiveOn: ["-EBCDIC", "-E", "-bits", "-b"],
+    },
+    {
+      name: ["-len", "-l"],
+      description: "Stop after writing <len> octets",
```
fedeci, line 58:
```suggestion
      description: "Stop after writing <len> octets",
      args: {
        name: "len",
      },
```

fedeci (approved):
Thanks @seenukarthi!

PR #1884, opened 2023-03-19
feat(clilol): update spec

mcornick:

### DIFF
--- a/src/clilol.ts
+++ b/src/clilol.ts
@@ -31,7 +31,6 @@ const completionSpec: Fig.Spec = {
           name: "paste",
           description: "Create or update a paste",
           args: { name: "title", description: "Title of the paste" },
-
           options: [
             {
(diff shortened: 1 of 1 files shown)


src/clilol.ts
mcornick, line 1: I'm not sure where these whitespace deletions are coming from. https://github.com/mcornick/clilol/blob/v1.0.0/clilol.ts doesn't remove the whitespace and matches https://github.com/withfig/autocomplete/blob/master/src/clilol.ts. Is it something in the withfig/push-to-fig-autocomplete-action workflow?

fedeci: Yeah, we run prettier on the action workflow. I would say that the generated spec should not be commited to prevent this kind of problems. I am merging this one in anyway.

fedeci (approved):
:)

PR #1892, opened 2023-03-25
feat: add `aws-vault` spec

swallowtail62:
This PR adds [aws-vault](https://github.com/99designs/aws-vault) completion spec.

### DIFF
--- a/src/aws-vault.ts
+++ b/src/aws-vault.ts
@@ -1,4 +1,4 @@
-const profilesGeneratro: Fig.Generator = {
+const profilesGenerator: Fig.Generator = {
   script: "aws-vault list --profiles",
   postProcess(out) {
     return out.split("\n").map((name) => ({ name }));
--- a/src/aws-vault.ts
+++ b/src/aws-vault.ts
@@ -1,0 +1,168 @@
+const profilesGeneratro: Fig.Generator = {
+  script: "aws-vault list --profiles",
+  postProcess(out) {
+    return out.split("\n").map((name) => ({ name }));
+  },
+};
+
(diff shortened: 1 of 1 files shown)


swallowtail62:

src/aws-vault.ts
```diff
@@ -0,0 +1,168 @@
+const profilesGeneratro: Fig.Generator = {
```
fedeci, line 1: typo:
```suggestion
const profilesGenerator: Fig.Generator = {
```

src/aws-vault.ts
```diff
@@ -0,0 +1,168 @@
...
+  name: "aws-vault",
+  description:
+    "A vault for securely storing and accessing AWS credentials in development environments",
+  subcommands: [
+    {
+      name: "help",
+      description: "Show help about the command",
+      args: {
+        name: "command",
+        suggestions: helpSuggestions,
```
fedeci, line 146: This is a nice sugar we provide :)
```suggestion
        template: "help",
```

fedeci (changes requested):
Thanks @swallowtail62! I left a couple minor issues to solve but the spec looks good overall.

src/aws-vault.ts
swallowtail62: Oops, sorry. Thanks for pointing that out!

src/aws-vault.ts
swallowtail62: @fedeci 
Thanks! I fixed it, check it out.

fedeci (approved):

fedeci:
Thanks @swallowtail62!

PR #1896, opened 2023-03-26
feat: add sls completion spec

t3yamoto:
This PR adds `sls` completion spec.
`sls` is a shorthand for the same [serverless](https://github.com/withfig/autocomplete/blob/master/src/serverless.ts).

### DIFF
--- a/src/sls.ts
+++ b/src/sls.ts
@@ -1,5 +1,6 @@
+import serverless from "./serverless";
 const completionSpec: Fig.Spec = {
+  ...serverless,
   name: "sls",
-  loadSpec: "serverless",
 };
 export default completionSpec;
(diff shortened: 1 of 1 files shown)


src/sls.ts
```diff
@@ -0,0 +1,5 @@
+const completionSpec: Fig.Spec = {
+  name: "sls",
+  loadSpec: "serverless",
+};
```
fedeci, line 4: This is more performant:

```suggestion
import serverless from "./serverless";
const completionSpec: Fig.Spec = {
  ...serverless,
  name: "sls",
};
```

fedeci (changes requested):

PR #1908, opened 2023-04-07
feat(xc): add xc autocomplete

mckelveygreg:
Just started using a nifty new tool [xc](https://github.com/joerdav/xc) to run scripts straight from the `README.md`, however, I missed being able to have autocomplete for scripts defined in a `package.json`. 

Thank you Fig for your tool! I can't look back to a terminal without autocomplete, so I guess its time to contribute!

Let me know if there is anything else I can do. I was unsure if I picked the correct pattern for dynamic sub-commands since it has to pick them up from your readme. And I wasn't sure if I applied the options that were command specific in the right location, vs you can run the command with no arguments and it will list what subcommands are available.

### DIFF
--- a/src/xc.ts
+++ b/src/xc.ts
@@ -1,0 +1,95 @@
+/**
+ * https://github.com/joerdav/xc
+ * xc - Simple, Convenient, Markdown-based task runner.
+ * v0.2.0
+ */
+const completionSpec: Fig.Spec = {
+  name: "xc",
+  description: "List tasks from an xc-compatible markdown file",
+  generateSpec: async (context, executeShellCommand) => {
+    const options: Fig.Option[] = [
+      {
+        name: ["-f", "-file"],
+        args: {
+          name: "path",
+          template: "filepaths",
+        },
+        description:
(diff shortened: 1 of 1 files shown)


mckelveygreg:

mckelveygreg:
recheck

Ozgurix (approved):

Ozgurix (approved):
I would like to work on this

fedeci:
@Ozgurix this is a PR so there is nothing to work on apart from reviewing code :) I suggest you to look up on the issues of this repo to understand which specs still need to be added!

src/xc.ts
```diff
@@ -0,0 +1,95 @@
...
+        description: desc,
+        options,
+      }));
+
+    return {
+      name: "xc",
+      subcommands,
+    };
+  },
+  requiresSubcommand: false,
```
fedeci, line 54: This is the default:
```suggestion
```

fedeci (approved):
Thanks @mckelveygreg! I am committing a nitpick I found, but the spec is great!

PR #1918, opened 2023-04-16
feat: add `ipatool` completion spec

tmwrnr:
feature

n/a

Adds a new spec for [ipatool](https://github.com/majd/ipatool)

Additional info:
n/a

### DIFF
--- a/src/ipatool.ts
+++ b/src/ipatool.ts
@@ -15,7 +15,7 @@ type App = {
 
 const bundleIdentifierGenerator: Fig.Generator = {
   script: (context) => {
-    const identifier = context.at(3);
+    const identifier = context[context.length - 1];
     if (!identifier) return "";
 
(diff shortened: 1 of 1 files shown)


src/ipatool.ts
```diff
@@ -0,0 +1,195 @@
...
+  id: number;
+  bundleID: string;
+  name: string;
+  version: string;
+  price: number;
+};
+
+const bundleIdentifierGenerator: Fig.Generator = {
+  script: (context) => {
+    const identifier = context.at(3);
```
fedeci, line 18: Keeping `3` as a constant may result in wrong suggestions, instead the last item of the context is always the current argument, see https://fig.io/docs/reference/generator/contextual#writing-a-contextual-script-function
```suggestion
    const identifier = context[context.length - 1];
```

fedeci:
Thanks!

fedeci (approved):

PR #1927, opened 2023-04-20
feat(ncu): add support for npm-check-updates

ikosta:
Adds support for https://www.npmjs.com/package/npm-check-updates aka `ncu`

### DIFF
--- a/src/ncu.ts
+++ b/src/ncu.ts
@@ -76,6 +76,7 @@ const optionsDict: NcuOptions = {
   },
   cacheFile: {
     name: "--cacheFile",
+    dependsOn: ["--cache"],
     description:
       'Filepath for the cache file. Only works with --cache. (default: "~/.ncu-cache.json")',
     args: {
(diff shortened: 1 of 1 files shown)


fedeci (changes requested):
Thanks @ikosta! I left some notes to fix before merging. I am wondering why you have decided to implement the the options in a separate object and not directly in the spec? (Just curiosity it is fine to me)

ikosta:
> Thanks @ikosta! I left some notes to fix before merging. I am wondering why you have decided to implement the the options in a separate object and not directly in the spec? (Just curiosity it is fine to me)

I think it's easier to maintain if there are changes in the cli and new functionality uses same options.

Since this spec has no subcommands it seems overpowered but only future will proof we wrong.

Thanks for the comments I will check them out.

ikosta:
@fedeci just checked all the requested changes and these make completely sense. Thanks!

fedeci (approved):

PR #1934, opened 2023-04-26
feat: add np spec

uncenter:
Basic implementation of the spec for `np`. Still needs generators for versioning and branches I believe, but all options and arguments are here.


fedeci (approved):
Thanks for the PR @uncenter! I am merging this because it is a working spec and it is great to have it as soon as possible, and I strongly suggest you to open another PR with generators to maximise the UX!

Issue #1942, opened 2023-01-19
git commands don't handle dash (-) shorthand


afberg:
### Checks

- [X] I have searched [github.com/withfig/fig/issues](https://github.com/withfig/fig/issues?q=) and there are no duplicates of my issue
- [X] I have run `fig doctor` in the affected terminal session
- [X] I have run `fig restart` and replicated the issue again

### Operating system

macOS 12.3.0 (21E230)

### Expected behaviour

When writing e.g. `git rebase -` I expect to be able to submit the command and rebase on the last checked out branch.

### Actual behaviour

Fig doesn't recognize the dash as a branch name, and instead tries to autocomplete to `--abort`

### Steps to reproduce

Type `git rebase -` and hit enter in a fig–enabled terminal

### Environment

```yaml
fig-details:
  - 2.12.0
hardware-info:
  - model: 
  - model-id: 
  - chip-id: Apple M1 Max
  - cores: 10
  - mem: 64.00 GB
os-info:
  - macOS 12.3.0 (21E230)
environment:
  - shell: /bin/zsh
  - terminal: iterm
  - cwd: /Users/anton.forsberg/universe
  - exe-path: /opt/homebrew/bin/fig
  - install-method: brew
  - env-vars:
    - FIGTERM_SESSION_ID: 5af38834-ce8a-49fb-91f2-dea8c39363d7
    - FIG_HOSTNAME: VXY4HDQ2YH
    - FIG_SET_PARENT_CHECK: 1
    - FIG_TERM: 2.8.0
    - PATH: /opt/homebrew/bin:/opt/homebrew/sbin:/Users/anton.forsberg/.nvm/versions/node/v16.18.0/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/homebrew/bin:/opt/homebrew/sbin:/Users/anton.forsberg/.fig/bin:/Users/anton.forsberg/.local/bin:/Users/anton.forsberg/Library/Application Support/JetBrains/Toolbox/scripts
    - SHELL: /bin/zsh
    - TERM: xterm-256color
    - __CFBundleIdentifier: com.googlecode.iterm2
    - FIG_PID: 1276
    - FIG_SET_PARENT: 5af38834-ce8a-49fb-91f2-dea8c39363d7
```

Gautam-Arora24:
I think the easiest solution is to press `ESC` and then hitting the Enter button.

fedeci:
[...]

PR #1948, opened 2023-05-08
fix(git): add new option to handle shorthand dash

Gautam-Arora24:
## What does this PR introduces?
This PR adds support for shorthand `-` in git rebase command.

Fixes: #1942

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -4885,6 +4885,10 @@ const completionSpec: Fig.Spec = {
       description: "Reapply commits on top of another base tip",
       options: [
         {
+          name: "-",
+          description: "Rebase on the last checked out branch",
+        },
+        {
           name: "--onto",
           description:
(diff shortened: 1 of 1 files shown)


Gautam-Arora24:

src/git.ts
```diff
@@ -4884,6 +4884,10 @@ const completionSpec: Fig.Spec = {
       name: "rebase",
       description: "Reapply commits on top of another base tip",
       options: [
+        {
+          name: "-",
+          description: "Rebase on the last checked out branch",
+        },
```
fedeci, line 4890: In this case `-` replaces the "base ref" argument so it should be considered as an argument and not an option.
Can you please update the `base` argument of `rebase` to look something similar to this:
```ts
{
  name: "base",
  generators: gitGenerators.localBranches,
  suggestions: [{
    name: "-",
    description: "Use the last ref as the base"
  }],
  filterStrategy: "fuzzy",
  isOptional: true,
},
```

fedeci (changes requested):
@Gautam-Arora24 thanks for the PR! I left a comment to address before merging

fedeci (approved):
Thanks!

PR #1964, opened 2023-05-13
feat(go-task): get nested and root tasks

ddanielcruzz:
Current Fig Generator doesn't work when having nested Taskfiles and also doesn't list root tasks. 

### Before

I have my Taskfiles split in different files, and reference them all in one file at the root of my project, current autocompletion doesn't get the tasks in my root as nor the tasks until I get into the directory containing all my Taskfiles.  

https://github.com/withfig/autocomplete/assets/20308945/5bc2d32f-86db-48cf-8eef-4b552df91439

### After

My new approach works as expected with the root tasks while maintaining previous behavior

https://github.com/withfig/autocomplete/assets/20308945/f4661727-458c-42e0-9320-1b01c9382a1c

### DIFF
--- a/src/task/go-task.ts
+++ b/src/task/go-task.ts
@@ -1,44 +1,31 @@
 const TASKS_PRIORITY = 80;
 const TASKFILE_FLAGS = ["-t", "--taskfile"];
 const DIRECTORY_FLAGS = ["-d", "--dir"];
-const taskDirectoryFlags = new Set([...TASKFILE_FLAGS, ...DIRECTORY_FLAGS]);
 
 const tasksGenerator: Fig.Generator = {
-  custom: async (tokens, exec) => {
-    // No additional tasks after "-- "
-    if (tokens.join(" ").includes("-- ")) {
+  script: "task -a",
+  postProcess: (output) => {
+    if (output.includes("task: No Taskfile found")) {
       return [];
     }
 
-    let context = "";
-    const contextValueIndex =
-      tokens.length -
-      tokens
-        .slice()
-        .reverse()
-        .findIndex((token) => taskDirectoryFlags.has(token));
+    const result: Fig.Suggestion[] = output
+      .split("\n")
+      .filter((task) => task.startsWith("*"))
+      .map((task) => {
+        const taskInfo = task.slice(2).trim();
+        const [name, ...description] = taskInfo.split(" ");
 
-    // Add the last context flag if mentioned in tokens
-    if (contextValueIndex <= tokens.length) {
-      context = ` ${tokens[contextValueIndex - 1]} ${
-        tokens[contextValueIndex]
-      }`;
-    }
-
-    const tasksListText = await exec(`task${context} --list`);
-
-    return tasksListText
-      .split("\n* ")
-      .slice(1) // The first line is not a task
-      .map((taskText) => {
-        const [name, description] = taskText.split(": ");
         return {
-          name,
-          description,
-          type: "subcommand",
+          name: name.replace(/:$/, ""),
+
+          description: description.join(" ") || "Task",
+          icon: "🎯",
           priority: TASKS_PRIORITY,
         };
       });
+
+    return result;
   },
 };
 
--- a/src/task/go-task.ts
+++ b/src/task/go-task.ts
@@ -14,13 +14,12 @@ const tasksGenerator: Fig.Generator = {
       .filter((task) => task.startsWith("*"))
       .map((task) => {
         const taskInfo = task.slice(2).trim();
-        const [name, ...description] = taskInfo.split(" ");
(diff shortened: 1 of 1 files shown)


ddanielcruzz:

fedeci (approved):
Thanks! LGTM. @AlanSyue, I've seen you wrote this spec, can you give me a feedback on whether context still matters when using `--list-all`? I am not an user of go-task but from what I have seen in the docs this spec should be good.

AlanSyue:
@fedeci Sorry, I haven't used `go-task` before. Last time, I only used linter to format this spec.

fedeci:
My bad, I've actually inspected the commit history , but the 5mins that elapsed between the inspection and the review were enough to make me forget who created and who edited it😅 @yoavbls you are the creator, any feedback on this?

src/task/go-task.ts
```diff
@@ -1,44 +1,31 @@
...
-      .map((taskText) => {
-        const [name, description] = taskText.split(": ");
         return {
-          name,
-          description,
-          type: "subcommand",
+          name: name.replace(/:$/, ""),
+
+          description: description.join(" ") || "Task",
+          icon: "🎯",
```
yoavbls, line 23: I found this icon a little bit wierd, what about using `"fig://icon?type=command"` instead to be aligned with all the other fig commands?
or use the `go-task` icon: `https://taskfile.dev/img/favicon.ico`

src/task/go-task.ts
```diff
@@ -1,44 +1,31 @@
 const TASKS_PRIORITY = 80;
 const TASKFILE_FLAGS = ["-t", "--taskfile"];
 const DIRECTORY_FLAGS = ["-d", "--dir"];
-const taskDirectoryFlags = new Set([...TASKFILE_FLAGS, ...DIRECTORY_FLAGS]);
 
 const tasksGenerator: Fig.Generator = {
-  custom: async (tokens, exec) => {
-    // No additional tasks after "-- "
-    if (tokens.join(" ").includes("-- ")) {
+  script: "task -a",
```
yoavbls, line 6: Great! you can also add the `task -a` command to the completion spec

src/task/go-task.ts
```diff
@@ -1,44 +1,31 @@
...
-      .slice(1) // The first line is not a task
-      .map((taskText) => {
-        const [name, description] = taskText.split(": ");
         return {
-          name,
-          description,
-          type: "subcommand",
+          name: name.replace(/:$/, ""),
+
+          description: description.join(" ") || "Task",
```
yoavbls, line 22: I know that nobody likes regex but becaue you're already using it for the name let me suggest this thing instead:

```typescript
const { name, description } = task.match(
  /\* (?<name>[\w\:\-]+): +(?<description>[\w\s]+)/
).groups;

return {
    name,
    description,
    ...
}
```

yoavbls: If you don't like the regex you can try this which is more clear in my opinion:

```typescript 
const [namePart, descriptionPart] = task.split(": ");
return {
    name: namePart.slice(2);
    description: descriptionPart.trim() || undefined;
    ...
}

```

yoavbls: Keep notice that in both cases when there isn't a description it returns undefined which is more correct than "Task" here. fig will display "No description" automatically

yoavbls (changes requested):
It looks good and indeed works with nested and root tasks!
I left a few comments. Let me know what you think 🙂

src/task/go-task.ts
fedeci: Yeah, it should be together with `--list-all`

ddanielcruzz:
@fedeci Sure, I'll do it tomorrow.

src/task/go-task.ts
ddanielcruzz: Done, I went with `"fig://icon?type=command"`

src/task/go-task.ts
ddanielcruzz: Added!

src/task/go-task.ts
ddanielcruzz: I ended up with this 
```ts
 const [name, description] = taskInfo.split(": ");

        return {
          name: name.replace(/:$/, ""),
          description: description?.trim(),
          icon: "fig://icon?type=command",
          priority: TASKS_PRIORITY,
        };
```

I kept the `replace` since the colon was present in root tasks

Left is the previous approach, right is the one you suggested, see how `formatwrite` still has a colon. 
Screenshot 2023-06-02 at 10 55 04

yoavbls (approved):
@ddanielcruzz  Looks good to me 🙂

fedeci:
Thanks!

PR #1976, opened 2023-05-17
feat(remotion): update spec + autogeneration

JonnyBurger:
Instead of writing the spec manually, we are trying out to generate the Fig spec from our monorepo, so that when people add new options, they can also add autocompletions in the same PR.

In the future, we can even import our CLI parsing logic and provide to Fig the data that we also use to parse the arguments.
We run the Fig ESLint rules also in our repo before sending it to Fig.

Because of that, this is a huge diff, but in the future it will be smaller.


fedeci:
@JonnyBurger thanks! Have you checked out https://fig.io/docs/guides/integrating/push-to-autocomplete-action#what-is-push-to-autocomplete to automatically push to this repo?

fedeci (approved):
Thanks!

PR #1995, opened 2023-06-08
feat: add guided option and optimize profile option in SAM CLI

AlanSyue:
## Summary
1. Add guided option
image

2. Fetch AWS profile name when using profile option

## Test result
截圖 2023-06-08 上午9 00 14

### DIFF
--- a/src/sam.ts
+++ b/src/sam.ts
@@ -189,6 +189,8 @@ const profileOption: Fig.Option = {
   description:
     "Select a specific profile from your credential file to get AWS credentials",
   args: {
+    name: "profile name",
+    description: "The name of the AWS profile",
     generators: awsProfileGenerator,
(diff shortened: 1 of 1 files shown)


src/sam.ts
```diff
@@ -187,11 +189,17 @@ const profileOption: Fig.Option = {
   description:
     "Select a specific profile from your credential file to get AWS credentials",
   args: {
-    name: "profile name",
```
grant0417, line 190: Could you leave both the `name` and `description` lines?

grant0417 (changes requested):

src/sam.ts
AlanSyue: @grant0417 fixed, thanks!

PR #2040, opened 2023-07-26
add [history] spec

programmer4229:


grant0417:
Hey @programmer4229 can you sign the CLA?

src/history.ts
```diff
@@ -0,0 +1,29 @@
...
+      name: "|",
+      description: "Divider to run other commands on history",
+      subcommands: [
+        {
+          name: "grep",
+          description: "Search history",
+        },
+      ],
+    },
+  ],
```
mschrage, line 19: These should not be subcommands. `0` is an argument. (You can pass a number of history items you wish to show.)

The pipe and "grep" should be removed entirely since these are not syntax of the history command, but other shell constructs and commands.

mschrage (changes requested):

PR #2077, opened 2023-09-04
feat(wasp): Add wasp autocomplete

Zeko369:
This PR adds support for [Wasp](https://wasp-lang.dev/) CLI

- some of the deploy commands are missing

### DIFF
--- a/src/wasp.ts
+++ b/src/wasp.ts
@@ -1,0 +1,281 @@
+const ICON =
+  "https://user-images.githubusercontent.com/1536647/77317442-78625700-6d0b-11ea-9822-0fb21e557e87.png";
+
+const newCommand: Fig.Subcommand = {
+  name: "new",
+  icon: ICON,
+  description: "Create a new Wasp project",
+  args: {
+    name: "project-name",
+    description: "Name of the project",
+  },
+  options: [
+    {
+      name: ["--template", "-t"],
+      icon: "📄",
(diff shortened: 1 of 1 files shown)


src/wasp.ts
```diff
@@ -0,0 +1,281 @@
...
+
+const completionSpec: Fig.Spec = {
+  name: "wasp",
+  description:
+    "Amazing fullstack framework for building web apps with less boilerplate",
+  generateSpec: async (context, executeShellCommand) => {
+    const isWaspDir =
+      (await executeShellCommand(
+        `while [ ! -f "$PWD/main.wasp" ] && [ "$PWD" != "/" ]; do cd ..; done; [ -f "$PWD/main.wasp" ] && echo "true" || echo "false"`
+      )) == "true";
```
mschrage, line 257: This would be a good thing for us to abstract into @withfig/autocomplete-tools.

src/wasp.ts
```diff
@@ -0,0 +1,281 @@
...
+  requiresSubcommand: false,
+  subcommands: [
+    {
+      name: "generate",
+      description: "Generates bash completion script",
+      insertValue: `${"\b".repeat(
+        COMPLETION.length + 1
+      )}${COMPLETION}:generate`,
+    },
+  ],
```
mschrage, line 88: What are you trying to do here?

PR #2109, opened 2023-10-02
feat(intelij based editors): Reuse idea spec for other intelij based editors

Zeko369:


grant0417 (approved):
Woah nice! Might be the most number of CLIs contributed in a single PR in a while!

PR #2112, opened 2023-10-02
feat(rails): Missing db commands and runner

Zeko369:


src/rails.ts
```diff
@@ -354,6 +354,20 @@ const dbOptions = [
...
 ];
 
+const environmentOption: Fig.Option = {
+  name: ["-e", "--environment"],
+  description:
+    "Specifies the environment to run this console under (test/development/production)",
+  args: {
+    name: "environment",
+    generators: {
+      script: "ls config/environments",
```
grant0417, line 364: Is this file also relative to the command being run? Is is possible it is in some parent directory?

Zeko369: Hey, I've setup a command to do that (I think we also mention this in another one of my PRs)

I'd want to extract this since I want to use it on other pending PRs + refactor some of the other CLIs so I'd move it to `@withfig/autocomplete-tools` or to a share lib file in this repo?

Thoughs?

Function -> https://github.com/withfig/autocomplete/pull/2112/files#diff-34f5c090cacdca031210e68db93072f048a5d6f31428fe107d0fdcbd19da4803R6-R50

Zeko369: CC extraction of this function by @mschrage 
 https://github.com/withfig/autocomplete/pull/2077#discussion_r1340749010

PR #2118, opened 2023-10-09
feat(premake): add premake spec

Drischdaan:
I added the spec for [premake](https://premake.github.io/).

Straight from the premake readme:
"Premake is a command line utility which reads a scripted definition of a software project, then uses it to perform build configuration tasks or generate project files for toolsets like Visual Studio, Xcode, and GNU Make. Premake's scripts are little [Lua](http://www.lua.org/) programs, so the sky's the limit!"

### DIFF
--- a/src/premake.ts
+++ b/src/premake.ts
@@ -1,0 +1,186 @@
+const options: Fig.Option[] = [
+  {
+    name: "--file",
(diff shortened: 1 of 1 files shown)


Drischdaan:

grant0417 (approved):

PR #2135, opened 2023-10-19
spec(bundle): Add gem autocomplete and binaries autocomplete

Zeko369:
This PR also includes
1. stubs for rspec and rubocop (will be done in a separate PR later, but the 2 most common things are added)
2. utility to convert a spec into a suggestion (that's gonna be used in other PRs i.e. npx to load i.e. create-vite to DRY up icons/names of bins across specs)

### DIFF
--- a/src/rubocop.ts
+++ b/src/rubocop.ts
@@ -1,0 +1,23 @@
+const completionSpec: Fig.Spec = {
+  name: "rubocop",
+  description:
+    "A Ruby static code analyzer and formatter, based on the community Ruby style guide",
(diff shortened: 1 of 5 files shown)


src/rubocop.ts
```diff
@@ -0,0 +1,23 @@
+const completionSpec: Fig.Spec = {
+  name: "rubocop",
+  description:
+    "A Ruby static code analyzer and formatter, based on the community Ruby style guide",
+  icon: "https://github.com/rubocop/rubocop/blob/master/logo/rubo-logo-symbol.png?raw=true",
```
grant0417, line 5: Are you able to replace this with a smaller version, it seems to load a very large version of the asset.

PR #2143, opened 2023-10-19
Better Teleport CLI support

kanersps:
This PR has two areas of focus, both meant to significantly improve usability for the "tsh" and "tctl" Teleport CLI tools. These two areas of focus are:

- Improving username@hostname autocomplete for `tsh ssh`
- Full support for the `tctl` binary, which was non existent

This is my first autocompletion spec PR, so please do be gentle. I've been using these completions for a little bit myself and found them incredibly useful. Certainly the updates I did for `tsh ssh` has been an incredible productivity enhancer.

A small question, and maybe a worry: The custom generator and trigger I use for `tsh ssh` might decrease performance based on the past iteration, but I do feel that improved UX is worth it. On my M1/M2 macs I do not notice any performance issues.

And P.S. I don't think your Git pre-commit hooks are properly working, as they cause a big error to popup. Let me know if you want to investigate this, but I just used "--no-verify" and ran the checks manually, not a big deal.


src/tctl.ts
```diff
@@ -0,0 +1,1957 @@
+import { type } from "node:os";
+
```
Zeko369, line 2: Pretty sure this was imported automatically by mistake right?

```suggestion
```

kanersps: Yup, funny how that got in there.

Currently fixing up some of the arguments and style a lil' and I'll push an update.

mschrage: @kanersps are you still working on this? Would love to get this contribution merged in :)

kanersps: Hey @mschrage ! I certainly am, I have it completely ready locally but I accidentally messed some Git-things up. Hopefully have this ready by today, but the commits might be a lil' messed up. I advise squashing before merging it.

kanersps:
This should be all the features I wanted to implement, I'm not sure if I'm entirely done with messing with the Teleport autocompletions specs, but this will do for this MR. 

Let me know if there are any changes required!

kanersps:
@mschrage Was there anything else that needs to be changed here, or can we merge? Locally 'publishing' this spec constantly is becoming a lil' annoying (probably a bug too, I presume, every so often I need to re-publish to my local namespace to ensure I have use these specs vs the one in this repo)

kanersps:
Good morning @superbryntendo . I see you made a quick commit, and I think that will break the functionality. I've specifically used a custom trigger, because I need to know the "username" portion as well to make a proper autocomplete suggestion.

Right now, with your change, the "username" variable does not exist and this will error out.

PR #2144, opened 2023-10-19
feat(typst): add `typst` completion spec

toddlerer:
Add spec for [Typst](https://typst.app/) CLI completion

### DIFF
--- a/src/typst.ts
+++ b/src/typst.ts
@@ -1,0 +1,376 @@
+import { filepaths } from "@fig/autocomplete-generators";
+
+const completionSpec: Fig.Spec = {
+  name: "typst",
+  description: "The Typst compiler",
+  icon: "📄",
+  subcommands: [
+    {
+      name: ["compile", "c"],
+      description: "Compiles an input file into a supported output format",
+      icon: "📄",
+      options: [
+        {
+          name: "--root",
+          description: "Configures the project root (for absolute paths)",
+          args: {
+            name: "root",
+            template: "folders",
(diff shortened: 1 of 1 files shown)


toddlerer:

src/typst.ts
```diff
@@ -0,0 +1,376 @@
...
+        name: "cert",
+        template: "filepaths",
+      },
+    },
+    {
+      name: ["-v", "--verbosity"],
+      description:
+        "Sets the level of logging verbosity: -v = warning & error, -vv = info, -vvv = debug, -vvvv = trace",
+      isRepeatable: 4,
+    },
```
mschrage, line 364: A bit of an edge case, but I think it makes sense to split these into two options.

```suggestion
    {
      name: ["-v"],
      description:
        "Sets the level of logging verbosity",
      isRepeatable: 4,
    },
    {
         name: ["--verbosity"],
         args: {
            name: "level",
            suggestions: ["warning", "error", "info", "debug", "trace"]
         }
    }
    
```

toddlerer: Right. An option like that `--verbosity` might be easier to use. Since I am not an [typst](https://github.com/typst/typst) maintainer, it might be good if you could suggest it for typst.

grant0417: I believe this is verbatim how the command works right now, you can literally repeat `--verbosity` up to 4 times.

grant0417 (approved):
LGMT!

PR #866, opened 2022-01-22
Added azure all command and subCommand

ikushaldave:
**What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**

This will add all the azure command available in azure CLI.

**What is the current behavior? (You can also link to an open issue here)**

This part of autocompletion was missing in the fig.

**What is the new behavior (if this is a feature change)?**

This add most of azure command to fig autocompletion but have to work adding azure CLI options which is missing in this PR.

**Additional info:**

PR #2146, opened 2023-10-23
feat: implement `az` spec

cpendery:
Implements the azure command line via a diff versioned spec, picking up the work started in #866. The code is generated via my [autocomplete-az](https://github.com/cpendery/autocomplete-az) repo via `npx autocomplete-az`. 

Due to `az` cli surface changes introduced in every minor version change, as reflected by the [release notes](https://github.com/MicrosoftDocs/azure-docs-cli/blob/main/docs-ref-conceptual/release-notes-azure-cli.md), I thought it best to add new specs for each minor version, but happy for feedback if there is a better approach or any concerns with this


cpendery:
Additionally, I'm unable to test the spec as I don't have a macbook, so if a maintainer could double check that the way I setup the `loadSpec` actually works for a versioned spec, that would be super helpful

cpendery:

mschrage:
Thank you so much for this massive contribution @cpendery! I will test locally to confirm `loadSpec` works as expected.

mschrage:
@cpendery Also can you resolve the linting errors and the type checking errors?

cpendery:
> @cpendery Also can you resolve the linting errors and the type checking errors?

@mschrage happy to fix the linting errors, but it looks like the PR makes the type checker run out of memory, any advice on how to resolve that?

cpendery:
Also @mschrage does doing the version specs like this make sense for every minor release? I don't want to bloat the repo unnecessarily. Would it be better to just use generateSpec with versions when things change?

grant0417:
Fixed some small issues with this but this looks great! Thanks so much for the contrib!

PR #2158, opened 2023-11-02
feat(podman): add podman completion spec

diegofcornejo:
In addition to the specs for Podman, 'pretty-quick --staged' was removed from package.json. It was not possible to run the pre-commit because it was giving the error '_prettier.resolveConfig.sync is not a function', which is documented at https://github.com/azz/pretty-quick/issues/164. The package has not been updated for more than 2 years.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,8 +51,7 @@
   },
   "lint-staged": {
     "*.ts": [
-      "eslint --fix",
-      "pretty-quick --staged"
+      "eslint --fix"
     ]
   },
   "files": [
--- a/src/podman.ts
+++ b/src/podman.ts
@@ -1,0 +1,3429 @@
+const postProcessDockerPs: Fig.Generator["postProcess"] = (out) => {
+  return out.split("\n").map((i) => {
+    try {
+      const parsedJSON: Record<string, string> = JSON.parse(i);
+      return {
+        name: parsedJSON.Names,
+        displayName: `${parsedJSON.Names} (${parsedJSON.Image})`,
+        icon: "fig://icon?type=docker",
+      };
+    } catch (error) {
+      console.error(error);
+    }
+  });
+};
+
+const sharedPostProcess: Fig.Generator["postProcess"] = (out) => {
+  return out
+    .split("\n")
+    .map((line) => JSON.parse(line))
+    .map((i) => ({
+      name: i.name || i.Name, //Added for backwards compatibility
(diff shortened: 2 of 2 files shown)


diegofcornejo:

diegofcornejo:
> Seems like the type check is still failing

Hi @grant0417 , I'm not sure what's wrong because I've checked the code and types in docker.ts. The linter is showing errors. On the other hand, every autocomplete with script (function way) has the same implementation, but in those cases, it seems like the type check doesn't fail."

```js

// podman.ts
script: function (context) {
      if (context[context.length - 1] === "") return undefined;
      const searchTerm = context[context.length - 1];
      return ["podman", "search", searchTerm, "--format", "{{ json . }}"];
    }

script: function (context) {
              let fileFlagIndex, dockerfilePath;
              if (context.includes("-f")) {
                fileFlagIndex = context.indexOf("-f");
                dockerfilePath = context[fileFlagIndex + 1];
              } else if (context.includes("--file")) {
                fileFlagIndex = context.indexOf("--file");
                dockerfilePath = context[fileFlagIndex + 1];
              } else {
                dockerfilePath = "$PWD/Dockerfile";
              }

              return `\\grep -iE 'FROM.*AS' "${dockerfilePath}"`;
            }
```

```js
// docker.ts
script: function (context) {
      if (context[context.length - 1] === "") return undefined;
      const searchTerm = context[context.length - 1];
      return ["docker", "search", searchTerm, "--format", "{{ json . }}"];
    }

script: function (context) {
              let fileFlagIndex, dockerfilePath;
              if (context.includes("-f")) {
                fileFlagIndex = context.indexOf("-f");
                dockerfilePath = context[fileFlagIndex + 1];
              } else if (context.includes("--file")) {
                fileFlagIndex = context.indexOf("--file");
                dockerfilePath = context[fileFlagIndex + 1];
              } else {
                dockerfilePath = "Dockerfile";
              }

              return ["grep", "-iE", "FROM.*AS", dockerfilePath];
            }
```

diegofcornejo:
> If the linter is showing errors for docker you may need to run `yarn install` again to update your dependencies, the final line of the docker one `return ["grep", "-iE", "FROM.*AS", dockerfilePath];` is what it should look like.

I understand now, I fixed the line, everything should be fine!

grant0417 (approved):

Issue #2173, opened 2023-11-14
[nx] autocomplete doesn't work properly anymore 


overbit:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [x] I would like to work on this.

### What CLI tool does this relate to?

nx

### Which statement makes the most sense?

There is a bug with this completion spec

### Issue Details

Similar issue reported https://github.com/withfig/autocomplete/issues/1928, but this time it came back after this release:
https://github.com/withfig/autocomplete/pull/2161

To add more details, the nx target that are not shown anymore in the autocomplete are ones that defined in the monorepo

overbit:
@grant0417 could you please validate if #2161 is the reason of this issue?

PR #2182, opened 2023-11-22
fix(nx): cache generation issue due to wrong shell command

overbit:
Close: #2173


overbit:

PR #2217, opened 2023-12-28
Create spec for barnard59

tpluscode:
https://github.com/zazuko/barnard59/


tpluscode:

tpluscode:
recheck

src/barnard59.ts
```diff
@@ -0,0 +1,411 @@
...
+        },
+        {
+          name: ["--pipeline"],
+          description: "IRI of the pipeline description",
+          args: { name: "iri", isOptional: true },
+        },
+        {
+          name: ["--variable"],
+          description: "variable key value pairs",
+          args: { name: "name=value", default: "[object Map]" },
```
grant0417, line 22: This is likely a wrong default.

src/barnard59.ts
```diff
@@ -0,0 +1,411 @@
...
+            {
+              name: ["--cube"],
+              description: "cube URI",
+              isRequired: true,
+              args: { name: "cube" },
+            },
+            {
+              name: ["--variable"],
+              description: "variable key value pairs",
+              args: { name: "name=value", default: "[object Map]" },
```
grant0417, line 64: Same here

src/barnard59.ts
```diff
@@ -0,0 +1,411 @@
...
+            },
+            {
+              name: ["--profileFormat"],
+              description: "Override the profile format",
+              args: { name: "profileFormat" },
+            },
+            {
+              name: ["--variable"],
+              description: "variable key value pairs",
+              args: { name: "name=value", default: "[object Map]" },
```
grant0417, line 104: Same here

grant0417 (changes requested):
Looks good besides a lot of instances of `default: "[object Map]"`

tpluscode:
Thank you for the review @grant0417 . As you can see, this was generated...

About the bogus defaults. We have options with custom transforms which build a `Map` object. The default is an empty map, hence the JavaScript's default string representation that object. I would not like to alter the code to account for fig's autocomplete as it requires additional guards in multiple places to handle a possible undefined value for that option. 

What about the linting problems? I can fix some casing but it also complains about `display help for command` which is generated the commander itself. And I don't know what to do about these which say `If `X` is a single value it should not be enclosed in square brackets`. I think that needs fixing in `@fig/complete-commander`

tpluscode:
ping @grant0417

grant0417:
Hey @tpluscode, I have updated the `@fig/complete-commander` package with the fixes for the `[object Map]` default, can you regenerate your spec and submit again?

tpluscode:
Done @grant0417

PR #2222, opened 2024-01-04
Switch to pnpm

grant0417:


Zeko369:
Any specific reason for `pnpm` over `yarn4` / `bun`?

grant0417:
Just what I have been using in other projects, any reason you have for `yarn4` over `pnpm`?

Zeko369:
> Just what I have been using in other projects, any reason you have for `yarn4` over `pnpm`?

Nope, every project I'm using is either `pnpm` or `yarn@1` with some new being in `bun`

Also since fig currently only runs on macos `bun install` might be a better option overall (since it's not on windows yet)

With warm caches + updated lock files it's:
```
bun: 1.372s
pnpm: 3s
```

grant0417:
I definitely appreciate the speed increases of bun but not sure I want to buy in fully since we are only using as a package manager and not for the other benefits. We also are looking at cross platform support so sticking with yarn or pnpm probably makes sense.

mschrage (approved):

PR #2243, opened 2024-01-19
Update with newest Remotion options

JonnyBurger:
A few new options we added!

Btw, I know there are better options to submit it automatically, we still plan to do it in the long-term and automatically derive Fig schemas from our actual logic.

For now, I just submit the file that we semi-automate in our monorepo from time to time.


grant0417 (approved):
LGTM, thanks! Also cool to see vulkan support, will have to check it out again soon!

PR #2261, opened 2024-02-06
Adam changes

atjbramley:
Couple of corrections and changes:

- auth is API keys not Username + MP
- corrected descriptions
- shortened config - server urls to read better in GUI


atjbramley:
Sorry - opened in origin instead of working branch :/

PR #2300, opened 2024-03-23
feat(z): include cwd folders in suggestions for zoxides `z` command

ScottRobinson03:
## What
The [existing spec](https://github.com/withfig/autocomplete/blob/master/src/z.ts) for the `z` command includes folders of the cwd for [rupa/z]( https://github.com/rupa/z), but not [zoxide](https://github.com/ajeetdsouza/zoxide)'s `z`. This PR adds autocomplete suggestions for zoxide's `z` command and therefore resolves aws/codewhisperer-command-line-discussions#114.

I also added icons as per the following:

:file_folder: ... = not in zoxide history, but in cwd/pwd
:floppy_disk: ... = in zoxide history AND cwd/pwd
:floppy_disk:/... = in zoxide history, but NOT cwd/pwd (notice the leading `/` in path)

Example:
image

## Known Issues:

For some reason I can't seem to consistently get all of zoxide's cwd suggestions above the non-zoxide cwd suggestions. I'm using the `priority` key to try and achieve this, but it doesn't seem to work here for some reason? Would appreciate assistance in debugging this.

image

### DIFF
--- a/src/z.ts
+++ b/src/z.ts
@@ -125,9 +125,12 @@ const zoxideCompletionSpec: Fig.Spec = {
     name: "directory",
     filterStrategy: "fuzzy",
     suggestCurrentToken: true,
-    isVariadic: true,
     generators: {
-      custom: async (tokens, executeShellCommand) => {
+      custom: async (
+        tokens,
+        executeShellCommand,
+        { currentWorkingDirectory }
+      ) => {
         let args;
         if (tokens.length < 2 || tokens[1] === "") {
           args = ["query", "--list", "--score"];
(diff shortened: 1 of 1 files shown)


src/z.ts
```diff
@@ -125,9 +125,12 @@ const zoxideCompletionSpec: Fig.Spec = {
     name: "directory",
     filterStrategy: "fuzzy",
     suggestCurrentToken: true,
-    isVariadic: true,
```
ScottRobinson03, line 128: Opted to remove this `isVariadic: true` since whilst the argument is strictly speaking variadic, the autocomplete for it is going to be wrong 99% of the time and so figured it's better to not give suggestions than give wrong suggestions.

ScottRobinson03:

grant0417:
The issue with priority not working is likely history which will result in previously selected items to be at the top even when a high priority is set, this is something I will need to look into fixing but at the moment this should be fine.

PR #2356, opened 2024-06-02
Update npm to use the cli for package and version queries

### LINKED ISSUE #2255: [npm] support private registry
### Sanity checks

npm

### Issue Details

Currently npm autocomplete uses `api.npms.io` to search package

https://github.com/withfig/autocomplete/blob/7922234c72777fc1a5be89699ad69be6720fa838/src/npm.ts#L32-L34

and `registry.npmjs.org` to search package version

https://github.com/withfig/autocomplete/blob/7922234c72777fc1a5be89699ad69be6720fa838/src/npm.ts#L48

Both api only support public npm registry.

My suggestion is to use `npm --json search` command for packages and `npm --json info` for versions.

bondz:
Fixes #2255 

Tested locally with public and scoped packages. Also, updating the install command filter strategy to `fuzzy` fixes an issue where scoped packages would not previously show up because the `@` symbol was not considered part of the search.

There's a bug where selecting a scoped package enters two `@` symbols currently.

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -26,34 +26,8 @@ export const createNpmSearchHandler =
       return [];
     }
     // Add optional keyword parameter
-    const keywordParameter =
-      keywords?.length > 0 ? `+keywords:${keywords.join(",")}` : "";
+    const keywordParameter = keywords?.length > 0 ? keywords.join(" ") : "";
 
-    const queryPackagesUrl = keywordParameter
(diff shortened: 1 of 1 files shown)


bondz:
This could potentially break mgnl, although it works on my machine™️

I don't have enough information about the tool.

bondz:
@grant0417 @mschrage any changes you'd like me to make?

grant0417:
I think the changes are good, I just need to stress test them and ensure we dont have any regressions. Will try to do this soon.

Issue #2268, opened 2024-02-11
branches deleted in local still show up in git hub after executing the push -all command


vijay-koyana:
### Sanity checks

- [X] My issue relates to a specific CLI completion spec (e.g. `git checkout` is missing options in `git` completion spec). If your issue is more general, please create your issue here: [withfig/fig](https://github.com/withfig/fig/issues/new?assignees=&labels=NEED_TO_LABEL&template=main_issue_template.yml)
- [X] I have searched [github.com/withfig/autocomplete/issues](https://github.com/withfig/autocomplete/issues?q=) and there are no duplicates of my issue
- [X] I would like to work on this.

### What CLI tool does this relate to?

git

### Which statement makes the most sense?

Something else

### Issue Details

i have a branch in my Git hub.
I deleted the branch using git branch -d <branch name>
then git push

now i am refreshing on the github page. the branch i deleted still exists.

Issue #2101, opened 2023-09-20
bug: Deleted branches still appearing in git checkout suggestions


anelrh26:
### Checks

- [X] I have searched [github.com/withfig/fig/issues](https://github.com/withfig/fig/issues?q=) and there are no duplicates of my issue
- [X] I have run `fig doctor` in the affected terminal session
- [X] I have run `fig restart` and replicated the issue again

### Operating system

macOS 13.3.1 (22E772610a)

### Expected behaviour

After deleting my local and remote git branches, I expect the branch names don't show up in the suggestions when I type `git checkout` 

### Actual behaviour

When I delete my local and remote git branches, the branch names still showing up in the suggestions when I type `git checkout` 

### Steps to reproduce

Scenario 1
1. Delete a local branch `git branch -D sample_branch`
2. Type `git checkout` and you will see `sample_branch` is in the suggestion list when it shouldn't

Scenario 2
1. Delete a remote branch `git push --delete origin sample_branch`
2. Update your local git environment with remote `git fetch --prune`
3. Type `git checkout` and you will see `sample_branch` is in the suggestion list when it shouldn't

### Environment

```yaml
fig-details:
  - 2.16.0
hardware-info:
  - model:
  - model-id:
  - chip-id: Apple M1 Pro
  - cores: 10
  - mem: 16.00 GB
os-info:
  - macOS 13.3.1 (22E772610a)
environment:
  - shell: /bin/zsh
  - terminal: iterm
  - cwd: /Users/me/Projects/project
  - exe-path: /usr/local/bin/fig
  - install-method: unknown
  - env-vars:
    - FIGTERM_SESSION_ID: 83d705f2-bf14-4acd-a93b-4b0b949778af
    - FIG_SET_PARENT_CHECK: 1
    - FIG_TERM: 2.16.0
[...]

PR #2412, opened 2024-07-21
fix(git): Separate local and remote branches in branch display#2101

do-not-do-that:
## Bug Cause
When retrieving the branch list, both local and remote branches are fetched and exposed with the same names.
As a result, when users use the autocomplete feature, they see suggestions for branches that have been deleted locally (or remotely).
#2101 #2268

## Additional Features
- If both local and remote branches exist, display as before (description: Branch).
- If only a local branch exists, add "Local branch" to the description.
- If only a remote branch exists, set name to remotes/origin/{branch}, use an alert icon, set description to "Remote branch", and lower the priority.

## Scenario
image

- local & remote : main, develop
- local only : localBranch
- remote only : remoteBranch
### AS-IS
image
image
image

### TO-BE
image
image
image

There could be multiple solutions, such as showing only local branches. Please provide suggestions, and I will make adjustments accordingly.

### DIFF
--- a/src/git.ts
+++ b/src/git.ts
@@ -63,6 +63,8 @@ const postProcessBranches =
     }
 
     const seen = new Set<string>();
+    const local = new Set<string>();
+    const remote = new Set<string>();
     return output
       .split("\n")
       .filter((line) => !line.trim().startsWith("HEAD"))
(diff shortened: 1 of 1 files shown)


do-not-do-that:

PR #2485, opened 2024-10-16
feat(spec): add anvil and cast

coolestowl:
add specs for anvil and cast


coolestowl:

PR #2569, opened 2025-02-21
readme: add ghostty terminal ref

huskyjp:
## Overview
Added a ghostty terminal link support which already merged here: https://github.com/aws/amazon-q-developer-cli/pull/335


mschrage:
recheck

